PR #22, opened 2015-11-26
Multiple element types

nschloe:


dmsartoris:
I tried converting a fluent mesh (with tets, tris and prisms) to a gmsh format using these lines of code:

import meshio
m = meshio.read('my_fluent_mesh.msh', file_format='ansys')
m.write('gmsh_mesh.msh', file_format='gmsh')

The errror I get get:  "WriteError: Specify entity information (gmsh:dim_tags in point_data) to deal with more than one cell type."

How should I modify my code above to handle multiple cell types?

PR #24, opened 2015-11-26
Improved PERMAS export

nilswagner:
Better export in case of multiple element types.

### DIFF
--- a/meshio/permas_io.py
+++ b/meshio/permas_io.py
@@ -132,55 +132,79 @@ def write(
                 (k+1, x[0], x[1], x[2])
                 )
 
-        # Write elements.
-        # translate the number of points per element (aka the type) to a PERMAS
-        # enum.
-        type_to_number = {
-            1: 15,  # point
-            2: 1,  # line
-            3: 2,  # triangle
-            4: 4,  # tetrahedron
-            6: 6,  # prism
-            8: 5,  # hexahedron
-            }
+        #
+        # Avoid non-unique element numbers in case of multiple element types by num_ele !!!
+        #
+        num_ele = 0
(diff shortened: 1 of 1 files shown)


meshio/permas_io.py
```diff
@@ -132,55 +132,79 @@ def write(
...
+        #
+        # Avoid non-unique element numbers in case of multiple element types by num_ele !!!
+        #
+        num_ele = 0
+        for eltyp in cells.keys(): # Loop over available element types
+            
+            if eltyp == 'line':
+                fh.write('!\n')
+                fh.write('        $ELEMENT TYPE = PLOTL2 ESET = PLOTL2\n')
+                cell = cells[eltyp]
```
nschloe: `cell` can be retrieved by `for elem_type, cell in cells.iteritems()` above.

meshio/permas_io.py
```diff
@@ -132,55 +132,79 @@ def write(
...
+                for k, c in enumerate(cell):
+                    n = len(c)
+                    form = '        %8d ' + ' '.join(n * ['%8d']) + '\n'
+                    fh.write(form % (k+1, c[0]+1, c[1]+1))
+                num_ele += len(cell)
+
+            elif eltyp == 'quad':
```
nschloe: All of the `if-else` clauses are equal except for the `$ELEMENT TYPE` string and the number of format values. The first can be put into a dictionary, the second concatenated into one tuple with `num_ele+k+1`, and we don't need this huge repetition anymore.

nschloe:
Can we close this one?

nilswagner:
Yes.
Am 03.12.2015 22:38 schrieb "Nico Schlömer" <notifications@github.com>:

PR #26, opened 2015-11-29
Further improvements

nilswagner:


nschloe:
Again, line breaks.

Also, does this PR supersede PR #24?

nilswagner:
Applied dos2unix. Multiple element types are supported.

meshio/permas_io.py
```diff
@@ -132,59 +154,81 @@ def write(
...
+        fh.write('    $END CONSTRAINTS\n')
+        fh.write('!\n')
+        fh.write('    $LOADING NAME = LOADVAR_1\n')
+        fh.write('    $END LOADING\n')
+        fh.write('!\n')
         fh.write('$EXIT COMPONENT\n')
         fh.write('!\n')
+        fh.write('$ENTER MATERIAL\n')
+        fh.write('!\n')
+        fh.write('    $MATERIAL NAME = STEEL TYPE = ISO\n')
```
nschloe: Are these leftovers from testing?

meshio/permas_io.py
```diff
@@ -48,6 +48,19 @@ def read(filename):
                 else:
                     cells['triangle'] = [data[-3:]]
 
+        if re.search('\$ELEMENT TYPE = QUAD4', line):
+            while True:
+                line = f.readline()
+                if not line:
+                    break
+                if line.startswith('!'):
+                    break
```
nschloe: Those two statements can be condensed to
```
if not line or line.startswith('!'):
```

meshio/permas_io.py
```diff
@@ -100,6 +113,7 @@ def read(filename):
     points = numpy.reshape(points, newshape=(len(points)/3, 3))
 
     for key in cells:
+        print 'key', key
```
nschloe: The reader shouldn't output anything.

meshio/permas_io.py
```diff
@@ -132,59 +154,81 @@ def write(
...
-                    ))
+        #
+        # Avoid non-unique element numbers in case of multiple element types by num_ele !!!
+        #
+        num_ele = 0
+        
+        for meshio_type, cell in cells.iteritems():
+            numcells, num_local_nodes = cell.shape
+            permas_type = meshio_to_permas_type[meshio_type]
+            if permas_type != 'NA':
```
nschloe: Better remove this line and the corresponding entry in the dictionary. If the original mesh contains vertex elements and they cannot be written in PERMAS, the user should receive notification, i.e., an error.

meshio/permas_io.py
nilswagner: No. I have added those lines in order to run the PERMAS Job directly. At least one material definition is needed.  Element sets are assigned to materials within the system variant.

nschloe: But why these specific values? This seems to be coming from a particular application. If the format file is invalid without a MATERIAL, then we should probably name it `DUMMY_MATERIAL` to make clear that this isn't something that's meant to be used.

meshio/permas_io.py
nilswagner: meshio-convert should write the remaining (supported element types) afterwards. What kind of error message should be raised ?

nschloe: If you just remove the `if` and the entry from the `dict`, you'll get a `KeyError` if one of the `meshio_type`s is `'vertex'`. That should be fine.

meshio/permas_io.py
```diff
@@ -38,9 +38,7 @@ def read(filename):
         if re.search('\$ELEMENT TYPE = TRIA3', line):
```
nschloe: All of the `if`s for the different element types can be condensed into one.

nilswagner: How can I achieve that ?

nschloe: Same way as we did the other day:
 * Create a dictionary that translates meshio types into `TRIA3`, `QUAD4` etc
 * Create the search query `'\$ELEMENT TYPE = %s' % permas_type`
 * Then, the only thing that different is the number of point references that need to be extracted (e.g., `-4`). Those can be put in the same dictionary actually, e.g.,
```python
{
  'triangle': (3, 'TRI3')
  # ...
}
```

meshio/permas_io.py
```diff
@@ -132,59 +108,79 @@ def write(
...
-            8: 5,  # hexahedron
+        meshio_to_permas_type = {
+            'line': 'PLOTL2',
+            'triangle': 'TRIA3',
+            'quad': 'QUAD4',
+            'tetra': 'TET4',
+            'hexahedron': 'HEXE8',
+            'wedge': 'PENTA6',
+            'pyramid': 'PYRA5'
             }
```
nschloe: We have almost the exact same dictionary above. This can possibly be shared.

meshio/permas_io.py
```diff
@@ -132,59 +108,79 @@ def write(
...
+        fh.write('    $LOADING NAME = LOADVAR_1\n')
+        fh.write('    $END LOADING\n')
+        fh.write('!\n')
         fh.write('$EXIT COMPONENT\n')
         fh.write('!\n')
+        fh.write('$ENTER MATERIAL\n')
+        fh.write('!\n')
+        fh.write('    $MATERIAL NAME = DUMMY_MATERIAL TYPE = ISO\n')
+        fh.write('!\n')
+        fh.write('        $ELASTIC  GENERAL  INPUT = DATA\n')
```
nschloe: If these numbers are changed to `0.0`, I think the PR is good to go.

PR #54, opened 2016-08-16
Use the future module to make iteritems compatible with both python2 …

timmyL17:
…and python3


nschloe:
Well, tests fail because of
```
from future.utils import iteritems
```
I guess instead of `iteritems` we can just use `items`, too. Would be simpler.

PR #55, opened 2016-08-16
change iteritems to items for python3 compatibility

timmyL17:
Ok we can just change to items(). There might be a slight performance hit in python2, but hopefully not much. It's also cleaner.


nschloe:
We're only doing management stuff with the `items()`, so the performance degradation is insignificant.

PR #62, opened 2017-01-03
Updated msh_io to include a few higher order elements

cbcoutinho:
Howdy @nschloe,

Just a small PR w.r.t gmsh file input: I added some additional elements to meshio.

I needed some additional elements from a .msh file, so I just updated msh_io. I thought you might be interested in incorporating these additional elements. It's nowhere near comprehensive of all of gmsh's available elements, but this was enough for me.

Thanks for the great tool!

Chris

### DIFF
--- a/meshio/msh_io.py
+++ b/meshio/msh_io.py
@@ -56,7 +56,14 @@ def read_buffer(f):
                     3: ('quad', 4),
                     4: ('tetra', 4),
                     5: ('hexahedron', 8),
-                    6: ('wedge', 6)
+                    6: ('wedge', 6),
+                    7: ('pyramid', 5),
+                    8: ('line3', 3),
+                    9: ('triangle6', 6),
+                    10: ('quad9', 9),
+                    11: ('tetra10', 10),
+                    12: ('hexahedron27', 27),
+                    36: ('quad16', 16)
                     }
             for k, line in enumerate(islice(f, num_cells)):
                 # Throw away the index immediately;


nschloe:
Thanks for PR!

With the changes, what can you do that you couldn't do before?

cbcoutinho:
I've been writing my own 2D FE solver that handles higher order meshes (e.g. bi-quadratic and bi-cubic quads), which gmsh can output natively using the `-order #` switch. Meshio only had the built-in elements of order 1, which are also the default so you probably never ran into the issue. When outputting meshes of order 2 and above, the element ID is outside of the predescribed list of IDs you wrote in msh_io,py resulting in an error.

Here's a small example .geo file of a 2D square:
```
// Inputs
length = 2;
width  = 2;
meshSize = 1;
//+
Point(1) = {0,      0,      0, meshSize};
Point(2) = {width,  0,      0, meshSize};
Point(3) = {width,  length, 0, meshSize};
Point(4) = {0,      length, 0, meshSize};
//+
Line(1) = {1, 2};
Line(2) = {2, 3};
Line(3) = {3, 4};
Line(4) = {4, 1};
//+
Line Loop(1) = {1, 2, 3, 4};
//+
Plane Surface(1) = {1};
Transfinite Surface {1};
Recombine Surface {1};
Physical Surface( "domain" ) = {1};
```

I create a mesh by running this input file through gmsh to create a .msh file as follows:
`gmsh test2D.geo -2 -order 1 -o test2D.msh`

The `-2` switch is for a 2D mesh, and the `-order 1` is for order 1 elements (quads have 4 nodes, triangles would have 3, etc). Running the resulting .msh file through meshio results in no errors.

Now rerunning the input .geo file at a higher order creates elements using different element IDs ([see docs here](http://gmsh.info/doc/texinfo/gmsh.html#MSH-ASCII-file-format) for predefined element IDs under _elem-type_). These element IDs were not included into msh_io.py, and resulting in an error in meshio.

Here's how to run gmsh at an order of 2:
`gmsh test2D.geo -2 -order 2 -o test2D.msh`

That results in the following error:
```
In [1]: import meshio

In [2]: meshio.read('test2D.msh')
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-3-b305bebde9c2> in <module>()
----> 1 meshio.read('test2D.msh')

/home/chris/anaconda3/lib/python3.5/site-packages/meshio/__init__.py in read(filename, file_format)
     81 
     82     if file_format == 'gmsh':
---> 83         return msh_io.read(filename)
     84     elif file_format == 'medit':
     85         return medit_io.read(filename)

/home/chris/anaconda3/lib/python3.5/site-packages/meshio/msh_io.py in read(filename)
     15     '''
     16     with open(filename) as f:
---> 17         points, cells = read_buffer(f)
     18 
     19     return points, cells, {}, {}, {}

/home/chris/anaconda3/lib/python3.5/site-packages/meshio/msh_io.py in read_buffer(f)
     62                 # Throw away the index immediately;
     63                 data = numpy.array(line.split(), dtype=int)
---> 64                 t = gmsh_to_meshio_type[data[1]]
     65                 # Subtract one to account for the fact that python indices
     66                 # are 0-based.

KeyError: 10
```

The element type ID `10` is associated with a 9-node quadrilateral. Adding it in like I did gets ride of the error I was seeing.

---

Now I realize this was kind of hackish, and I don't understand the rest of meshio to notice how this interoperates with the rest of your code. It may be that some other parts crash if you want to convert gmsh format to some other format that doesn't support higher order elements - that's a question I have for you. If that's the case, maybe you could throw a warning or something (i.e. 'Careful, some meshio capabilities do not support higher order meshes, stick to reading meshes only')

nschloe:
Ah I see, you just need meshio for reading. That's fair enough I guess.

Issue #43, opened 2015-12-23
treat gmsh's physical groups


nschloe:
[The Gmsh data format](http://geuz.org/gmsh/doc/texinfo/gmsh.html#MSH-ASCII-file-format) offers several of data fields, e.g., `ElementData`, `NodeData` etc. On top of that, there a "Physical Groups", effectively tagging entities. It's not quite clear yet where this information fits in with other mesh formats.

nschloe:
@dajuno, can you provide an example file with physical groups?

dajuno:
nschloe, I will in january, going on a small trip now =)
Cheers

dajuno:
@nschloe, here's a simple 3D example with physical surfaces and a physical volume: https://gist.github.com/dajuno/691fa18db83f7b36db5c
(Unit cube where all 6 surfaces are tagged differently.)
And with a cylindrical subdomain: https://gist.github.com/dajuno/6545e964cf9c499dc4e1
I.e., there are 2 physical volumes. The physical surfaces are the cylindrical surface (interface), and the outer surfaces (front and back separated into circle and "cut out" cube).

gmsh has physical lines and points, too. Relevant for 2D cases.

Until now, in order to convert the gmsh mesh to HDF5 I have been using `dolfin-convert` to convert to XML, creating mesh.xml, mesh_facet_region.xml & mesh_physical_region.xml, and then a small script using the dolfin functions to create a hdf5 mesh, based on:
```
from dolfin import *
mesh = Mesh('mesh.xml')
subdomains = MeshFunction("size_t", mesh, "mesh_physical_region.xml")
boundaries = MeshFunction("size_t", mesh, "mesh_facet_region.xml")

# Write HDF5
hdf = HDF5File(mesh.mpi_comm(), "mesh.h5", "w")
hdf.write(mesh, "/mesh")
hdf.write(subdomains, "/subdomains")
hdf.write(boundaries, "/boundaries")
```
and
```
# Read HDF5
mesh = Mesh()
hdf = HDF5File(mesh.mpi_comm(), "mesh.h5", "r")
hdf.read(mesh, "/mesh", False)
subdomains = CellFunction("size_t", mesh)
hdf.read(subdomains, "/subdomains")
boundaries = FacetFunction("size_t", mesh)
hdf.read(boundaries, "/boundaries")
```
taken somewhere from the fenics Q&A.

Let me know if you need more info or if I can be of any help!

dajuno:
[...]

PR #63, opened 2017-01-04
Add addtional element types, include cell_data and field_data

cbcoutinho:
Howdy @nschloe,

I also needed some access to physical groups, so here's an update I made to return the physical group information if available. Otherwise, it just returns empty dictionaries as before. Feel free to hold off on merging for a bit, I'm still messing around with it.

I'm not sure what point_data is supposed to be, is that related to another mesh format?

Cheers,
Chris

### DIFF
--- a/meshio/msh_io.py
+++ b/meshio/msh_io.py
@@ -96,7 +96,7 @@ def read_buffer(f):
 
                 # There should be a 1-to-1 match between elements,
                 # assuming that the elements are in consecutive order
-                if bool(field_data) or data[0] != 0:
+                if bool(field_data) or data[3] != 0:
                     if t[0] in cell_data:
                         cell_data[t[0]].append(data[3])
                     else:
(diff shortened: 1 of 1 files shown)


nschloe:
`point_data` is the data in mesh vertices (or wherever your DOFs are). I guess that's a common thing across most mesh formats.

nschloe:
Many people requested the Gmsh physical group feature; see, e.g., https://github.com/nschloe/meshio/issues/43. Extra points if we get those use cases covered. (I'll help out where necessary.)

cbcoutinho:
Oh alright, I didn't don't have point data so I just left it empty. 

I'm not at my computer to test their 3d example, but at a quick glance I think it should work. 

One downside in the way I implemented physical groups is omitting the dimension. I personally didn't need it, but if that's an issue for others we can try to come up with a better solution.

cbcoutinho:
Another drawback - if an element is part of multiple groups, there's going to be a problem.

nschloe:
Well, something is better than nothing. It'd be great if we could include a test for this such that I can remember in the future what meshio can an can't do.

cbcoutinho:
Do you have a test format? I could write one, but I have no idea how Travis
works.

Chris

PR #103, opened 2017-07-06
Fixed undefined variable gmsh_io

cbcoutinho:


nschloe:
`points` is an array, so `points = []` would be more appropriate.

cbcoutinho:
Good point - done

nschloe:
Not sure why circleci won't build. Let me check back with their support.

nschloe:
Aha, CircliCI's workflows (which meshio uses) doesn't support forked PRs yet, https://circleci.com/docs/2.0/faq/#can-i-build-fork-prs-using-workflows. I could swear that it briefly worked a while ago, but this might have been a fluke. Will fix this in master.

PR #111, opened 2017-08-14
updated broken link to H5M

gdmcbain:
I also found http://www.mcs.anl.gov/~fathom/moab-docs/h5mmain.html but I think it's older.


nschloe:
Can we make this http**s**?

PR #114, opened 2017-08-15
Skip tests depending on optional VTK

ghisvail:


ghisvail:
You may apply the same pattern for tests requiring `lxml` and 'h5py'.

PR #122, opened 2017-09-01
Log a warning instead of raising an exception for unknown tags

lgiraldi:
When using fenics/mshr to create a mesh, a tag <data /> is added to the xml file that causes meshio to raise an exception in _read_mesh function.


nschloe:
pylint says:
```
W: 64,16: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)
```

lgiraldi:
Fixed.

PR #133, opened 2017-10-25
Consider vertex

nilswagner:


nilswagner:
1-node point aka vertex is supported in permas_io.py

Issue #145, opened 2017-11-13
GMSH: Add quad16 and hexahedron64


dalcinl:
The patch is trivial, if you prefer a PR, let me know and I'll submit one with the diff below.

```diff
diff --git a/meshio/gmsh_io.py b/meshio/gmsh_io.py
index fc53da2..36aa8a8 100644
--- a/meshio/gmsh_io.py
+++ b/meshio/gmsh_io.py
@@ -30,6 +30,7 @@ num_nodes_per_cell = {
     'pyramid14': 14,
     'line4': 4,
     'quad16': 16,
+    'hexahedron64': 64,
     }
 
 # Translate meshio types to gmsh codes
@@ -52,6 +53,7 @@ _gmsh_to_meshio_type = {
         14: 'pyramid14',
         26: 'line4',
         36: 'quad16',
+        92: 'hexahedron64',
         }
 _meshio_to_gmsh_type = {v: k for k, v in _gmsh_to_meshio_type.items()}
```

nschloe:
Perhaps we can enhance this with the rest of the gmsh types (if there are any).

nschloe:
And yes, a PR would make things really easy for me.

PR #146, opened 2017-11-14
Gmsh: High order simplex and tensor cells

dalcinl:
As requested in #145, I'm adding support for all high order cells.


nschloe:
Perfect, thanks!

PR #150, opened 2017-11-19
Gmsh: Support for writing optional PhysicalNames section

dalcinl:


nschloe:
This still needs a test in `test/test_gmsh.py`. Now that I look at this, I see that `field_data` isn't tested very well at all. Let's add this. We should probably do something like:

   1. Add a `add_field_data()` method in `test/helpers.py` (just like the `add_point_data` and `add_cell_data` methods there).
   2. Add a test in `test/test_gmsh.py` where any mesh is equipped with the field data.
   3. Assert in `test/helpers.py::write_read` that input field data and output field data are the same.

nschloe:
Best ping me once you've done this (I'm not getting notification for PR pushes.)

dalcinl:
@nschloe PR updated. Do you want me to extend testing with bad field data to trigger the `logging.warning()` path and improve coverage?

nschloe:
Thanks for the fixes! We can increase the coverage later.

PR #164, opened 2017-12-11
Change names for higher order elements

chrisrichardson:
Names are abbreviated in XDMF format.

### DIFF
--- a/meshio/xdmf_io.py
+++ b/meshio/xdmf_io.py
@@ -24,6 +24,8 @@ def read(filename):
 
 
 numpy_to_xdmf_dtype = {
+    numpy.dtype(numpy.uint32): ('UInt', '4'),
+    numpy.dtype(numpy.uint64): ('UInt', '8'),
     numpy.dtype(numpy.int32): ('Int', '4'),
     numpy.dtype(numpy.int64): ('Int', '8'),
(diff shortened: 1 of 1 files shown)


nschloe:
In http://www.xdmf.org/index.php/XDMF_Model_and_Format#XML_Element_.28Xdmf_ClassName.29_and_Default_XML_Attributes, I see the keywords as they are implemented now. As usual, there's appears to be no consistency. I'd be fine with _adding_ the abbreviated keywords. The reversed list `meshio_to_xdmf_type` should probably be made explicit.

nschloe:
Hang on, I have a better idea.

chrisrichardson:
You're probably right - maybe it accepts both `Tri_6` and `Triangle_6`. We can check the VTK source code...

nschloe:
I've coughed up something in #165, will merge and release if it passes.

chrisrichardson:
So, this page: http://www.xdmf.org/index.php/XDMF_Model_and_Format has `Tri_6` and `Tet_10` etc., and it seems to work in paraview... but the Xdmf source has `Triangle_6` etc. I can ask Dave Demarle about it.

nschloe:
Never mind asking, #165 allows for both variants so we should be good.

nschloe:
Alright, check out 1.10.3.

Issue #186, opened 2018-01-17
read nodal/cell data from med


tianyikillua:
I start to work on reading field (in general) data from a med file, as you can see [here](https://github.com/tianyikillua/meshio/commit/4106f66d0f1a44bdceaa3ae4082e2f6a0665b825).

Reading nodal field seems easy and should already work.

For reading cell field it may demand some work. The MED format is strongly influenced by Code_Aster, and they in general define "cell field" directly on gauss points. We can adopt the following approach

- for elements where lies only 1 gauss point (linear simplices, or reduced integration), simple (TODO);
- otherwise, we can just perform an arithmetic mean of all the values at the Gauss points inside one particular element.

In the latter case, `meshio` does not only "reading" but also some "computation"...it seems outside its scope. Hence we can also just neglect these fields when the number of Gauss points is larger than 1.

nschloe:
Nice! You could first make a PR for nodal data, then for cell data if it's that much more complex.

Cell data can be anything stored "per cell", so I don't see nothing wrong storing a vector that keeps the data at whatever points MED prescribes. This could also be plugged into any other mesh format, just that the viewers wouldn't interpret the data as "values at Gauss points" but simply as vectors associated with cells. Never mind that I guess.

You're right: If you want to convert a MED file with such cell data to, e.g., VTU, and would like ParaView to show the data correctly, you'd have to do some transformation. I wouldn't mind doing that in the MED reader though, as an option. One could then use, e.g.,
```
meshio-convert in.med out.vtu
```
to get the 1-to-1 vector transformation, and
```
meshio-convert --translate-med-cell-data in.med out.vtu
```
to generate a single cell value from the Gauss-point values.

By the way, until now I haven't seen any description of the MED file online, so I'll have to rely on you to get it right. If you have a link to the standard, that would be great.

tianyikillua:
I'll make a PR for the nodal data in the first place. I attach [here](https://github.com/nschloe/meshio/files/1639089/test-med-data.zip) a test MED file generated by a Code_Aster calculation.
```
points, cells, point_data, cell_data, field_data = meshio.read("test.rmed", file_format="med")
```
should give you a `point_data` containing three fields: the displacement, and stress/strain averaged at nodes (`SIGM_NOEU` and `EPSI_NOEU`) computed by Code_Aster.

Attention, the ordering for these symmetric 2nd order tensors is the following in MED
```
XX YY ZZ XY XZ YZ
```
[...]

PR #187, opened 2018-01-17
Read/write nodal/cell data from MED

tianyikillua:
We should now be able to read/write nodal/cell data from a MED file, see also #186.
- `NOEU`: nodal data
- `ELGA` : Gauss point data (for writing, only fields defined on exactly 1 Gauss point per cell can be written)
- `ELNO` : Gauss point data extrapolated at nodes per element (some basically the DG field)

For reading, the array of `ELGA` or `ELNO` fields has 3 dimensions `nbr`, `nco` and `nga` where `nbr` refers to the number of the concerned particular (geometric typed) cells, `nco` is the number of physical components of the field, and `nga` is the number of Gauss or elemental nodal points per cell. 

If `nga == 1`, then only one supporting point is defined per cell (the most usual case). The array will be cut off to be of shape `(nbr, nco)`. For scalar fields (`nco == 1`), it will be further cut off to be an 1-d array (convention in `meshio`). In other cases, the array is of shape `(nbr, nco, nga)`.

If `nga == nn` where `nn` is the number of nodes per cell, then we consider it to be a `ELNO` field. Otherwise it is a general `ELGA` field defined on `nga` Gauss points. Reading general `ELGA` fields is ok but writing them needs additional work, since we need to write Gauss point information (location, weights) into the MED file. It can be done in the future.

### DIFF
--- a/meshio/__about__.py
+++ b/meshio/__about__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-__version__ = '1.11.4'
+__version__ = '1.11.5'
 __author__ = u'Nico Schlömer'
 __author_email__ = 'nico.schloemer@gmail.com'
 __copyright__ = \
--- a/meshio/helpers.py
+++ b/meshio/helpers.py
@@ -184,7 +184,10 @@ def write(filename,
             write_binary=(file_format == 'gmsh-binary')
             )
     elif file_format == 'med':
-        med_io.write(filename, points, cells, point_data=point_data)
+        med_io.write(
+            filename, points, cells,
+            point_data=point_data,
+            cell_data=cell_data)
     elif file_format == 'medit':
         medit_io.write(filename, points, cells)
     elif file_format == 'dolfin-xml':
--- a/meshio/helpers.py
+++ b/meshio/helpers.py
@@ -182,7 +182,7 @@ def write(filename,
             write_binary=(file_format == 'gmsh-binary')
             )
     elif file_format == 'med':
-        med_io.write(filename, points, cells)
+        med_io.write(filename, points, cells, point_data=point_data)
     elif file_format == 'medit':
         medit_io.write(filename, points, cells)
     elif file_format == 'dolfin-xml':
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -5,6 +5,7 @@ I/O for MED/Salome, cf.
 <http://docs.salome-platform.org/latest/dev/MEDCoupling/med-file.html>.
 
 .. moduleauthor:: Nico Schlömer <nico.schloemer@gmail.com>
+                  Tianyi Li <tianyikillua@gmail.com>
 '''
 import numpy
 
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -92,12 +92,14 @@ def _read_data(cha):
     cell_data = {}
     field_data = {}
     for name, data in cha.items():
-        submeshes = data.keys()
-        assert len(submeshes) == 1
-        data = data[list(submeshes)[0]]
+        supps = ['NOE', 'MAI']
+        if all(supp not in data for supp in supps):
+            submeshes = data.keys()
+            assert len(submeshes) == 1
+            data = data[list(submeshes)[0]]
         supp = list(data.keys())[0]
 
-        if supp == 'NOE':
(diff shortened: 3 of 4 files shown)


tianyikillua:
Hum...Do I need to implement the corresponding `write` functions or otherwise the unittest will fail?

nschloe:
Thanks for the PR, looking good so far. It still needs a test though, and in order to do that, we also need point data support for _writing_. What the tests do, essentially, is to create meshes with some data attached to it, write that out to a file, then read the file and check if the data is still there (and didn't change).  Check out `test/test_vtu.py`, it should be obvious what you need to add to `test/test_medit.py`.

nschloe:
Also, about the ordering of tensor data: It'd be great if one could read a medit file, write it to vtu, and get a proper representation of tensor data. If I understand correctly, this is not the case right now.
Perhaps we should reorder the tensor data when reading and writing with medit. What do you think?

nschloe:
Thanks for the good work here! I've never been able to properly test MEDIT mesh files in the real world, so I've developed the I/O half "blind". You can certainly expect to find some bugs.

tianyikillua:
I'm still working on a bug concerning reading data in ParaView-MED (Salome). Constructing a MED file using `h5py` is like writing FEniCS using Assembly.

Concerning the ordering for 2nd symmetric tensors, we have to assume (or prescribe) that

- all the data that the user import from `meshio`  (rather it be a point data or cell data, from arbitrary mesh format), that contain exactly 6 values correspond indeed to the 6 components of a symmetric 2nd order tensor
- that the ordering is given according to VTK convention (11, 22, 33, 12, 23, 13)

However it is not the case in general. These 6 values can have arbitrary meanings...I would rather that the user take care of the ordering himself. What do you think? The problem is that we don't have a name for each component in `meshio`.

nschloe:
> Constructing a MED file using h5py is like writing FEniCS using Assembly.

I feel you! :smile:

> The problem is that we don't have a name for each component in meshio.

What do you mean by _component_? Component of the tensor?

If I understand correctly, the problem is that it'd be nice to translate MEDIT tensors to VTK files so one can look at them in ParaView or so. For this, however, it's required to stick to VTK's tensor ordering `(11, 22, 33, 12, 23, 13)`, and you want to be more flexible there. Right?

tianyikillua:
For reading array-valued data from MED, not a problem because in MED we have a name for each component via the `NOM` attribute of the field group: we have `SIXX, SIYY, SIZZ, SIXY, SIXZ, SIYZ`. If we prescribe that the tensor-valued data contained in `*_data` of `meshio` should always respect the VTK ordering convention, we just reverse the 5th and 6th component. This permits the natural conversion from MED to VTU while respecting the ordering.
```
points, cells, point_data, *_ = meshio.read("mesh.med")
meshio.write("mesh.vtu", points, cells, point_data)
```

For writing array-valued data to a MED file, since in `meshio` we don't have a *tag* for each component  of a tensor-valued data, we don't know for sure if we need to reverse the 5th and the 6th column. Suppose that a user stores for each node `(x, y)` the following nodal array-valued data `v`
```
v = (x, y, x+y, x-y, x**2, y**2)
```
this data doesn't have a tensor structure but unluckily possess 6 components. So for writing to MED maybe we should let the user do whatever he likes.

nschloe:
I understand. I had been thinking about VTK's own tensor type (page 6 in [the specs](https://www.vtk.org/wp-content/uploads/2015/04/file-formats.pdf)), not arbitrary vector data. So, yeah, the problem is that in meshio we don't discriminate between vector-data of length 6 and symmetric tensor data.

Perhaps the cleanest solution would be to read and write all the vectors just as they are. If you want to convert a MEDIT tensor into a VTK tensor, you'll have to do extra work, just as you suggested earlier.

tianyikillua:
For nodal reading/writing it's ok I think. For reading/writing cell datum that is defined at exactly 1 Gauss point should also be easy. I would take some time maybe this weekend.

nschloe:
Just let me know whenever you're ready. If you want you can also bump the version number in `__about__.py` so I can release the thing right away then.

tianyikillua:
It's ready, before it becomes endless...Let's call it a "pull-request" Some improvements are to be done in the future. I'll leave you take care of the release stuff :-)

nschloe:
Thanks very much for the nice contribution!

PR #194, opened 2018-01-25
Add support for wedge elements in ExodusII

acroucher:
I would like support for wedge elements in ExodusII output- it looks like this is all that's needed? (and thanks for your work on meshio!)


nschloe:
Indeed. Thanks for the contribution!

nschloe:
I just released 1.11.6 with the fix.

Issue #208, opened 2018-03-13
KeyError: 'line' in ansys_io.py line 453, in write


ronaldtanner:
Using meshio 1.11.9 (python 3.4) gives me a keyerror when trying to convert a 2D gmesh file [model.zip](https://github.com/nschloe/meshio/files/1807202/model.zip) to ansys:
```
meshio-convert --input-format gmsh-ascii --output-format ansys-binary model.msh model.ansys
```

nschloe:
That's because ANSYS files don't support `line` elements.

nschloe:
Closing for lack of response. Feel free to reopen at any time.

PR #209, opened 2018-03-22
add check for filename and file_format during read and write

iitrabhi:
The filename check was made only in read and not write. Also, no check was made to see whether the user-specified file format is supported by the code. This lead to 'KeyError' on the prompt which is not very descriptive for the end user. Also, there was code repetition in 'read' and 'write' modules. I have added a new function 'check_filename' that would check whether the specified file format is supported by the code. This is in regards to Issue #208.

### DIFF
--- a/meshio/helpers.py
+++ b/meshio/helpers.py
@@ -83,6 +83,7 @@ def check_filename(filename, file_format=None):
     # Checks the input and output file name and format.
     # https://stackoverflow.com/q/4843173/353337
     assert isinstance(filename, str), 'Filename is not a string.: {}'.format(filename)
+    assert isinstance(filename, str), 'Invalid input \'{}\''.format(filename)
 
     if not file_format:
         # deduce file format from extension
--- a/meshio/helpers.py
+++ b/meshio/helpers.py
@@ -82,7 +82,6 @@ _extension_to_filetype = {
 def check_filename(filename, file_format=None):
     # Checks the input and output file name and format.
     # https://stackoverflow.com/q/4843173/353337
-    assert isinstance(filename, str), 'Filename is not a string.: {}'.format(filename)
     assert isinstance(filename, str), 'Invalid input \'{}\''.format(filename)
 
     if not file_format:
--- a/meshio/helpers.py
+++ b/meshio/helpers.py
@@ -82,17 +82,16 @@ _extension_to_filetype = {
 def check_filename(filename, file_format=None):
     # Checks the input and output file name and format.
     # https://stackoverflow.com/q/4843173/353337
-    assert isinstance(filename, str), 'No such file or directory: {}'.format(filename)
+    assert isinstance(filename, str), 'Filename is not a string.: {}'.format(filename)
 
     if not file_format:
         # deduce file format from extension
         extension = '.' + os.path.basename(filename).split(os.extsep, 1)[-1]
         # check whether the file format is supported
-        if extension in _extension_to_filetype:
-            file_format = _extension_to_filetype[extension]
-        else:
-            assert False, ('Unknown file extension \'{}\' of \'{}\'.'
-                           .format(extension, filename))
+        assert extension in _extension_to_filetype, ('Unknown file extension \'{}\' of \'{}\'.'
+                                                     .format(extension, filename))
+        file_format = _extension_to_filetype[extension]
+
     return filename, file_format
 
 
@@ -140,9 +139,8 @@ def read(filename, file_format=None):
         'exodus': exodus_io,
     }
     # check whether the user specified format is supported
-    if file_format not in format_to_reader:
-        assert False, ('Unknown file format \'{}\' of \'{}\'.'
-                       .format(file_format, filename))
+    assert file_format in format_to_reader, ('Unknown file format \'{}\' of \'{}\'.'
+                                             .format(file_format, filename))
     return format_to_reader[file_format].read(filename)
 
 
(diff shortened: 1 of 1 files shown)


meshio/helpers.py
```diff
@@ -76,7 +76,24 @@
...
     '.xdmf': 'xdmf',
     '.xmf': 'xdmf',
-    }
+}
+
+
+def check_filename(filename, file_format=None):
+    # Checks the input and output file name and format.
+    # https://stackoverflow.com/q/4843173/353337
+    assert isinstance(filename, str), 'No such file or directory: {}'.format(filename)
```
nschloe: The error message is misleading, it should be "Is no string". Not sure if this is needed at all.

meshio/helpers.py
```diff
@@ -76,7 +76,24 @@
...
+
+    if not file_format:
+        # deduce file format from extension
+        extension = '.' + os.path.basename(filename).split(os.extsep, 1)[-1]
+        # check whether the file format is supported
+        if extension in _extension_to_filetype:
```
nschloe: Perhaps better:
```python
try:
    # bla
except KeyError:
    raise KeyError('Could not deduce finly type from extension \'{}\''.format(extension))
```

meshio/helpers.py
```diff
@@ -126,8 +138,11 @@ def read(filename, file_format=None):
         #
         'xdmf': xdmf_io,
         'exodus': exodus_io,
-        }
-
+    }
+    # check whether the user specified format is supported
+    if file_format not in format_to_reader:
+        assert False, ('Unknown file format \'{}\' of \'{}\'.'
```
nschloe: Better:
```python
assert file_format in format_to_reader, 'Error message bla bla'
```

meshio/helpers.py
```diff
@@ -76,7 +76,24 @@
...
+    assert isinstance(filename, str), 'No such file or directory: {}'.format(filename)
+
+    if not file_format:
+        # deduce file format from extension
+        extension = '.' + os.path.basename(filename).split(os.extsep, 1)[-1]
+        # check whether the file format is supported
+        if extension in _extension_to_filetype:
+            file_format = _extension_to_filetype[extension]
+        else:
+            assert False, ('Unknown file extension \'{}\' of \'{}\'.'
```
nschloe: Instead of `assert False`, do an
```python
assert extension in _extension_to_filetype, 'Error message bla'
```
before.

nschloe:
Issue #208 is different, but I still like this PR. Better error message are better. :+1:.

meshio/helpers.py
nschloe: Sure. I don't know anymore why this assertion was inserted at all though. Anyhow, let's keep it. Perhaps the input can also be shown in the error message (à la `Invalid input '123'.`)

meshio/helpers.py
iitrabhi: Thanks for the comments. will incorporate them into the code.

meshio/helpers.py
```diff
@@ -76,7 +76,23 @@
...
+
+    if not file_format:
+        # deduce file format from extension
+        extension = '.' + os.path.basename(filename).split(os.extsep, 1)[-1]
+        # check whether the file format is supported
+        assert extension in _extension_to_filetype, ('Unknown file extension \'{}\' of \'{}\'.'
+                                                     .format(extension, filename))
+        file_format = _extension_to_filetype[extension]
+
+    return filename, file_format
```
nschloe: Does this really need to return the `filename`? You could call this function `format_from_filename`). Also, the `file_format` argument is somewhat weird.

iitrabhi: If the user specifies a different extension in `filename` and `file_format` then currently the code gives preference to `file_format` but while writing the output gives extension given in the `filename`. This could sometime lead to confusion.

I made the above function to check the following:

- Whether the input filename is a string.
- If the file_format is not specified then get the file_format from the filename.
- If there is a conflict between the formats specified via extension and file_format, then, give preference to the user-specified file_format.

The final check was not made since the `read` and `write` method handle `file_format` in different ways. Thus, in its current state, the function doesn't require to return `filename` but has to return `file_format`.

nschloe: Fair enough, but you still don't need to return `filename`, right? It's never changed.

nschloe: Also, the `None` check should happen outside the function for clarity, i.e.,
```python
if file_format is None:
    file_format = format_from_filename(filename)
```
This is much clearer.

iitrabhi: But if we do this way then we have to separately make the filename is a string check in both read and write. In the current form, we can leave the check inside the format_from_filename and then the functionality would be as follows:
- If the user has specified a format then the function would check for the validity of the filename only.
- If he has not specified a format then the function would get the format from the extension.

nschloe: > But if we do this way then we have to separately make the filename is a string check in both read and write

I don't get it. Why?

iitrabhi: Because if the user specifies a `file_format` then the code would not check whether the filename is a string as it would not call the function. So, we need to check separately in both `read` and `write`.

nschloe:
In the meantime, a few `assert`s with proper error messages on failure have landed in meshio. This obsoletes this PR I think. Feel free to reopen if you think that this is still useful.

iitrabhi:
Thanks for considering the suggestion. The changes are good.

PR #215, opened 2018-04-04
Enhance MED I/O

tianyikillua:
- Add more element types for I/O
- Support reading time-series data

A `Displacement` field can have many time steps within a single MED file. It will be converted to many *independent* `point_data` or `cell_data`: `Displacement[$i] - $t`.

### DIFF
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -76,15 +76,24 @@ def _read_data(cha):
     cell_data = {}
     field_data = {}
     for name, data in cha.items():
-        ts = data.keys()
-        data = data[list(ts)[-1]]  # only read the last time-step
+        ts = sorted(data.keys())  # associated time-steps
+        if len(ts) == 1:  # single time-step
+            names = [name]  # do not change field name
+        else:  # many time-steps
+            names = [None] * len(ts)
+            for i, key in enumerate(ts):
+                t = data[key].attrs['PDT']  # current time
+                names[i] = name + "[{:d}] - {:g}".format(i, t)
 
         # MED field can contain multiple types of data
-        for supp in data:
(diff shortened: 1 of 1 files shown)


meshio/med_io.py
```diff
@@ -9,14 +9,20 @@
 '''
 import numpy
 
-
+# [code_aster_src]/catalo/cataelem/Commons/mesh_types.py
```
nschloe: An actual HTTPS link would be nice.

meshio/med_io.py
```diff
@@ -70,15 +76,24 @@ def _read_data(cha):
...
-        ts = data.keys()
-        data = data[list(ts)[-1]]  # only read the last time-step
+        ts = sorted(data.keys())  # associated time-steps
+        if len(ts) == 1:  # single time-step
+            names = [name]  # do not change field name
+        else:  # many time-steps
+            names = [None] * len(ts)
+            for i, key in enumerate(ts):
+                t = data[key].attrs['PDT']  # current time
+                names[i] = name + "[{:d}] - {:.2f}".format(i, t)
```
nschloe: Why only `.2f`?

meshio/med_io.py
tianyikillua: Ok I will replace it by `https://bitbucket.org/code_aster/codeaster-src/src/`

meshio/med_io.py
nschloe: You can also just add another commit; I'll squash when merging to master.

meshio/med_io.py
nschloe: With a location in code please. Just the repo isn't enough.

nschloe:
Thanks for the nice PR. -- As usual! :+1:

Issue #200, opened 2018-02-07
Error with relative paths


nicoguaro:
I am trying to read a mesh that is in a folder above the current one using the following code

    import meshio

    fname = "../square.msh"
    points, cells, point_data, cell_data, field_data = \
         meshio.read(fname)

But I obtain the following error

    Traceback (most recent call last):
    
      File "<ipython-input-5-c5965f4f0bf1>", line 1, in <module>
        runfile('/home/nguarinz/Documents/meshio_test.py', wdir='/home/nguarinz/Documents')
    
      File "/home/nguarinz/anaconda2/lib/python2.7/site-packages/spyder/utils/site/sitecustomize.py", line 705, in runfile
        execfile(filename, namespace)
    
      File "/home/nguarinz/anaconda2/lib/python2.7/site-packages/spyder/utils/site/sitecustomize.py", line 94, in execfile
        builtins.execfile(filename, *where)
    
      File "/home/nguarinz/Documents/meshio_test.py", line 6, in <module>
        meshio.read(fname)
    
      File "/home/nguarinz/anaconda2/lib/python2.7/site-packages/meshio/helpers.py", line 99, in read
        file_format = _extension_to_filetype[extension]
    
    KeyError: '../square.msh'

Checking lin 99 of [helpers.py](https://github.com/nschloe/meshio/blob/master/meshio/helpers.py) (98 in the current branch), it seems to be because it takes the extension as all the characters after the first point. I suppose that something like this

    extension = filename[filename.rfind("."):]

might work.

I am using meshio 1.11.2 with Python 2.7.13 (Anaconda 64 bits) in Linux Mint 18.2.

gdmcbain:
I was going to propose the more direct [`os.path.splitext`](https://docs.python.org/3/library/os.path.html?highlight=splitext#os.path.splitext), but saw that it had been commented out in the corresponding part of `helpers.write` in d2411cce1f4b6c08a1e3131187b49b20025243d5:

```python
        # _, extension = os.path.splitext(filename)
        extension = '.' + filename.split(os.extsep, 1)[-1]
```

I think that that would have worked in your case, i.e. `os.path.splitext('../square.msh')[-1] == '.msh'`.

gdmcbain:
[...]

PR #228, opened 2018-05-08
Support extensions with more than 1 dot.

leoschwarz:
Currenty file type detection based on extension fails, if there is a
filename of the form `mesh.1.off` instead of `mesh.off` (with only a
single dot). This fix replaces the custom code with the default
`os.path.splitext` function which always only returns the rightmost part
or an empty string.


test/test_helpers.py
```diff
@@ -22,6 +22,8 @@
     'test.vtk',
     'test.vtu',
     'test.xmf',
+    'test.1.off',
```
leoschwarz: I'm not sure if you want to merge it like this or if I should add an example for all variants or a dedicated test?

nschloe:
This won't play nicely with double extensions like `.mesh.gz` (see the PERMAS format). Not sure why we don't have a test for this.

leoschwarz:
My bad, I should have slept over the commit before sending it and do more testing, thanks for pointing this out. I can add a test for that tomorrow.

I guess another and more robust approach would be to split at the first dot and check if it is a known extension (`.mesh.gz` would be caught this way) and if not part by part of it would be stripped away from left to right until something is found or nothing is left (`.1.mesh` would not resolve, but then `.mesh` would).

gdmcbain:
On double extensions, [above](https://github.com/nschloe/meshio/pull/228#issuecomment-387528059)

> This won't play nicely with double extensions like .mesh.gz (see the PERMAS format). Not sure why we don't have a test for this.

I thought there was.  I had mentioned there being a 'test.post.gz' in test/test_helpers.py in the discussion of #200 on [2018-02-08](https://github.com/nschloe/meshio/issues/200#issuecomment-364001420) but see that the second suffix was lopped in 6ea9fd9.  I'm not immediately sure why it was lopped.

leoschwarz:
@gdmcbain Actually when I try adding the tests that were removed I get the error `OSError: Not a gzipped file (b'!\n')`.

I think this should not be merged before the there are some tests for it.

nschloe:
`test_helpers.py` got kicked out in another PR. Can you just move the new file names into the respective `test_*` files?

Issue #230, opened 2018-05-09
test_helpers fails without h5py


gdmcbain:
In order to start looking at #228  and #200, I cloned meshio in a fresh virtual environment (Python 3.6 or 3.7).  It fails on `test_helpers.py` with
> E           ModuleNotFoundError: No module named 'h5py'

I see that this is avoided in `.circleci/config.yml` by `apt-get install python3-h5py`.

gdmcbain:
The attempt to fix this in c832d40 isn't right; it unintentionally skips the whole `test_helpers` module rather than just the test on `.med`.

gdmcbain:
Oops, netCDF4 is needed too for `test.e`.

leoschwarz:
I don't think tests should just be skipped if some dependency is not installed. Since running the package tests is not something to be expected to be done by every downstream user of a library, I don't think it's wrong to expect to have all dependencies (including optional ones if they are supported on all relevant platforms) be installed when running the tests.

However I ran into the same issue, maybe it would suffice to add the requirements to the setup.py `tests_require` field?

nschloe:
> However I ran into the same issue, maybe it would suffice to add the requirements to the setup.py tests_require field?

h5py and netcdf are actually (optional) dependencies of meshio itself, not only the tests. They are listed in setup.py and you install them with, e.g., `pip install .[all]`. This is the canonical way.

I'm okay skipping tests if a particular library cannot be imported; after all, it's the circleci tests that is eventually decisive. On your local machine, you want to make life as easy as possible.

nschloe:
Should be fixed now.

PR #231, opened 2018-05-09
separate tests needing h5py to be skipped without it

gdmcbain:
This is an attempt at fixing #230.


nschloe:
Closing in favor of #234.

PR #232, opened 2018-05-09
Test integration in setup.py.

### LINKED ISSUE #230: test_helpers fails without h5py

leoschwarz:
With this it becomes possible to just run `python setup.py test` or `python setup.py
pytest` and it will install all test dependencies and run the test.

This would fix #230.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[aliases]
+test=pytest
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,16 @@ with open(os.path.join(base_dir, 'meshio', '__about__.py'), 'rb') as f:
 def read(fname):
     return codecs.open(os.path.join(base_dir, fname), encoding='utf-8').read()
 
+install_requires = [
+    'numpy',
+    'pipdate >=0.2.0, <0.3.0',
+    ]
+
+extras_require = {
+    'all': ['netCDF4', 'h5py'],
+    'exodus': ['netCDF4'],
+    'hdf5': ['h5py'],  # MED, MOAB, XDMF formats
+    }
 
 setup(
     name='meshio',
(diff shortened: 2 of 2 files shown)


nschloe:
I see this is listed in https://docs.pytest.org/en/latest/goodpractices.html but I'm not sure why you would want `python3 setup.py test` to work. Is this the some sort of default? Never used it anywhere.

leoschwarz:
If you want we can remove the alias if that is what bothers you (so that it would still be possible to execute the tests with `python3 setup.py pytest`)? Or are you against having the test runner in setup.py in general?

The alias would follow the usual convention of [setuptools](https://setuptools.readthedocs.io/en/latest/setuptools.html#test) test invocation. `pytest-runner` provides the magic to actually let you run the tests with `python3 setup.py pytest`.

In general it's not a big priority of mine to get this merged since there are enough virtual environment tools (I'm using Conda these days), but it was mainly meant as a way to do the testing and having the dependencies and everything be installed with a single command.

nschloe:
> Or are you against having the test runner in setup.py in general?

I'm not really sure. At first glance it seems to add some complication where it's not necessary. Just calling `pytest` instead of `python3 setup.py test` seems easy enough for me.

Well okay, I recognize it more canonical to have `python3 setup.py test`, so let's just get this in. Doesn't really matter. Could you adapt `.circleci` and `README.md` accordingly?

leoschwarz:
I just realized that it's a bit harder to have this integrated well with coverage generation (will need additional code, and one will have to be sure that no pitfalls are introduced in the future: see https://github.com/pytest-dev/pytest-cov/issues/147).

I'm actually ok leaving things as they are for now and closing this unmerged. (My main pain was installing the optional dependencies and having to manually change the PYTHONPATH or run `python3 setup.py install --force && pytest` in a virtual environment.)

Btw is the code really compatible with Python 2 like it says in the setup.py? I did not bother to check and the CI only checks Python 3, but I'm not sure my recent changes are compatible with Python 2 and whether they actually should be?

nschloe:
> My main pain was installing the optional dependencies

This should be handled by https://github.com/nschloe/meshio/pull/234.

> having to manually change the PYTHONPATH

Ah, yes. Well, that's what I do anyways, my `~/.bashrc` is full of `PYTHONPATH` additions. Hopefully this didn't cause you too much pain.

> I'm actually ok leaving things as they are for now and closing this unmerged. 

Let's do this. If this happens to be a problem in the future, we can always come back to it.

> Btw is the code really compatible with Python 2 like it says in the setup.py? 

Mostly yes, I think, but indeed it's untested. Let's maintain the claim until somebody comes along who has a problem. In a year and a half, Python 2 hits EOL anyways.

PR #238, opened 2018-05-11
medit_io: More permissive whitespace handling.

### LINKED ISSUE #237: Medit correct whitespace parsing
I'll look into fixing this bug in the next days, however just documenting it here already.

I'm experiencing this trying to parse `tetgen` output in MEDIT mesh format, meshio seems to be more strict in regards to not accepting line breaks between certain items unlike MEDIT.

Example `.mesh` file which currently fails to parse: https://gist.github.com/leoschwarz/5e6fc03941d6fddf9bbd57d45fe4ef84 (You can reproduce this by invoking tetgen with [this surface](https://gist.github.com/leoschwarz/3ef49c07bf9f82a00a9aa74c6760bb53) mesh if you like.)

According to the description in the MEDIT technical record (https://www.ljll.math.upmc.fr/frey/publications/RT-0253.pdf section 7.2):

> The blanks, ”newline” or <CR> and tabs are considered as item separators. A comment line starts with the character # and ends at the end of the line. The comments are placed exclusively between the fields.

Note it says "item seperator" and not "fields". Now I'm not 100% certain how this is meant (it's not very explicit with the format definition), but in the wild there seems to be the interpretation that this can mean 

```
Dimension
3
```

is also valid.

leoschwarz:
I am not sure if you like the approach of introducing a reader class, but I don't think it should cause any performance regressions since the number of intermediary lists created etc should be around the same.

This mostly fixes #237. But I realized that other than the whitespace part there is also the `Corners` field which is not understood by meshio currently. I'll open a different issue to track missing medit_io functionality, because I probably won't be able to get to it so soon. Should I send different pieces of functionality indepentenly or would you prefer one big PR with all the commits?

### DIFF
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -10,7 +10,7 @@ for something like a specification.
 
 .. moduleauthor:: Nico Schlömer <nico.schloemer@gmail.com>
 '''
-from itertools import islice
+import re
 import logging
 import numpy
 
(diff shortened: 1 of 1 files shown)


meshio/medit_io.py
```diff
@@ -22,56 +22,82 @@ def read(filename):
...
-            line = next(islice(f, 1))
-            num_verts = int(line)
+            # The first value is the number of nodes
+            num_verts = int(reader.next_item())
             points = numpy.empty((num_verts, dim), dtype=float)
-            for k, line in enumerate(islice(f, num_verts)):
+            for k in range(num_verts):
                 # Throw away the label immediately
-                points[k] = numpy.array(line.split(), dtype=float)[:-1]
+                points[k, :] = numpy.array(
```
nschloe: No `:` necessary.

meshio/medit_io.py
```diff
@@ -22,56 +22,82 @@ def read(filename):
...
+            # The first value is the number of elements
+            num_cells = int(reader.next_item())
             cell_data = numpy.empty((num_cells, num), dtype=int)
-            for k, line in enumerate(islice(f, num_cells)):
-                data = numpy.array(line.split(), dtype=int)
+            for k in range(num_cells):
+                data = numpy.array(reader.next_items(num + 1), dtype=int)
                 # Throw away the label
-                cell_data[k] = data[:-1]
+                cell_data[k, :] = data[:-1]
```
nschloe: No `:` necessary.

nschloe:
No, iterator is alright. Just some minor style fixes and this is ready to be merged.

PR #245, opened 2018-05-28
Gmsh: Support for periodicity

dalcinl:
This PR adds support for reading/writing `$Periodic` sections in Gmsh files. Note however than I'm stashing the periodic information in the field data entry named `gmsh:periodic`. I don't really like it, but I think we do not have a better mechanism yet to pass this stuff around.

### DIFF
--- a/meshio/gmsh_io.py
+++ b/meshio/gmsh_io.py
@@ -339,6 +339,32 @@ def _read_cells(f, cells, int_size, is_ascii):
     return has_additional_tag_data, output_cell_tags
 
 
+def _read_periodic(f):
+    periodic = []
+    num_periodic = int(f.readline().decode("utf-8"))
+    for _ in range(num_periodic):
+        line = f.readline().decode("utf-8")
+        edim, stag, mtag = [int(s) for s in line.split()]
+        line = f.readline().decode("utf-8").strip()
+        if line.startswith("Affine"):
+            transform = line
+            num_nodes = int(f.readline().decode("utf-8"))
+        else:
+            transform = None
+            num_nodes = int(line)
+        slave_master = []
+        for _ in range(num_nodes):
+            line = f.readline().decode("utf-8")
+            snode, mnode = [int(s) for s in line.split()]
+            slave_master.append([snode, mnode])
+        slave_master = numpy.array(slave_master, dtype=int).reshape(-1, 2)
+        slave_master -= 1  # Subtract one, Python is 0-based
+        periodic.append([edim, (stag, mtag), transform, slave_master])
+    line = f.readline().decode("utf-8")
+    assert line.strip() == "$EndPeriodic"
+    return periodic
+
+
 def _read_data(f, tag, data_dict, int_size, data_size, is_ascii):
     # Read string tags
     num_string_tags = int(f.readline().decode("utf-8"))
@@ -396,6 +422,7 @@ def read_buffer(f):
     cell_data_raw = {}
     cell_tags = {}
     point_data = {}
+    periodic = None
 
     is_ascii = None
     int_size = 4
@@ -418,6 +445,8 @@ def read_buffer(f):
             has_additional_tag_data, cell_tags = _read_cells(
                 f, cells, int_size, is_ascii
             )
+        elif environ == "Periodic":
+            periodic = _read_periodic(f)
         elif environ == "NodeData":
             _read_data(f, "NodeData", point_data, int_size, data_size, is_ascii)
         else:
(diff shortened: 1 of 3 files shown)


nschloe:
Still a pylint error, looks fine otherwise.

dalcinl:
@nschloe Now pylint is OK, but this PR as it is obviously make coverage go down. I bet you don't support periodicity in pygmsh, so one way to improve coverage would be to add some small msh file with a periodic section, though I don't like my own idea.

nschloe:
This PR needs some updating. Perhaps with the new `mesh` data type, periodic data must not be forced into `FieldData` anymore.

dalcinl:
Yes, sure, but the new `Mesh` type requires some more love. Are you sure `field_data` still makes sense there as a generic attribute? Why you added `node_sets` as a generic attribute (only used in exodus) ? And finally, there is no general mechanism  or "protocol" yet so we can pass around extra stuff like Gmsh periodicity. All these details have to be decided eventually. 

So, I'll start with  a quick proposal to help updating this PR:

We can define `Mesh` this way:

```python
class Mesh(object):
    def __init__(self, ..., **kwargs):
        ...
        self.__dict__.update(kwargs)
```

and then the Gmsh reader can do:

```
def read(...):
    return Mesh(..., gmsh_periodic=periodic)
```

Are you OK with that approach?

nschloe:
> Are you sure field_data still makes sense there as a generic attribute?

Not sure. Well, VTK and VTU have `field_data`, so there needs to be some attribute to host it.

> Why you added node_sets as a generic attribute (only used in exodus) ?

And Medit. Perhaps others as well.

> And finally, there is no general mechanism or "protocol" yet so we can pass around extra stuff like Gmsh periodicity.

The idea was to add an attribute in `Mesh` for every entity that is supported by a mesh type. Your `__dict__` suggestion essentially does the same, except that is uses less code. For now, I'd prefer the explicit approach, i.e., initializating all possible attributes in `Mesh` with `None`, including `gmsh_periodic`. This makes is easy for any developer to see what's already there. Using `__dict__`, it could happen that a PR for exodus calls node sets `node_sets`, and another for medit calls it `nodesets` such that there's no interoperability between the two.

If, in the future, we see an inflation of attributes, we can think about using a `__dict__`.

Agreed?

dalcinl:
No, I do not agree. Inflation has to be prevented, not dealt with later (trust me, I was born in Argentina!!). 

If `node_sets` is used by more than one format, then I agree on handling it as a generic attribute. Same thing for `field_data`, though I think the name of this attributes comes from VTK, and it is a terrible name, quite confusing. 

But for something that is very format-specific, like `gmsh_periodic`, handling that in the top-level mesh type is noisy and IMHO smells to bad design.

Perhaps a better alternative to consider is this one:

```python
class Mesh(object):
    def __init__(self, ..., **metadata):
        ...
        self.metadata = metadata
```

Of course, this is not a democracy and the final decision is on you. So could you please explicitly state your final decision about how  to expose the Gmsh periodic section information within the new `Mesh` type? Do you still want a new `gmsh_periodic=None` arg in the class initializer, and a  `self.gmsh_periodic` atribute in the instance?

nschloe:
> Inflation has to be prevented, not dealt with later.

If we see that, over the years, we accumulate 50 data attributes in `Mesh`, all with their own separate `thisandthat=None` as `__init__` argument, and we decide this is too messy, can't we simply use drop in one of the patterns you suggest? Would that change the API?

dalcinl:
This PR implements `gmsh_periodic` in such a way that you can roundtrip, that is, you can read and write, and get a new file that contains exactly the same as the original one. So my feeling is that `gmsh_periodic` is quite Gmsh-specific, so I would prefer to keep the `gmsh_` prefix in the attribute. Please note I have no experience with other mesh formats, do I do not know how they handle periodicity. I guess you don't know that either, otherwise `meshio` would have some support for it. 

So, to be in the safe side, let's call the thing `gmsh_periodic`, I'll add it as an arg to the constructor, and as an instance attribute, and get this PR updated. OK?

nschloe:
Sounds good to me. If the unlikely thing happens and another mesh format implements exactly the same thing, we might rename the variable, but really I don't see that happening anytime soon.

meshio/gmsh_io.py
```diff
@@ -418,11 +445,14 @@ def read_buffer(f):
...
             )
+        elif environ == "Periodic":
+            periodic = _read_periodic(f)
         elif environ == "NodeData":
             _read_data(f, "NodeData", point_data, int_size, data_size, is_ascii)
-        else:
-            assert environ == "ElementData", "Unknown environment '{}'.".format(environ)
+        elif environ == "ElementData":
             _read_data(f, "ElementData", cell_data_raw, int_size, data_size, is_ascii)
+        else:
```
nschloe: This line won't be tested, so let's go with the assertion.

PR #254, opened 2018-06-12
Add node sets read & write function

hillyuan:
1.  Add node sets data struct (dictionary) into class mesh.
2.  Add node sets raed/write function onto exodus_io
3.  Add test cases

### DIFF
--- a/meshio/exodus_io.py
+++ b/meshio/exodus_io.py
@@ -211,12 +211,14 @@ def write(filename, mesh):
         )
         for k, (name, data) in enumerate(mesh.point_data.items()):
             node_data[0, k] = data
-            
-    #node sets
+
+    # node sets
     num_nsets = len(mesh.nsets)
-    if num_nsets>0:
-        data = rootgrp.createVariable("ns_prop1", "i4", "num_node_sets") 
-        data_names = rootgrp.createVariable("ns_names", "S1", ("num_node_sets", "len_string"))
+    if num_nsets > 0:
+        data = rootgrp.createVariable("ns_prop1", "i4", "num_node_sets")
+        data_names = rootgrp.createVariable(
+            "ns_names", "S1", ("num_node_sets", "len_string")
+        )
         for k, name in enumerate(mesh.nsets.keys()):
             data[k] = k
             for i, letter in enumerate(name):
--- a/meshio/exodus_io.py
+++ b/meshio/exodus_io.py
@@ -80,6 +80,9 @@ def read(filename):
     point_data_names = []
     pd = []
     cells = {}
+    ns_names = []
+    ns = []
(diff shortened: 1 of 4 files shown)


meshio/mesh.py
```diff
@@ -4,12 +4,21 @@
...
 class Mesh(object):
-    def __init__(self, points, cells, point_data=None, cell_data=None, field_data=None):
+    def __init__(
+        self,
+        points,
+        cells,
+        point_data=None,
+        cell_data=None,
+        field_data=None,
+        nsets=None,
```
nschloe: Let's be more verbose here and call it `node_sets`.

hillyuan: OK. Modifing it!

meshio/exodus_io.py
```diff
@@ -80,6 +80,9 @@ def read(filename):
     point_data_names = []
     pd = []
     cells = {}
+    ns_names = []
```
nschloe: Same here: `ns_names` -> `node_set_names`.

meshio/exodus_io.py
```diff
@@ -80,6 +80,9 @@ def read(filename):
...
+    ns = []
```
nschloe: -> `node_sets`. Ah no wait, that name is already taken. What's the difference between the node set dictionary `{}` and this list here? Perhaps one of the two should get a different name.

meshio/exodus_io.py
```diff
@@ -202,5 +212,24 @@ def write(filename, mesh):
         for k, (name, data) in enumerate(mesh.point_data.items()):
             node_data[0, k] = data
 
+    # node sets
+    num_nsets = len(mesh.node_sets)
```
nschloe: `num_node_sets`

test/helpers.py
```diff
@@ -216,6 +216,12 @@ def add_field_data(mesh, value, dtype):
     return mesh2
 
 
+def add_node_sets(mesh):
+    mesh2 = copy.deepcopy(mesh)
+    mesh2.node_sets = {"nset": numpy.array([1, 2])}
```
nschloe: Does it have to be called `"nset"`. Perhaps something completely different would be appropriate, like `"metal"` or your favorite band.

nschloe:
Not all comments have been addressed yet.

meshio/exodus_io.py
hillyuan: you means the temporary name 'ns' ? I think it is not important just like you use 'pd' above as point_data. I just follows your naming style. Please modify its name as you like.

nschloe: Hehe, good point. When reviewing PRs I tend to be more catholic than the pope.

PR #266, opened 2018-06-29
Fix syntax

nilswagner:


nilswagner:
Exclamation marks define a comment in PERMAS. All subsequent characters in the line are ignored. Keywords begin with a $ sign.

Issue #280, opened 2018-08-27
MSH4


gdmcbain:
Gmsh 4.0.0 was released on Wed. 22, featuring ‘new MSH4 format’; see http://gmsh.info/CHANGELOG.txt:
> 4.0.0 (August 22, 2018): new C++, C, Python and Julia API; new MSH4 format

[MSH file format version 4](http://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-_0028version-4_0029) is generated by default in Gmsh 4, which is now the current stable release, but cannot be read with `meshio.read` which asserts that the MSH format is version 2. 

To reproduce, get say Gmsh's [tutorial/t1.geo](https://gitlab.onelab.info/gmsh/gmsh/blob/master/tutorial/t1.geo) and mesh it
```shell
gmsh -2 t1.geo
```
to produce t1.msh (attached here gzipped as [t1.msh.gz](https://github.com/nschloe/meshio/files/2322131/t1.msh.gz)).  Then
```shell
python -c 'import meshio; meshio.read("t1.msh")'
```
raises
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/users/geordie/.py3.7/skfem/lib/python3.7/site-packages/meshio/helpers.py", line 162, in read
    return format_to_reader[file_format].read(filename)
  File "/usr/local/users/geordie/.py3.7/skfem/lib/python3.7/site-packages/meshio/gmsh_io.py", line 165, in read
    mesh = read_buffer(f)
  File "/usr/local/users/geordie/.py3.7/skfem/lib/python3.7/site-packages/meshio/gmsh_io.py", line 452, in read_buffer
    data_size, is_ascii = _read_header(f, int_size)
  File "/usr/local/users/geordie/.py3.7/skfem/lib/python3.7/site-packages/meshio/gmsh_io.py", line 175, in _read_header
    assert str_list[0][0] == "2", "Need mesh format 2"
AssertionError: Need mesh format 2
```

gdmcbain:
`test/test_gmsh` doesn't pick this up because it uses `meshio.gmsh_io.write` which writes MSH2.

nschloe:
Geez, _another_ new mesh format out of gmsh?! Looking into it...

gdmcbain:
I know, that was my initial reaction too; however, I'm hoping that it might be faster to read.  Profiling revealed that reading the .msh file was the bottleneck in my last finite element application.

[...]

PR #282, opened 2018-08-27
shlex.split PhysicalNames to accept spaces

### LINKED ISSUE #281: PhysicalNames with spaces
While investigating #280 and its example based on the Gmsh [tutorial/t1.geo](https://gitlab.onelab.info/gmsh/gmsh/blob/master/tutorial/t1.geo), I noticed that its `PhysicalNames` are strings with spaces
```c++
Physical Curve("My second curve (automatic physical tag)") = {3} ;
Physical Curve("My third curve (physical tag 5)", 5) = {4} ;
Physical Surface("My surface") = {1} ;
```

Currently `meshio.read` has
```python
        key = line.split(" ")[2].replace('"', "").replace("\n", "")
```
which will read each key as "My".

gdmcbain:
This aims to fix #281.

I haven't been able to test it properly because I've got Gmsh 4 which outputs MSH4 #280.

The idea of using `shlex.split` instead of `str.split` is from https://stackoverflow.com/questions/79968/split-a-string-by-spaces-preserving-quoted-substrings-in-python.

### DIFF
--- a/meshio/gmsh_io.py
+++ b/meshio/gmsh_io.py
@@ -195,9 +195,7 @@ def _read_physical_names(f, field_data):
     for _ in range(num_phys_names):
         line = shlex.split(f.readline().decode("utf-8"))
         key = line[2]
-        phys_group = int(line[1])
-        phys_dim = int(line[0])
-        value = numpy.array([phys_group, phys_dim], dtype=int)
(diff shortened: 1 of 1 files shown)


nschloe:
Thanks!

Issue #284, opened 2018-08-27
binary MSH4 decoding


gdmcbain:
#283 fixes the minimal failing example of ASCII MSH4 given for #280 and generated by
```shell
gmsh -2 t1.geo
```
with [t1.geo](https://gitlab.onelab.info/gmsh/gmsh/blob/master/tutorial/t1.geo) being the first Gmsh tutorial, but the binary MSH4 format raises a different error.
```shell
gmsh -2 -bin t1.geo
```
Attached gzipped as [t1.msh.gz](https://github.com/nschloe/meshio/files/2325929/t1.msh.gz).  Then
```shell
python -c 'import meshio; meshio.read("t1.msh")'
```
raises
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/users/geordie/.py3.7/meshio/src/meshio/meshio/helpers.py", line 171, in read
    return format_to_reader[file_format].read(filename)
  File "/usr/local/users/geordie/.py3.7/meshio/src/meshio/meshio/msh_io/main.py", line 11, in read
    mesh = read_buffer(f)
  File "/usr/local/users/geordie/.py3.7/meshio/src/meshio/meshio/msh_io/main.py", line 30, in read_buffer
    return reader.read_buffer(f, is_ascii, int_size, data_size)
  File "/usr/local/users/geordie/.py3.7/meshio/src/meshio/meshio/msh_io/msh4.py", line 66, in read_buffer
    line = f.readline().decode("utf-8").strip()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9a in position 100: invalid start byte
```

Maybe it doesn't make sense to decode binary files?

gdmcbain:
The first two blocks (MeshFormat, PhysicalNames) of the binary file are parsed correctly.

The third, `Entities`, is ignored as if a `Comment`.  (Does it contain anything useful?  A separate question.)  But the contents of this block are binary and in skipping through the content to find the end of the block is where the `.decode("utf-8")` raises the `UnicodeDecodeError`.

gdmcbain:
Maybe it's safe enough to pass the UnicodeDecodeError inside a section like `Entities` which is being ignored?  152e54e

With that in place in branch msh4-binary, another UnicodeDecodeError is raised later by `_read_nodes`.

nschloe:
> Does it contain anything useful? A separate question.

I asked myself the same question. We'll find out in the future.

nschloe:
[...]

PR #286, opened 2018-08-28
WIP: read binary MSH4

gdmcbain:
WIP for fixing #284

### DIFF
--- a/meshio/msh_io/main.py
+++ b/meshio/msh_io/main.py
@@ -33,7 +33,7 @@ def read_buffer(f):
 def _read_header(f, int_size):
     line = f.readline().decode("utf-8")
     # Split the line
-    # 2.2 0 8
+    # version(double) binary(int) size(unsigned long)
     # into its components.
     str_list = list(filter(None, line.split()))
     format_version = str_list[0][0]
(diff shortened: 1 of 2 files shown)


gdmcbain:
The approach employed doesn't seem to be working.  Closer inspection of the binary format required.

gdmcbain:
As a quick indication of what wasn't working, firstly I seemed to have to treat `numNodes(unsigned long)` as if it were actually a C `int`.
```python
            line = f.read(len(struct.pack("4i", *[1]*4)))
            tag_entity, dim_entity, type_node, num_nodes = struct.unpack("4i", line)
```
Otherwise (using `"3iL"`), `num_nodes` ended up with values like 4294967296.  Not sure if this is to do with padding or what.

Secondly, with 
```python
                line = f.read(len(struct.pack("i3d", *[1]*(1+3))))
                tag, x, y, z = struct.unpack("i3d", line)
```
the co-ordinates agreed with those in an ASCII version of the file but the tags were weird, like (truncated)
```
array([          0,           0,           0,           0,           0,
                 6,  -320460155,  1067743969,           0,           0,
                 0,           0,           2,  1070721925,           0,
                 0,          19, -1717986918,  1069128089, -1383442275,
        1070176665,           0,           0,          27, -1717986918,
```

nschloe:
I more or less got it now. PR incoming.

Issue #175, opened 2017-12-21
implement regions


nschloe:
Some data formats have the concept of regions/submeshes/domains, e.g., gmsh and Exodus. Think about how to best integrate this into meshio.

gdmcbain:
Reading about [Exodus boundary conditions](https://cubit.sandia.gov/public/13.2/help_manual/WebHelp/finite_element_model/exodus/model_definitions.htm), I see that besides what it calls 'element blocks', corresponding to the above, it also has 'nodesets' and 'sidesets', the last two being useful for boundary and initial conditions, and 'sides' being sides of elements.
   I can immediately envisage 'nodesets' being useful for Dirichlet conditions.

gdmcbain:
Exodus Element Blocks are stricter than Gmsh Physical Regions; whereas the latter encode only the continuous concept of subdomain, the former mixes in the discrete notion of type of element.

> Element Blocks (also referred to as simply, Blocks) are a logical grouping of elements all having the same basic geometry and number of nodes. All elements within an Element Block are required to have the same element type. (ibid)

In a 'quad-dominant' mesh of a surface, this would necessitate two blocks, for the triangles and quadrilaterals.  I can see computational advantages to this: the connectivity would be described by a rectangular matrix rather than a list. But it also seems like an admixture of concerns if one just wants to specify that a subdomain is of one material.

cbcoutinho:
What would the computational advantage be of having the elemental group
data as matrix instead of a list? For one, a list (or list of lists) would
be much sparser, plus this information is already provided by meshio in the
form of dictionaries with list/array values, which I think is now provided
by the 'cells' and 'cell_data' objects.

gdmcbain:
Hmm. I was trying to second-guess why Exodus insisted on one type of element per block.
   In which case each element has the same number of nodes and the array with entry [e, n] the global index of the n-th node in the e-th element would be 100% dense and perfectly rectangular, which I thought (at the risk of prematurely optimizing) might be more efficient to allocate and look up.
  I didn't find an explicit statement of the motivation for the Exodus restriction yet, but thought it might be something like that.

gdmcbain:
As to the relation between this and what's currently in the meshio cell_data and field_data, I think they're like inverses, as discussed in the [final comments yesterday](https://github.com/nschloe/meshio/issues/169#issuecomment-353254001) on #169.

nschloe:
[...]

PR #289, opened 2018-08-29
WIP: harvesting physical tags from $Entities

gdmcbain:
This is a start, following the discussion in nschloe/pygmsh#209.  So far only for binary.

### DIFF
--- a/meshio/msh_io/msh4.py
+++ b/meshio/msh_io/msh4.py
@@ -53,7 +53,7 @@ def read_buffer(f, is_ascii, int_size, data_size):
         elif environ == "Periodic":
             periodic = _read_periodic(f)
         elif environ == "Entities":
-            _read_entities(f, is_ascii, int_size, data_size)
+            physical_tags = _read_entities(f, is_ascii, int_size, data_size)
         elif environ == "NodeData":
             _read_data(f, "NodeData", point_data, int_size, data_size, is_ascii)
         elif environ == "ElementData":
@@ -92,6 +92,7 @@ def read_buffer(f, is_ascii, int_size, data_size):
 
 def _read_entities(f, is_ascii, int_size, data_size):
     # More or less skip over them for now
+    physical_tags = [[], [], [], []]      # dims 0, 1, 2, 3
     if is_ascii:
         line = f.readline().decode("utf-8").strip()
         while line != "$EndEntities":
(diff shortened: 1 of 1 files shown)


nschloe:
Before this can go anywhere, we have to make a decision on how to use the tag groups in MSH4. I'm not sure about that yet. We used to simply create integer node/element data, but really that's quite ineffective. Perhaps we can do something that is common to MSH4 and Exodus which also has node/element sets. (And another mesh format, perhaps Medit?)

gdmcbain:
Excellent, yes. Let's use the arrival of MSH4 as an opportunity to revive the discussion of regions #175.

gdmcbain:
Pressing on with this, not with the intention of it going anywhere, but just to explore what MSH4 might giving us.  I hope this might help inform a continuation of #280 rather than prejudice it or a revived discussion of regions #175; for the latter, I've launched ‘tagged subsets of elements’ #290.

gdmcbain:
Applying this (the continuous integration failures are just `black`, so irrelevant as this code isn't destined for inclusion) to `t1`:
```python
[list(map(list, t)) for t in physical_tags]
```
gives
```python
[[[1], [1], [], []], 
 [[2], [2], [3], [5]], 
 [[6]], 
 []]
```
The indices 0, 1, 2, 3 of the outer list correspond to the dimension of the entities, i.e. points, curves, surfaces, volume. 

The values at index 1 (i.e. lines), `[[2], [2], [3], [5]]` correspond to those found for the same example nschloe/pygmsh#209.

nschloe:
Black update required.

gdmcbain:
Oh, sorry, this doesn't do much yet.  It's only work-in-progress.  It does read the `$Entities` and extract the physical tags but only stores them in a local unused variable (hence the `noqa F841` to keep `flake8` quiet); it doesn't return them in the `Mesh.cell_data`, &c.  I'd paused on this pending the decision on how to use the tag groups of MSH4, meanwhile working on the use of MEDIT mesh as an alternative.  I think it's harmless enough anyway.

nschloe:
Alright then, I agree. Let's just open another PR once this is ready.

gdmcbain:
Ah, I figured out that it might be possible to share a lot more of the code for the two cases of `bool is_ascii` (#280, #284).  Using the `sep` keyword of [`numpy.fromfile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.fromfile.html):
```python
from functools import partial
...
fromfile = partial(numpy.fromfile, sep=" " if is_ascii else "")
```

Then just use the same calls to `fromfile` in either case.

This will be handy for binary MEDIT mesh #295 too.

Issue #290, opened 2018-08-30
tagged subsets of elements


gdmcbain:
The arrival of MSH4 #280 with Gmsh 4 broke some applications downstream of meshio (e.g. nschloe/pygmsh#209, kinnala/scikit-fem#41) but also included innovation in the treatment of ‘entities’ and ‘physical tags’ which prompts a review of the discussion of how to implement tagged subsets of elements, a.k.a. 'regions' #175, which goes beyond this single format or format version.

While #280 was closed once meshio was able to read the nodes and elements #283 #287, and data defined thereon #288, `meshio.read` does not read the physical tags from MSH4 as it did from the now legacy version of Gmsh's native format, MSH2.  These physical tags belong to elements (of various dimension from zero up to that of the domain) and are used in applications to define subdomains (materials, switching terms), mixed boundary conditions, zones for measurements, etc.  For an elementary example in which the Laplace equation is solved in a plane region with boundary divided into two parts with specified different uniform potentials and the rest insulation, see [gh:kinnala/scikit-fem/examples/ex13.py](https://github.com/kinnala/scikit-fem/blob/master/examples/ex13.py).

As more installations of Gmsh are upgraded to v. 4, repeats of breakages like those of @dajuno in #37 

> Can I expect the converter to handle gmsh's physical groups?

@gpkc in #43 

> This is actually very important to gmsh users. It's a very straightforward way to set boundary conditions, for example. This should not be neglected probably.
 
@cbcoutinho  in #169 

> I use gmsh physical groups pretty extensively, so it was a surprise to me when all of my tests started failing with an updated version of meshio.

(although here it would be an updated Gmsh rather than meshio) or @dalcinl in #197 

>  A few months back, regions were encoded in output Gmsh files just fine. Now they are not, 

(although this was about `meshio.write` rather than `meshio.read`).

There was a lengthy discussion on how to implement regions in #175 which was closed with the tags going into `cell_data[type]['gmsh:physical']` in #199, but this may need revisiting; in reviewing some sketchy exploration of the physical tags in MSH4 #289 , @nschloe wrote:
> we have to make a decision on how to use the tag groups in MSH4. I'm not sure about that yet. We used to simply create integer node/element data, but really that's quite ineffective. Perhaps we can do something that is common to MSH4 and Exodus which also has node/element sets. (And another mesh format, perhaps Medit?)

This issue is launched to discuss how to use the tags in MSH4 in a way which is common to other formats.

gdmcbain:
[...]

PR #292, opened 2018-08-30
MeshVersionFormatted 1/2 is single/double precision MEDIT mesh

### LINKED ISSUE #291: MeshVersionFormatted 2 in MEDIT mesh
While looking at the MEDIT mesh format for #290 and nschloe/pygmsh#212, I found that `meshio.read` fails on the MEDIT mesh output by Gmsh (4.0.1-git-296dd75f0 here) because the header is
```
 MeshVersionFormatted 2
```
whereas `medit_io.read_buffer` expects version 1:
```python
        if keyword == "MeshVersionFormatted":
            assert reader.next_item() == "1"
```
which is consistent with Frey (2001, §7.2.1, as cited in #290): 
> The mesh file must start with the descriptor:
> `MeshVersionFormatted 1`

To reproduce the `AssertionError`:
```shell
gmsh -2 -format mesh -string Mesh.SaveElementTagType=2; t1.geo
python -c 'import meshio; meshio.read("t1.mesh")'
```
Apart from first line, the `t1.mesh` seems entirely consistent with Frey's specification.

gdmcbain:
This fixes #291.

### DIFF
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -117,7 +117,13 @@ def read_buffer(file):
 
 def write(filename, mesh):
     with open(filename, "wb") as fh:
-        fh.write(b"MeshVersionFormatted 2\n")
+        if mesh.points.dtype == c_float:
+            fh.write(b"MeshVersionFormatted 1\n")
+        elif mesh.points.dtype == c_double:
+            fh.write(b"MeshVersionFormatted 2\n")
+        else:
+            raise NotImplementedError("Unknown points.dtype '{}'.".format(
+                mesh.points.dtype))
         fh.write(b"# Created by meshio\n")
 
         n, d = mesh.points.shape
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -112,11 +112,8 @@ def read_buffer(file):
 
 def write(filename, mesh):
     with open(filename, "wb") as fh:
-        if mesh.points.dtype == c_float:
-            version = 1
-        else:
-            assert mesh.points.dtype == c_double
-            version = 2
+        version = {numpy.dtype(c_float): 1,
+                   numpy.dtype(c_double): 2}[mesh.points.dtype]
         # N. B.: PEP 461 Adding % formatting to bytes and bytearray
         fh.write(b"MeshVersionFormatted %d\n" % version)
 
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -6,6 +6,7 @@ I/O for Medit's format, cf.
 Check out
 <https://hal.inria.fr/inria-00069921/fr/>
 <https://www.ljll.math.upmc.fr/frey/publications/RT-0253.pdf>
+<https://www.math.u-bordeaux.fr/~dobrzyns/logiciels/RT-422/node58.html>
 for something like a specification.
 """
 from ctypes import c_float, c_double
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -89,8 +89,9 @@ def read_buffer(file):
             num_verts = int(reader.next_item())
             points = numpy.empty((num_verts, dim), dtype=dtype)
             for k in range(num_verts):
+                points[k] = numpy.array(reader.next_items(dim), dtype=dtype)
                 # Throw away the label immediately
-                points[k] = numpy.array(reader.next_items(dim + 1), dtype=dtype)[:-1]
+                _ = reader.next_items(1)
         elif keyword in meshio_from_medit:
             meshio_name, num = meshio_from_medit[keyword]
             # The first value is the number of elements
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -91,7 +91,7 @@ def read_buffer(file):
             for k in range(num_verts):
                 points[k] = numpy.array(reader.next_items(dim), dtype=dtype)
                 # Throw away the label immediately
-                _ = reader.next_items(1)
+                reader.next_items(1)
         elif keyword in meshio_from_medit:
             meshio_name, num = meshio_from_medit[keyword]
             # The first value is the number of elements
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -112,8 +112,7 @@ def read_buffer(file):
 
 def write(filename, mesh):
     with open(filename, "wb") as fh:
-        version = {numpy.dtype(c_float): 1,
-                   numpy.dtype(c_double): 2}[mesh.points.dtype]
+        version = {numpy.dtype(c_float): 1, numpy.dtype(c_double): 2}[mesh.points.dtype]
         # N. B.: PEP 461 Adding % formatting to bytes and bytearray
         fh.write(b"MeshVersionFormatted %d\n" % version)
 
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -19,9 +19,9 @@ from .mesh import Mesh
 
 def read(filename):
     with open(filename) as f:
-        mesh = read_buffer(f)
+        points, cells = read_buffer(f)
 
-    return mesh
+    return Mesh(points, cells)
 
 
 class _ItemReader:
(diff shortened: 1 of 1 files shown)


gdmcbain:
The test fails because with 
```
MeshVersionFormatted 1
```
it's reading `c_float` and getting coordinates differing in the _n_-th decimal place
```
> tri_mesh.points
array([[0.        , 0.        , 0.        ],
       [0.33333333, 0.        , 0.        ],
       [0.33333333, 0.33333333, 0.        ],
       [0.        , 0.33333333, 0.        ]]))
> out_mesh.points
array([[0.        , 0.        , 0.        ],
       [0.33333334, 0.        , 0.        ],
       [0.33333334, 0.33333334, 0.        ],
       [0.        , 0.33333334, 0.        ]], dtype=float32)
```
which fails `test.helpers.generic_io` which wants
```python
   (abs(out_mesh.points - tri_mesh.points) < 1.0e-15).all()
```

I think this is reasonably gotten around by writing `MeshVersionFormatted 2` so that it's read back with `dtype=c_double`.

gdmcbain:
What on Earth does one call the keys in `point_data`, `cell_data['line']`, `cell_data['triangle']`, &c.?  I guess this gets back to #175 and #290.

Here I've tried to mimic current (or rather recent, pre #280) behaviour of meshio and used "gmsh:physical" for `cell_data` despite its absurdity here and then "medit:ref" for `point_data`.   I was never entirely comfortable with `"gmsh:physical"` but #175 had gone on long enough by the time it was closed with #199; hopefully something will come out of #290.

meshio/medit_io.py
```diff
@@ -68,47 +80,44 @@ def read_buffer(file):
...
             assert keyword == "End", "Unknown keyword '{}'.".format(keyword)
 
-    return points, cells
+    return Mesh(points, cells, point_data=point_data, cell_data=cell_data)
 
 
 def write(filename, mesh):
     with open(filename, "wb") as fh:
-        fh.write(b"MeshVersionFormatted 1\n")
+        fh.write(b"MeshVersionFormatted 2\n")
```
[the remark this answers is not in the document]
gdmcbain: You're completely right.  Please excuse my haste.  I'll look up how [`numpy.dtype`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.html) corresponds to "single" and "double" precision; though I'm a bit scared that the latter aren't terribly well defined.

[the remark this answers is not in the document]
nschloe: You can just use `mesh.points.dtype == c_double/c_float`, I think. And `assert` that it's either one of them. (And perhaps assert that all float data arrays are of the same type, but that may be too strict.)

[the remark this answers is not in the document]
gdmcbain:
```
       if mesh.points.dtype == c_float:
            fh.write(b"MeshVersionFormatted 1\n")
        elif mesh.points.dtype == c_double:
            fh.write(b"MeshVersionFormatted 2\n")
        else:
            raise NotImplementedError("Unknown points.dtype '{}'.".format(
                mesh.points.dtype))
```
a9e39b226ac9ef55fa8536da57c63688a5bcb41d

meshio/medit_io.py
```diff
@@ -68,47 +80,50 @@ def read_buffer(file):
...
 
 def write(filename, mesh):
     with open(filename, "wb") as fh:
-        fh.write(b"MeshVersionFormatted 1\n")
+        if mesh.points.dtype == c_float:
+            fh.write(b"MeshVersionFormatted 1\n")
+        elif mesh.points.dtype == c_double:
+            fh.write(b"MeshVersionFormatted 2\n")
+        else:
+            raise NotImplementedError("Unknown points.dtype '{}'.".format(
```
nschloe: Will never be tested. I prefer
```
if  mesh.points.dtype == c_float:
    # ...
else:
   assert mesh.points.dtype == c_double, "Unknown points.dtype '{}'.format(...)
```

gdmcbain: I liked the dict-lookup but no:
```
                     {c_float: 1,
>                     c_double: 2}[mesh.points.dtype])
E           KeyError: dtype('float64')
```
so I went with the `if`-`else`; I guess getitem uses a different kind of equality than `==`.

meshio/medit_io.py
```diff
@@ -118,13 +118,11 @@ def read_buffer(file):
 def write(filename, mesh):
     with open(filename, "wb") as fh:
         if mesh.points.dtype == c_float:
```
nschloe: This `if`-`else` could really be condensed into `{c_float: 1, c_double: 2}[mesh.points.dtype]`. No need for an assertion then either.

meshio/medit_io.py
```diff
@@ -118,13 +118,11 @@ def read_buffer(file):
...
-        elif mesh.points.dtype == c_double:
-            fh.write(b"MeshVersionFormatted 2\n")
+            version = 1
         else:
-            raise NotImplementedError(
-                "Unknown points.dtype '{}'.".format(mesh.points.dtype)
-            )
+            assert mesh.points.dtype == c_double
+            version = 2
+        fh.write(b"MeshVersionFormatted %d\n" % version)
```
nschloe: New-style `format` please.

nschloe:
So point data and cell data tests remain disabled for now because there's no writer yet, right?

meshio/medit_io.py
gdmcbain: Hmm, no: `AttributeError: 'bytes' object has no attribute 'format'`.

The makeshift at https://stackoverflow.com/questions/15710515/python-3-bytes-formatting is reverting to old-style `%`.

meshio/medit_io.py
gdmcbain: I thought so too but ([above](https://github.com/nschloe/meshio/pull/292#discussion_r213922359)) hit a KeyError with `dtype('float64')`.  Tricky as 
```python
numpy.dtype('float64') == ctypes.c_double
```
but
```python
numpy.dtype('float64') not in {ctypes.c_double: None}
```
(I'm still scratching my head about that one.)

gdmcbain:
Ah, no, that would be because I didn't know about them…  Sorry.  Tomorrow.

meshio/medit_io.py
nschloe: That's really weird! Okay, best add a comment explaining that, otherwise I see myself bumping into the same issue in while when trying to "fix" it.

meshio/medit_io.py
nschloe: Nice. Best add a link to that question in the code.

gdmcbain:
Ha, yes; the sun never sets on meshio. :)

meshio/medit_io.py
gdmcbain: Oh, I see that there's another idiom already in use immediately following:
```
        dim = "\nDimension {}\n".format(d)
        fh.write(dim.encode("utf-8"))
```

I'll adopt that for consistency.

By the way, is all this `open(…, "wb")`, `.write(b"…")`, `.write(_.encode("utf-8"))` stuff still needed in Python 3?  I note that `meshio.medit_io` only deals with text MEDIT mesh thus far, not binary #295.

meshio/medit_io.py
gdmcbain: Weirder:
```python
numpy.dtype(ctypes.c_double) in [ctypes.c_double]
```
but
```python
numpy.dtype(ctypes.c_double) not in {ctypes.c_double: 2}
```

gdmcbain: Your suggestion
```python
{c_float: 1, c_double: 2}[mesh.points.dtype]  # surprisingly fails
```
does work if the keys are wrapped with `dtype`:
```python
        version = {numpy.dtype(c_float): 1,
                   numpy.dtype(c_double): 2}[mesh.points.dtype]
```
That seems clear.  9dcf64bd5081be0dd37bf33731ecc25b8a95fe24

nschloe:
Good job.

Issue #296, opened 2018-08-31
read nodal and elemental tags from MEDIT mesh


gdmcbain:
The MEDIT mesh format has exactly one `int` tag for each of its "Vertices" and for each of its elements ("Edges", "Triangles", &c.).  These aren't currently read by `meshio.medit_io.read`.   

The elemental tags in particular are written by Gmsh (`-format mesh`) and reflect its "physical" tags with [option](http://gmsh.info/doc/texinfo/gmsh.html#index-Mesh_002eSaveElementTagType) 
```C++
Mesh.SaveElementTagType=2
```

This then would answer the question about tagged subsets of elements #290 of @nschloe in #289 about further work on decoding MSH4 #280:
> Before this can go anywhere, we have to make a decision on how to use the tag groups in MSH4. I'm not sure about that yet. We used to simply create integer node/element data, but really that's quite ineffective. Perhaps we can do something that is common to MSH4 and Exodus which also has node/element sets. (And another mesh format, perhaps Medit?)

If `meshio.read` read the elemental tags, it would make MEDIT mesh a competitive format for `pygmsh.generate_mesh` nschloe/pygmsh#212, particularly as, there being exactly one `int` tag, it should be possible to read all the elements of each type at once rather than line by line.

A start was made on this in bbc886c in #292 but the latter turns out to have enough complexity of its own so it was reverted.

gdmcbain:
A comment from #292:

> What on Earth does one call the keys in `point_data`, `cell_data['line']`, `cell_data['triangle']`, &c.? I guess this gets back to #175 and #290.

> Here I've tried to mimic current (or rather recent, pre #280) behaviour of `meshio` and used `"gmsh:physical"` for `cell_data` despite its absurdity here and then `"medit:ref"` for `point_data`. I was never entirely comfortable with `"gmsh:physical"` but #175 had gone on long enough by the time it was closed with #199; hopefully something will come out of #290.

nschloe:
> A start was made on this in bbc886c in #292 but the latter turns out to have enough complexity of its own so it was reverted.

So will you open another PR for this?

gdmcbain:
I think that this is fixed by #297.

nschloe:
Alright, let's close it then.

Issue #279, opened 2018-08-26
medit does not properly write physical groups / labels


izaid:
Hello,

This is a great library, I'm very much enjoying it. I just have a comment about the medit (.mesh) reader / writer.

It seems that even if my mesh has physical groups, the medit mesh will always be written with label=1. See https://github.com/nschloe/meshio/blob/master/meshio/medit_io.py#L148 for instance.

Here is an example of generating a mesh from pygmsh with two physical groups. I'd expect to get two labels in the .mesh file, but instead I only get one:

```python
import pygmsh as pg

geom = pg.built_in.Geometry()
scatterer = geom.add_ball((0, 0, 0), 1)
box = geom.add_box(-10, 10, -10, 10, -10, 10, holes = [scatterer.surface_loop])

geom.add_physical_volume([scatterer.volume])
geom.add_physical_volume([box.volume])

points, cells, point_data, cell_data, field_data = pg.generate_mesh(geom)

import meshio

mesh = meshio.Mesh(points, cells, point_data = point_data, cell_data=cell_data, field_data = field_data)
meshio.write('test.mesh', mesh)
```

gdmcbain:
I've just noticed this while working on MEDIT mesh for #291, #292, and nschloe/pygmsh#212.  I'm now adding the reading of the point and cell data from that format, I'll try and follow up with writing too.

gdmcbain:
The above example may be simplified from the real application, but if not one would probably not `generate_mesh` but rather write the GEO file from `geom.get_code()` and let Gmsh generate the MEDIT mesh file itself (`-format mesh -string "Mesh.SaveElementTagType=2;").

gdmcbain:
I believe that this is fixed by #297 though I can't rerun the example in the original post as `pygmsh.generate_mesh` is broken for Gmsh 4 nschloe/pygmsh#209 pending reading & writing of MSH4 #280.  I think it should work with Gmsh &leqslant; 3.

nschloe:
Let's tentatively close this; feel free to reopen if the merged fix doesn't work for you.

augustjohansson:
Hi! Thank you very much for your great work on meshio. I have an issue that's, what I think, identical to the original submission of user izaid: Physical labels doesn't seem to be written to the medit mesh file. I've tried using both gmsh 3.0.6 and 4.4.1 but I get identical output. Please let me know if you need more info!

Here's a small example where the unit square is split into two triangles with two different labels:
```
import pygmsh
import meshio

[...]

PR #297, opened 2018-08-31
WIP: minimally harvest MEDIT mesh ref

gdmcbain:
This fulfils #296 with minimal changes.

TODO:

- Testing.

- What to call the elemental tags; i.e. not "gmsh:physical".

- Try out for real in nschloe/pygmsh#212

- Read each type of element as a block rather than linewise.

### DIFF
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -99,9 +99,7 @@ def read_buffer(file):
             meshio_name, num = meshio_from_medit[keyword]
             # The first value is the number of elements
             num_cells = int(reader.next_item())
-            cell_data[meshio_name] = {
-                "medit:ref": numpy.empty(num_cells, dtype=int)
-            }
+            cell_data[meshio_name] = {"medit:ref": numpy.empty(num_cells, dtype=int)}
             cells1 = numpy.empty((num_cells, num), dtype=int)
             for k in range(num_cells):
                 data = numpy.array(reader.next_items(num + 1), dtype=int)
--- a/meshio/medit_io.py
+++ b/meshio/medit_io.py
@@ -67,7 +67,7 @@ def read_buffer(file):
         "Triangles": ("triangle", 3),
         "Quadrilaterals": ("quad", 4),
         "Tetrahedra": ("tetra", 4),
-        "Hexahedra": ("hexahedra", 8),
+        "Hexahedra": ("hexahedron", 8),
     }
 
     reader = _ItemReader(file)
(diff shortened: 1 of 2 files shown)


gdmcbain:
On testing.  In #292, @nschloe [asked](https://github.com/nschloe/meshio/pull/292#issuecomment-417213007):
> So point data and cell data tests remain disabled for now because there's no writer yet, right?

On looking at `helpers.add_point_data` and `add_cell_data`, I realize that they're inappropriate for MEDIT mesh.  This format does not hold point or cell data.  It does have an `int` tag for each point and cell, but I'm pretty sure that it would be an unidiomatic abuse to put floating-point data in there.

Instead, for MEDIT, according to Frey (2001, §7.2.4 ‘The `bb` format’), it is stored in a separate auxiliary file:

> This ASCII file contains scalar/tensor values associated with mesh entities. This file (`xxx.bb`) has
the following structure :
> _dim, nbmet, nbval, type_
> <i>v<sup>i</sup>, i</i> ∈ [1¸…, _nbmet, nbval_]
 
Reading and writing those does sound like a worthwhile future project but it does look quite distinct from the handling of regions #290, even though since #175 regions have been implemented in meshio sort of as if they were the same kind of thing as node or cell data.

I think it would be a mistake to read a MEDIT `mesh` file and extract the tags and then write them back out in a MEDIT `bb` file.

gdmcbain:
Dobrzynski (§A.0.0.1.2 ‘`SOL` file‘)[’ https://www.math.u-bordeaux.fr/~dobrzyns/logiciels/RT-422/node60.html] describes a different file format to the `bb`file.

nschloe:
> I think it would be a mistake to read a MEDIT mesh file and extract the tags and then write them back out in a MEDIT bb file.

Okay, sure. What we do in MSH2 for geometrical and physical tags is that the (`int`!) data gets an appropriate name that, when detected by the writer, will treat it differently than the other node/element data. We could do the same here.

gdmcbain:
I think that this is ready now, reading nodal and elemental _ref_, storing them as `point_data["medit:ref"]` and `cell_data[type]["medit:ref"]`, and writing them.  This thus fixes #296 and #279.

Of the items in the TODO at the top:

- This passes the original `test_medit`.

  - With `hex_mesh` added, though that was orthogonal to #296.

  - It doesn't attempt to add `add_point_data` or `add_cell_data` from `helpers` as floating-point data is handled quite differently to tags in MEDIT.  A new issue should be raised for this if required.

 - It doesn't attempt to test the new functionality as there aren't any tests for tagged subsets of elements yet.  Should a new test be added for this, like `test_gmsh.gmsh_periodic`?  (I note that the same feature for Gmsh, "gmsh:physical", isn't tested either.)

- Use of this for nschloe/pygmsh#212 to follow, not affecting this package.

- Reading each block of elements as a block should be a separate issue, as in #291 


Also, it doesn't attempt to preserve physical tags during conversion between Gmsh MSH2 and MEDIT mesh in either direction.  I think that this would be reasonably straightforward, but it depends on the mesh-independent representation of tagged subsets of elements #290.

nschloe:
Looks good to me.

Issue #294, opened 2018-08-30
Force embedding into 2D space


michalhabera:
See https://github.com/FEniCS/dolfinx/issues/159.

An option to force convert 3D mesh to 2D mesh (by stripping last component away) would be very useful. 
Some users are preparing planar (2D) meshes in `gmsh`, which embeds into 3D. Reading 3D mesh into finite element library results in larger system matrices and suboptimal computations for tensors of `rank >= 1`.

mrehor:
Well, I'm afraid that you're already messing with that a bit when reading xdmf3 mesh, see https://github.com/nschloe/meshio/blob/ecd790a2fbd3d8bb5c7048b5920dc1caf929be00/meshio/xdmf_io.py#L353-L354

But this is good. Each mesh that is potentially saved _as planar_ can be read again by meshio. I've created the pull request  #298 with a small fix adding the option `save_as_planar_mesh` to `meshio.xdmf_io.XdmfWriter`.

nschloe:
Alright now. I've looked at this issue again and found that meshio didn't properly handle 2D meshes at all. @mrehor gave a good example. The original reason for this was that some mesh formats (like VTK) don't support 2D meshes at all.

Now, I/O is handled properly (https://github.com/nschloe/meshio/pull/301) for those formats that support it (like XDMF). There also is a command-line option for `meshio-convert` that allows pruning the z-components if it's all 0. This should hopefully satisfy your use case.

mrehor:
Thanks for fixing this @nschloe, works as expected.

PR #298, opened 2018-08-31
Add possibility to save 2D meshes in XDMF (avoid embedding into 3D)

mrehor:
Users can opt to save a mesh in XDMF as being truly planar, i.e. with `GeometryType="XY"` and thus with geometric dimension `== 2`. Such mesh can be safely re-read through https://github.com/nschloe/meshio/blob/ecd790a2fbd3d8bb5c7048b5920dc1caf929be00/meshio/xdmf_io.py#L303 

See #294 and https://github.com/FEniCS/dolfinx/issues/159

### DIFF
--- a/meshio/xdmf_io.py
+++ b/meshio/xdmf_io.py
@@ -385,9 +385,9 @@ class XdmfReader(object):
 
 
 class XdmfWriter(object):
-    def __init__(self, filename, mesh,
-                 pretty_xml=True, data_format="HDF",
-                 save_as_planar_mesh=False):
+    def __init__(
+        self, filename, mesh, pretty_xml=True, data_format="HDF", save_as_planar=False
+    ):
         from lxml import etree as ET
 
         assert data_format in ["XML", "Binary", "HDF"], (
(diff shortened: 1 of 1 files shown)


nschloe:
See https://github.com/nschloe/meshio/issues/294#issuecomment-418289703.

mrehor:
@marazzaf Nope, the option `save_as_planar` hasn't been included. But you can use the option the following command-line option for `meshio-convert`:

`  --prune-z-0, -z       remove third (z) dimension if all points are 0`

See https://github.com/nschloe/meshio/issues/294#issuecomment-418289703

marazzaf:
> @marazzaf Nope, the option `save_as_planar` hasn't been included. But you can use the following command-line option for `meshio-convert` instead:
> 
> ` --prune-z-0, -z remove third (z) dimension if all points are 0`
> 
> See [#294 (comment)](https://github.com/nschloe/meshio/issues/294#issuecomment-418289703)

Thanks for the tip. I used:
`points=mesh.points[:, :2]`
in `meshio.Mesh`
and got a 2d mesh which is fine by me.

PR #301, opened 2018-09-04
properly handle two-dimensional meshes

nschloe:


meshio/cli.py
```diff
@@ -25,6 +25,13 @@ def main(argv=None):
     if args.prune:
         mesh.prune()
 
+    if (
+        args.prune_z_0
+        and mesh.points.shape[1] == 3
+        and numpy.all(numpy.abs(mesh.points[:, 2]) < 1.0e-13)
```
michalhabera: Aren't these type of assertions unnecessarily expensive? I don't know how `meshio` is optimised wrt. big meshes, but could be worth to sacrifice sanity checks for performance.

nschloe: Yeah, a little bit. I don't expect this to have any impact though. If one day somebody comes along with a runtime profile where this takes a visible share, I'll think about updating it.

PR #304, opened 2018-09-07
read physical tags from MSH4

gdmcbain:
This extends the reading of MSH4 #280, including binary MSH4 #284, in particular focusing on harvesting the physical tags from the `$Entities` section (begun in #289).

### DIFF
--- a/meshio/msh_io/main.py
+++ b/meshio/msh_io/main.py
@@ -14,8 +14,8 @@ def read(filename):
 
 
 def read_buffer(f):
-    # The format is specified at
-    # <http://gmsh.info//doc/texinfo/gmsh.html#MSH-ASCII-file-format>.
+    # The various versions of the format are specified at
+    # <http://gmsh.info//doc/texinfo/gmsh.html#File-formats>.
 
     line = f.readline().decode("utf-8")
     assert line.strip() == "$MeshFormat"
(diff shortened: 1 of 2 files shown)


nschloe:
Can we get tests for this?

gdmcbain:
In the test suite?  Hmm.  That would be good...  Like some of the tiny meshes in `test.helpers`?  Yes, that would be good.  Probably have to wait till next week.

Meanwhile, I've tested against:

* the examples from kinnala/scikit-fem#40
* nschloe/pygmsh#209, specifically `test/test_booleans`, after reverting from MEDIT mesh to the default (in Gmsh 4) MSH4
* the output of `gmsh -2 t1.geo`, with and without `-bin`

Note that this does not implement `"gmsh:geometrical"`, which seems to have been dumped.  I doubt anyone used it much; I never did.

gdmcbain:
Actually no, such a test wouldn't work. That would require writing physical tags in MSH4. This is only for reading them. This is to fix `pygmsh.generate_mesh` nschloe/meshio#209 which only needs `meshio.read`, not `meshio.write`. The MSH4 is written by Gmsh 4, not meshio.

`msh4.write` currently doesn't write an `$Entities` section. Is that valid?

I think that the relevant test for this PR is `test_booleans` in pygmsh.  This was the test for the equivalent functionality in Gmsh <4 writing MSH2.

nschloe:
> what about making use of a gmsh -convert subprocess

Good idea for local testing. For writing, let's just roll our own. It'll come to meshio sooner or later.

PR #308, opened 2018-09-20
Small interface tweaks

blechta:
* Accept `write(..., file_format='xmdf[3]-(binary|hdf|xml)')`
* Pass kwargs through `write_points_cells`, `write`  to the specific writer


nschloe:
Nice one, thank you.

PR #319, opened 2018-12-03
Backward compatibility: make Mesh class iterable and yield points, cells, point_data, cell_data, field_data

MuellerSeb:
With this little hack, you could use the `mesio.read` as before:
`points, cells, point_data, cell_data, field_data = meshio.read(infile)`

Maybe this could solve some incompatibilities in some other projects depending on meshio.

What do you think?

### DIFF
--- a/meshio/mesh.py
+++ b/meshio/mesh.py
@@ -44,6 +44,12 @@ class Mesh(object):
 
         return "\n".join(lines)
 
+    def __iter__(self):
+        for value in [
+            self.points, self.cells, self.point_data, self.cell_data, self.field_data,
+        ]:
+            yield value
+
     def prune(self):
         prune_list = []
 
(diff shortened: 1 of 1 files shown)


gdmcbain:
Against this, part of the original motivation for `class Mesh` was (https://github.com/nschloe/meshio/issues/175#issuecomment-361697523):

> Also, good idea about the container type. This would allow us to add more data without breaking the API (as we have to do now).

nschloe:
Yes, I think Geordie is right.
When switching to the container type, we bumped the major version to 2 which is exactly the indication that user code needs to change. If the user can't do it, she can just stick with meshio 1.

The problem with introducing the iterator functionality as in this PR is that, if we remove it in the future, we'll have to bump the major version _again_. And that'd be overkill.

Bottom line: if you need backwards compatibility, just stick to the same major version.

MuellerSeb:
Of course, I totally agree with that. Maybe this would have helped with a version in between and a future warning cycle. I will just update my projects.

Issue #322, opened 2019-01-09
read/write polygons


eolianoe:
Is there a way to read/write meshes with polygons?
For instance, [this file](https://github.com/nschloe/meshio/files/2740508/mesh.txt) is a dodecahedron with pentagonal faces. It seems that this kind of mesh can be saved only in a VTK file using VTK_POLYGON but the support of polygons is disabled in meshio.

nschloe:
For now, meshio is only for unstructured meshes. I think there's no other format that supports polygons like the legacy VTK does. Anyways, if you want to add support for reading, I'm open to pull requests.

eolianoe:
Thanks for the answer. I'll look into the magic of the `translate_cells` methods but my python is quite new so maybe it will be easier to write something independant to read such files.

nschloe:
IF you do, feel free to  close the issue request.

nschloe:
Closing for lack of feedback. Feel free to reopen at any time.

PR #323, opened 2019-01-10
WIP: support for reading VTK_POLYGON

eolianoe:
Following #322, here is a first draft to read VTK files with polygons inside. The current code does not support cells with a custom number of points and the "easy" way to support it, is to loop over the cells the fill the `numnodes` and `offsets` arrays.

The `cell_data` variable is not filled yet as I'm looking for a nice solution.


nschloe:
Looks interesting!

It'd be a good idea to add writing for polygons as well, so that this part of the code can be tested like the rest.

Small nitpick: Logical variables (like `polygon` in your draft) should have names indicating the Booleanness, e.g., `is_polygon` or `has_polygon`.

eolianoe:
The polygons can now be written, but cell_data is not done as the dictionnay of dictionnaries is a bit strange for me now.

nschloe:
I'm backpacking right now, but will look at it once I'm back.

eolianoe:
You closed #322, but I didn't get any feedback after your trip. Can I have your opinion on this PR?

nschloe:
Sorry, missed this one! Looking at it now.

nschloe:
Okay, so I think this is a valuable addition and your coding skill + familiarity with meshio is good enough for us to get somewhere.

What I don't like about the approach is that there's basically a separate `if` branch whenever polygons are involved. Perhaps there's no way around this because polygons can have any number of nodes, so this may be okay.

What we definitely need here are tests. Tests in meshio work in such a way that they first _write_ some data to a file, _read_ it afterwards, and the output data is then compared with the one we started with. So the next thing we need is a support for polygon writing here. Do you think you can do it? If yes, I could add some tests. After that, we could fix the code style here and there (nothing major) and merge. What do you think?

eolianoe:
The only way I see to remove the `if` is to use a list of list rather than an array of list to store the connectivity but it may break the others readers/writers.

If I'm not mistaken, the writer is working fine with the current code. For the tests, is it better to write one polygon or a set of polygons and triangles/quads ?

No problem for the code style :)

meshio/helpers.py
```diff
@@ -205,7 +205,8 @@ def write(filename, mesh, file_format=None, **kwargs):
 
     # check cells for sanity
     for key, value in mesh.cells.items():
-        assert value.shape[1] == num_nodes_per_cell[key]
+        if "polygon" not in key:
```
nschloe: This should be `key != "polygon"` to make sure the rest is tested.

meshio/vtk_io.py
```diff
@@ -359,25 +359,59 @@ def translate_cells(data, types, cell_data_raw):
...
+    has_polygon = meshio_to_vtk_type["polygon"] in bins
 
     # Deduct offsets from the cell types. This is much faster than manually
     # going through the data array. Slight disadvantage: This doesn't work for
     # cells with a custom number of points.
     numnodes = numpy.empty(len(types), dtype=int)
-    for tpe, idx in bins.items():
-        numnodes[idx] = vtk_type_to_numnodes[tpe]
-    offsets = numpy.cumsum(numnodes + 1) - (numnodes + 1)
+    if not has_polygon:
```
nschloe: Negated `if`s with an `else`, e.g.,
```
if not hungry:
    don't eat
else:
    eat
```
are harder to read than
```
if hungry:
    eat
else:
    don't eat
```
Please adapt all conditionals in the PR accordingly.

meshio/vtk_io.py
```diff
@@ -359,25 +359,59 @@ def translate_cells(data, types, cell_data_raw):
...
+        for tpe, idx in bins.items():
+            numnodes[idx] = vtk_type_to_numnodes[tpe]
+        offsets = numpy.cumsum(numnodes + 1) - (numnodes + 1)
+    else:
+        # If some polygons are in the VTK file, loop over the cells
+        nbcells = len(types)
+        offsets = numpy.empty(len(types), dtype=int)
+        offsets[0] = 0
+        for idx in numpy.arange(nbcells - 1):
```
nschloe: `numpy.arange` -> `range` (everywhere)

nschloe: We should also think about the possibility of removing this loop. It's gonna be super slow.

meshio/vtk_io.py
```diff
@@ -472,16 +506,27 @@ def _write_cells(f, cells, write_binary):
...
         for key in cells:
-            d = numpy.full(len(cells[key]), meshio_to_vtk_type[key]).astype(
-                numpy.dtype(">i4")
-            )
+            if "polygon" not in key:
+                d = numpy.full(len(cells[key]), meshio_to_vtk_type[key]).astype(
+                    numpy.dtype(">i4")
+                )
+            else:
+                d = numpy.full(len(cells[key]), meshio_to_vtk_type["polygon"]).astype(
```
nschloe: This looks like it's not going to work if both `polygon` and other cells are in the mesh.

meshio/helpers.py
eolianoe: Yes, but I do not want the rest to be tested as, for now, the key is "polygonN" with N the number of vertices/edges

meshio/vtk_io.py
eolianoe: why? It seems to work with my first tests

meshio/helpers.py
nschloe: If the key `polygonN`, then this conditional is always true. I think I misunderstand something here.

meshio/vtk_io.py
nschloe: If the `"polygon"` key _is_ in `cells` (along with other keys), then you always pick `meshio_to_vtk_type["polygon"]`. That's the part that looks suspicious to me.

eolianoe: Nope, there is loop over the keys and the test is made inside the loop

meshio/helpers.py
eolianoe: Yes, that's the point all polygon have the same VTK_TYPE

nschloe: Ah, okay. For clarity, this should be replaced by `key[:7] == "polygon"`.

eolianoe:
Here is a file to test the current code, it writes a pentagonal prism with some quadtrilaterals and triangles faces, then read it.
```
#!/usr/bin/env python3

import meshio
import numpy as np
import math

DIM = 3


def readMesh(filename):
    mesh = meshio.read(filename)
    points = mesh.points
    elements = mesh.cells
    return points, elements


def prism(n):
    h = 1.0
    radius = 0.5/np.sin(np.pi/n)

    theta = np.linspace(0.0, 2.0*np.pi, n+1)
    verts = np.zeros(shape=(2*n, DIM))
    for i in np.arange(0, n):
        verts[i, 0] = radius*math.cos(theta[i])
        verts[i, 1] = radius*math.sin(theta[i])
        verts[i, 2] = 0.
        verts[n+i, 0] = radius*math.cos(theta[i])
        verts[n+i, 1] = radius*math.sin(theta[i])
        verts[n+i, 2] = h

    faces = []
    polygonalFaces = np.zeros(shape=(2, n), dtype=int)
    for i in np.arange(0, n):
        faces.append([i, (i+1) % n, (i+1) % n+n, i+n])
        polygonalFaces[0][i] = i
        polygonalFaces[1][i] = n+i
    faces.append(polygonalFaces[0])
    faces.append(polygonalFaces[1])

    return verts, faces


def main():
    cells = {}
    i = 5
    filename = str("prism_"+str(i).zfill(3)+".vtk")
    points, faces = prism(i)
    triangles = []
    triangles.append([0, 1, 5])
    triangles.append([5, 6, 1])
    cells["triangle"] = np.array(triangles)
    cells["quad"] = np.array(faces[1:i])
    cells["polygon"] = np.array(faces[i:i+2])
    meshio.write_points_cells(
        filename,
        points,
        cells,
        file_format="vtk-ascii"
        )
    print(cells)
    cells.clear()

    points, elements = readMesh(filename)
    print(elements)
```

meshio/vtk_io.py
nschloe: Ah, same misunderstanding as above. When I read `in` I think lists. Please also replace by `key[:7] == "polygon"`.

nschloe:
Can you allow edits from maintainers? https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

eolianoe:
> Can you allow edits from maintainers? https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Done

nschloe:
Okay so, the tests are failing if different polygons are present in a mesh. Please fix! :smiley:

eolianoe:
I add polygons to the legacy reader and tests are running fine, do not hesitate to tell me about bad coding style

nschloe:
Merged & released. Thanks for a nice contribution!

eolianoe:
You're welcome, thanks for the feedback and the great library !

meshio/helpers.py
MuellerSeb: Why not use the pythonic way: ``key.startswith("polygon")``?

eolianoe: Because my python is quite new, what's your opinion @nschloe?

nschloe:
Don't mind either way.

Issue #330, opened 2019-02-26
MSH 4.1


gdmcbain:
[gmsh#444 MSH 4.x format revisions](https://gitlab.onelab.info/gmsh/gmsh/issues/444) mean that the latest Gmsh produces MSH 4.1 by default which can't be read by `meshio.read`; downstream, this affects nschloe/pygmsh#240 and kinnala/scikit-fem#129.

The minimal example from nschloe/pygmsh#240 serves here: 
[segment.msh.gz](https://github.com/nschloe/meshio/files/2903759/segment.msh.gz).  Attempting to `meshio.read` this raises:
> ValueError: not enough values to unpack (expected 1, got 0)

from

 https://github.com/nschloe/meshio/blob/e18867bc753bcfa32f0265b20f454ae1b1d46edf/meshio/msh_io/msh4.py#L99

gdmcbain:
This might be the cause (gmsh#444, [geuzaine 2019-02-22](https://gitlab.onelab.info/gmsh/gmsh/issues/444#note_6461)):
> The bounding box for point entities replaced by the 3 coordinates of the point (instead of the six bounding box values)

The previous line in `msh4._read_entities` was 
https://github.com/nschloe/meshio/blob/e18867bc753bcfa32f0265b20f454ae1b1d46edf/meshio/msh_io/msh4.py#L98

gdmcbain:
Next error is on trying to parse the first line of the `$Nodes` block of `segment.msh`
> 3 6 1 6
with 
https://github.com/nschloe/meshio/blob/e18867bc753bcfa32f0265b20f454ae1b1d46edf/meshio/msh_io/msh4.py#L117

Of course this raises
> ValueError: too many values to unpack (expected 2)

gdmcbain:
The spec [&sect;9.1](http://gmsh.info/dev/doc/texinfo/gmsh.html#MSH-file-format-_0028version-4_0029) says
> numEntityBlocks(size_t) numNodes(size_t) minNodeTag(size_t) maxNodeTag(size_t)

as [flagged](https://gitlab.onelab.info/gmsh/gmsh/issues/444#note_6461) in gmsh#444
> The minimum and the maximum tag of nodes (resp. elements) added in the header of the $Nodes (resp. $Elements) section, to facilitate the detection of sparse or dense numberings when reading the file

gdmcbain:
[...]

PR #331, opened 2019-02-26
update msh4 reader to MSH4.1

gdmcbain:
This is part of dealing with MSH4.1 #330, but I don't think it's a complete fix.


gdmcbain:
The above is enough to parse the minimal example segment.msh of #330.  It may not correctly parse other valid MSH4.1 and certainly does not make any attempt to take any advantage of the improvements announced in gmsh#444.

gdmcbain:
Yes, lots of tests failing; I'll get back to this tomorrow.

nschloe:
Can we really just update `msh4.py`? What about the gmsh4.0 meshes out there, we want to support them too. I thought we need a separate `msh41.py` which is almost the same as `msh4.py`, with some updates.

gdmcbain:
As of 7297dcb, this correctly parses the ASCII and binary versions of the segment.msh example.

gdmcbain:
Test suite passing with Gmsh 4.2.0-git-d474ec1e2.

gdmcbain:
There are four failures in the pygmsh tests; however these are unrelated, they're all to do with compound entities nschloe/pygmsh#112 nschloe/pygmsh#214.

gdmcbain:
As reported over at [gmsh#444](https://gitlab.onelab.info/gmsh/gmsh/issues/444#note_6575), I tried this out on a smallish problem of 140930 3-node triangular elements.  Passing `mesh_file_type='msh2'` to `pygmsh.generate_mesh` meant 21 s in `meshio.read`; omitting it to get the default MSH4.1 reduced that to 0.024 s on this branch.  Reading by blocks rather than lines is a good idea.

meshio/msh_io/main.py
```diff
@@ -22,22 +22,41 @@ def read_buffer(f):
     int_size = sizeof(c_int)
     format_version, data_size, is_ascii = _read_header(f, int_size)
 
-    assert format_version in ["2", "4"], "Need mesh format 2 or 4 (got {})".format(
-        format_version
-    )
+    readers = {2: msh2, 4.0: msh4_0, 4.1: msh4_1}
```
nschloe: I guess `2`, `4.0`, `4.1` should be strings. The implicit float comparison in `readers[format_version]` makes me a bit nervous.

meshio/msh_io/main.py
```diff
@@ -22,22 +22,41 @@ def read_buffer(f):
...
 
     return reader.read_buffer(f, is_ascii, int_size, data_size)
 
 
 def _read_header(f, int_size):
+    """Read the mesh format block
+
+    specified as
+
+     version(ASCII double; currently 4.1)
```
nschloe: See above. -> str

meshio/msh_io/msh4_1.py
```diff
@@ -0,0 +1,452 @@
...
+        else:
+            data.append((physical_tags[dim_entity][tag_entity], tpe, d))
+
+    if not is_ascii:
+        line = f.readline().decode("utf-8")
+        assert line == "\n"
+    line = f.readline().decode("utf-8")
+    assert line.strip() == "$EndElements"
+
+    # The msh4 elements array refers to the nodes by their tag, not the index. All other
```
nschloe: Here, we could check if the tag are already in order (as pointed out by Christophe), and if yes, set `itags` directly.

meshio/msh_io/msh4_1.py
```diff
@@ -0,0 +1,452 @@
...
+        fh.write("{} {}\n".format(1, len(points)).encode("utf-8"))
+
+        # tagEntity(int) dimEntity(int) typeNode(int) numNodes(unsigned long)
+        fh.write(
+            "{} {} {} {}\n".format(1, dim_entity, type_node, len(points)).encode(
+                "utf-8"
+            )
+        )
+
+        for k, x in enumerate(points):
```
nschloe: Is this still correct, i.e., are the tags still prepended to each line? We could think about removing this loop, too.

meshio/msh_io/msh4_1.py
```diff
@@ -0,0 +1,452 @@
...
+                    _geometric_dimension[cell_type],
+                    _meshio_to_gmsh_type[cell_type],
+                    node_idcs.shape[0],
+                ).encode("utf-8")
+            )
+            # increment indices by one to conform with gmsh standard
+            idcs = node_idcs + 1
+
+            fmt = " ".join(["{}"] * (num_nodes_per_cell[cell_type] + 1)) + "\n"
+            for idx in idcs:
```
nschloe: Another evil loop. Let's at least add a `TODO` comment here.

meshio/msh_io/msh4_1.py
```diff
@@ -0,0 +1,452 @@
...
+                fh.write(fmt.format(consecutive_index, *idx).encode("utf-8"))
+                consecutive_index += 1
+
+    fh.write("$EndElements\n".encode("utf-8"))
+    return
+
+
+_geometric_dimension = {
```
nschloe: Perhaps this can be shared between 4_0 and 4_1.

nschloe:
Tests!

meshio/msh_io/msh4_1.py
gdmcbain: This write method is plain wrong, it's just the meagre old MSH4.0 one, itself a quick update of the MSH2 method. I was going to defer writing MSH4.1.

gdmcbain:
Ah, O. K.  So tests in the current framework requires `write` #332, which I'd otherwise been hoping to defer.  I'll get onto it.

As back in #304, I feel a bit uncomfortable that the test of my `read` is to be my `write`, as if I were correcting my own Russian homework rather than getting a native speaker to do it.  That is, I think it might be better to test against authoritative reference implementations, Gmsh in the case of MSH.

meshio/msh_io/main.py
gdmcbain: Yes.  506685abbf715598b714160030a19231e0eaca52

also sharing this `Dict[str, Module]` with `msh_io.write`.

I note that according to the spec, in [MSH file format version 2](http://gmsh.info/dev/doc/texinfo/gmsh.html#MSH-file-format-version-2)
> `version-number` is a real number equal to 2.2

from which I infer that 2.0 and 2.1 are unsupported, but I'll leave that aside for the moment, treating all 2.x the same.

meshio/msh_io/main.py
gdmcbain: Yes.  That's just quoting the spec for [MSH file format (version 4)](http://gmsh.info/dev/doc/texinfo/gmsh.html#MSH-file-format-_0028version-4_0029), but I added a line noting that here it's kept as str. e32ccf795d630ef340b106dd91f0be31d8d045bb

meshio/msh_io/msh4_1.py
gdmcbain: Yes.  1bb3111555a31f0ef1b47d3dc46b134a55ccda24

meshio/msh_io/msh4_1.py
gdmcbain: No, it's wrong, syntax for MSH4.1 is (at least for `parameter == 0`)
> nodeTag(size_t)
>    ...
>   x(double) y(double) z(double)

I had been hoping to defer `msh4_1.write` #332 but will get onto this now.

gdmcbain: As reported, #332 is stalled.

gdmcbain:
At close of business today, the reader seems to be working O. K. (no known failures here or from pygmsh or scikit-fem) but writing the writer #332 is stalled.  Therefore no local tests yet.

meshio/msh_io/main.py
nschloe: > version(ASCII double; currently 4.1)

Smh. We'll stick with strings.

nschloe:
> As back in #304, I feel a bit uncomfortable that the test of my `read` is to be my `write`, as if I were correcting my own Russian homework rather than getting a native speaker to do it. That is, I think it might be better to test against authoritative reference implementations, Gmsh in the case of MSH

I agree. The best option would be to install some upstream package and verify the correctness of `write` against those. This would complicate testing _so_ much though. Imagine having to install Gmsh, VTK, etc. with correct version. I'm sure quite soon we wouldn't be testing anymore at all. Right now, I find we're striking a good balance making sure that the respective `read` and `write` functions are matching. When developing, at least one time one has to open a file with ParaView or whatever to make sure it's working as expected.

gdmcbain:
…thinking…

I agree with your assessment but am trying to come up with some way around it. Is this 'configuration management'? The continuous integration testing in pygmsh is pretty good; it copes with a bit of that. I don't know how to move it offline though. This is of course far outside the scope of the present issue but configuration management would also, I'm thinking, be very useful for reproducible simulation, both in research and industrially.

Anyway here I agree the immediate answer is to understand MSH 4.1 and write a writer #332. We've probably got a little while till the MSH 4.1 bugs start coming back up from downstream.

Issue #335, opened 2019-03-04
a test suite for readers


gdmcbain:
Following discussion of testing in #331 where there's a reasonable implementation of `read` for MSH 4.1 #330 but as yet no `write` #332 , the problem being that currently testing involves basically checking that `read` is the left inverse of `write` which is impossible without the latter.  The need for that approach is that it avoids non-Python dependencies (Gmsh, VTK, let alone proprietary products) and their various versions.

What about including in the test suite some authoritatively generated meshes?  Then one could test that the `Mesh` returned by `read` satisfied some basic properties (number of elements with a given tag, bounding box of points, …).

Besides enabling testing of readers lacking writers, this would seem to me to be a better test of a reader since it's reading a real native mesh of the given format.

nschloe:
Why not, I guess. The (possibly large) reference files shouldn't be in master though but in some other location, e.g., an orphaned branch. and then downloaded at test time via something like https://github.com/nschloe/meshplex/blob/master/test/helpers.py.

nschloe:
The more I think about, the more I like the idea. Thanks @gdmcbain for another great suggestion! I'll probably get going with this _today_.

nschloe:
Okay so, https://github.com/nschloe/meshio/pull/338 mostly fixes this. It's only one VTK mesh now, but the functionality is there and now it's just a matter of adding more meshes to the gh-pages branch and filling in the respective tests.

gdmcbain:
Some discussion under #365 led to [a statement of the rationale for keeping the exemplars in a separate branch](https://github.com/nschloe/meshio/pull/365#issuecomment-476164858), linked to here for future reference.

PR #342, opened 2019-03-13
MSH2.2 and MSH4.1 test meshes

gdmcbain:
Here's a couple of contributions to the test suite #335, generated by Gmsh 4.2.2-git-0766f664d from a common [ex17.geo.tar.gz](https://github.com/nschloe/meshio/files/2959646/ex17.geo.tar.gz) generated by pygmsh.  

It's a real example, from [scikit-fem/docs/examples/ex17.py](https://github.com/kinnala/scikit-fem/blob/6c1be376130f0152ead741b31dad7b212871f4d1/docs/examples/ex17.py), with 3-node triangular domain elements in two Physical Surfaces (subdomains) and 2-node line boundary elements in a Physical Line.

The two meshes were generated with
```shell
gmsh -2 -format msh2 -o ex17-2.2.msh ex17.geo
gmsh -2 -o ex17-4.1.msh ex17.geo
```

The second, being MSH4.1 #330, cannot be read by current master, pending merging of #331; it is read successfully (producing an identical `meshio.Mesh`) on branch gdmcbain:msh4-1-330.


gdmcbain:
I think that the corresponding tests have to go in a different pull request?  Because they're destined for a different branch.

nschloe:
Perhaps we can also add a README.md to the folder saying how to generate the files or where they're coming from.

gdmcbain:
Done.

To make that a bit cleaner, I've simplified the GEO file by hand, removing the pygmsh artefacts.  I also increased the characteristic length so that the files are an order of magnitude smaller; I figure a hundred elements is good as a thousand for the present purpose.

I'm not exactly sure how reproducible MSH files are from GEO (with respect to a measure like `mesh.points.sum()`).   I know that a lot of mesh generation techniques include random steps and even though Gmsh has done something to improve reproducibility on a single host by seeding the random number generator, I don't know how that goes if the operating system has a different random number generator?   I could imagine all kinds of things like that that could go wrong, but the idea of the suite is to have static mesh files; anyway, the README.md, insulated.geo, and Makefile show how to make meshes like those provided.

nschloe:
> I figure a hundred elements is good as a thousand for the present purpose.

:+1: 


> I'm not exactly sure how reproducible MSH files are from GEO (with respect to a measure like mesh.points.sum()).

Never mind that for now. The geo files are more than enough.

PR #343, opened 2019-03-14
only named arguments after *args

gdmcbain:
Maybe just me, but I found having the named positional argument `"2"` after `*args` confusing.

https://github.com/nschloe/meshio/blob/7995c68e10460aacf45af57c88c6dbf4af26b2e9/test/test_gmsh.py#L49 

Ignore if you disagree.

### DIFF
--- a/test/test_gmsh.py
+++ b/test/test_gmsh.py
@@ -78,7 +78,7 @@ def test_gmsh2(mesh, write_binary):
 )
 @pytest.mark.parametrize("write_binary", [False, True])
 def test_gmsh4(mesh, write_binary):
-    writer = partial(meshio.msh_io.write, fmt_version="2", write_binary=write_binary)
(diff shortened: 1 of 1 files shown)


test/test_gmsh.py
```diff
@@ -78,8 +78,7 @@ def writer(*args, **kwargs):
 )
 @pytest.mark.parametrize("write_binary", [False, True])
 def test_gmsh4(mesh, write_binary):
-    def writer(*args, **kwargs):
-        return meshio.msh_io.write(*args, "4", write_binary=write_binary, **kwargs)
+    writer = partial(meshio.msh_io.write, fmt_version="2", write_binary=write_binary)
```
nschloe: Should be `"4"` here, right?

gdmcbain: Oops!

PR #345, opened 2019-03-15
Read retracts write

gdmcbain:
So far the tests #338 #344 against the meshes from the suite #335 only test the readers, but they're also probably pretty good for testing the writers in the sense that read &sdot; write &sdot; read should produce the same as read.  This implements that for VTK and MSH2.2.

### DIFF
--- a/test/test_gmsh.py
+++ b/test/test_gmsh.py
@@ -102,7 +102,8 @@ def test_generic_io():
         )
     ],
 )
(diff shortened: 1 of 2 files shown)


gdmcbain:
https://github.com/nschloe/meshio/blob/20fa0104b751c8bf3868b8c081a09c7320c606ac/test/helpers.py#L275-L276

I agree that a mesh is equivalent under a permutation of the points, but under what circumstances will the points be deranged?  Are there any writers that do this?

Issue #347, opened 2019-03-15
MED 4.0 to .msh or .xml with entities


RemiTheWarrior:
Hi all,

I wish to convert a mesh from SALOME (MED 4.0) to .msh or .xml alongside with cell attributes.

Here's an exemple of a mesh from SALOME with 1 volume marked.
[MED.zip](https://github.com/nschloe/meshio/files/2970767/MED.zip)

Meshio seems to read it without issues. But, when converting to .msh and after opening it in GMSH, here's the mesh file I get.
[gmsh.zip](https://github.com/nschloe/meshio/files/2970766/gmsh.zip)

No sign of the entity "tungsten".

Thanks guys for your help!

Rem

PS: Screenshot of the med file re-imported in SALOME, the group is there as expected.
![image](https://user-images.githubusercontent.com/40028739/54428450-d3876400-471d-11e9-969d-8ba30db630ed.png)

RemiTheWarrior:
Update: 
I tried to convert the MED file to another MED file with meshio, it seems it removed the topological information ...
I'm using SALOME 9.2.2 .

![image](https://user-images.githubusercontent.com/40028739/54438490-01c56d80-4737-11e9-9009-b44fff413f9c.png)

gdmcbain:
Earlier this month we started putting together a test suite of authoritatively generated mesh files #335, e.g. MSH by Gmsh #342. Could this one serve as such an exemplar for MED? I don't know much about MED myself; is Salome its authoritative generator?

RemiTheWarrior:
I think it is though I'm far from being an expert. Anyway, this one can definitely serve a an exemple.

RemiTheWarrior:
I noticed in med_io.py, the function `read(filename)` doesn't seem to take into account groups which should be under FAS. Moreover in `write(filename, mesh, add_global_ids=True)`, the Subgroups seem not to be filled (just the blank fields).  The groups seem not be read nor written in any case.
This may be the reason why the groups vanish when converting from MED to MED ?

@nschloe Please correct me if I'm wrong I might have misunderstood something there

tianyikillua:
@RemiTheWarrior That's correct! I'm not sure if there are already established conventions as how to store node/element groups in a `meshio`-mesh class...Using these information to write to a MED file should be easy...

gdmcbain:
[...]

PR #350, opened 2019-03-19
WIP: read tagged subsets of elements from MED

gdmcbain:
This is a start on #347 'MED 4.0 to .msh or .xml with entities', i.e. being able to read MED like Gmsh does, preserving tagged subsets of elements #290.

It seems to be a bit at odds with the current implementation though, and MED is a big format which I don't claim to fully understand, so I don't want to take it too much further before resolving some of the topics in #347.

### DIFF
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -67,6 +67,8 @@ def read(filename):
         if families.any():
             cells[cell_type] = med_cell_type_group["NOD"][()].reshape(
                 num_nodes_per_cell[cell_type], -1).T - 1
+            if cell_type not in cell_data:
+                cell_data[cell_type] = {}
             cell_data[cell_type]["gmsh:physical"] = families
 
     return Mesh(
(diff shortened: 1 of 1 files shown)


meshio/med_io.py
```diff
@@ -51,21 +51,24 @@ def read(filename):
...
         point_data, cell_data, field_data = _read_data(cha)
     except KeyError:
         point_data, cell_data, field_data = {}, {}, {}
 
+    # Cells
+    cells = {}
+    mai = mesh["MAI"]
+    for med_cell_type, med_cell_type_group in mai.items():
+        cell_type = med_to_meshio_type[med_cell_type]
+        families = med_cell_type_group["FAM"][()]
```
tianyikillua: `"FAM"` may not exist so this may cause an error.

gdmcbain: Oh.  I was going on the example posted to #347 by @RemiTheWarrior.  I'll retract this proposal; I don't understand MED well enough.

I do recommend 776ab4a though, I think.

PR #351, opened 2019-03-20
iterate directly through types of cells present in MED

gdmcbain:
A minor improvement from the otherwise abandoned proposal to read tagged subsets of elements from MED #350 was to iterate directly through the given MED types of cells rather than through all known types of cells and see whether they're present.  A separate pull request was requested.

This also makes use of the known number of nodes for the type of cell rather than parsing it from MED's name for the type.

### DIFF
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -6,8 +6,8 @@ I/O for MED/Salome, cf.
 """
 import numpy
 
-from .common import num_nodes_per_cell
 from .mesh import Mesh
+
 
 # https://bitbucket.org/code_aster/codeaster-src/src/default/catalo/cataelem/Commons/mesh_types.py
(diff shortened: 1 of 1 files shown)


tianyikillua:
> A minor improvement from the otherwise abandoned proposal to read tagged subsets of elements from MED #350 was to iterate directly through the given MED types of cells rather than through all known types of cells and see whether they're present. A separate pull request was requested.
> 
> This also makes use of the known number of nodes for the type of cell rather than parsing it from MED's name for the type.

Your previous PR was almost ok in fact...😃With your permission I will start from your previous PR...

gdmcbain:
Yes, please do. It would be great to see more complete support for MED in meshio and progress on translation of tagged subsets of entities between formats.

Issue #265, opened 2018-06-26
gmsh physical groups -> dolfin XML errors


dajuno:
Hi,

there is an error in the conversion of gmsh meshes with tagged boundaries via physical lines/surfaces to dolfin XML. Consider the following unit square mesh, with boundary IDs 1 and 2: 

```
h = 0.1;
Point(1) = {0, 0, 0, h};
Point(2) = {1, 0, 0, h};
Point(3) = {1, 1, 0, h};
Point(4) = {0, 1, 0, h};

Line(1) = {1, 2};
Line(2) = {2, 3};
Line(3) = {3, 4};
Line(4) = {4, 1};

Line Loop(5) = {1, 2, 3, 4};
Plane Surface(1) = {5};

Physical Line(1) = {3};   // lid
Physical Line(2) = {1, 2, 4};   // wall
Physical Surface(0) = {1};

Mesh 2;
Save "square.msh";
```

- [ ] After conversion with `meshio-convert square.msh square.xml`, the `gmsh:geometrical` and the `gmsh:physical` files will both contain a list of all vertices, dimension 2, with labels 0 in one file, and 1 in the other.
Expected result (when converted with `dolfin-convert`) is a `facet_region` file that contains the labels (0, 1, 2) for all 27 facets of the mesh, and dim=1.
- [ ] Also, if `<mesh_function type="uint">` was used rather than `type="int"`, these files could be read by fenics' `MeshFunction("size_t", ...)` (instead of "int"). "uint" is more consistent, since "size_t" is used everywhere in the fenics documentation and `dolfin-convert` generates "uint".
- [ ] line breaks are missing in the  `gmsh:geometrical|phyiscal.xml` files.

Best wishes,
David

nschloe:
Okay, first thing I'm going to recommend to anyone using dolfin XML: Don't use it. Upstream as moved to XDMF, and you should probably do the same.

dajuno:
I agree. It's more worthwhile making the XDMF conversion hassle-free. Apart from issue #260, how do you read physical groups in FEniCS from XDMF created by meshio? I know you're a FEniCS user yourself ;)

nschloe:
> Apart from issue #260, how do you read physical groups in FEniCS from XDMF created by meshio?

I've never done this.

Issue #260, opened 2018-06-23
dolfin XDMF: unable to recognise cell type


dajuno:
Hi,

the XDMF format created with meshio (by conversion of gmsh meshes) is not compatible with dolfin (although the deprecation warning for the dolfin-xml format recommends using XDMF).
The conversion seems to work, but the resulting XDMF file cannot be read in dolfin.

Dolfin error message:
```
*** Error:   Unable to recognise cell type.                                         
*** Reason:  Unknown value "mixed".
```
Best wishes
David

nschloe:
Sounds like a FEniCS error to me.

dajuno:
Have you seen this on the FEniCS slack?
https://fenicsproject.slack.com/archives/C2K9V55SA/p1529598166000183
Apparently FEniCS expects different cell types to be stored as separate items, not together as mixed.

nschloe:
:+1: Let's discuss over there.

a1650078:
As a workaround, it is possible to remove the non-tetrahedron cells with the `--prune` option.

nschloe:
I'm thinking about fixing this. What I need is an XDMF file containing different cell types that actually works with FEniCS. Can anyone help out?

a1650078:
I only use FEniCS occasionally and remember pulling out a lot of hair once upon a time trying to unravel fenics mesh imports - I actually switched to meshio when my old workflow stopped working. It turned out that (I think) with .xdmf imports (probably others, I don't know) the facet information is just not in the mesh file at all. Instead, facet information must be in a separate xdmf file defining a MeshValueCollection. I discovered this [reading about another issue](https://fenicsproject.org/qa/3237/default-meshfunction-created-sparse-meshvaluecollection/) but there is some documentation somewhere too. My workaround was to write a few lines of python code to recover the facets and create a FacetFunction from the cells manually. I think meshio generating this file would much easier to work with though

gdmcbain:
> 👍 Let's discuss over there.

Unfortunately [there](https://fenicsproject.slack.com/archives/C2K9V55SA/p1529598166000183) I'm told
> Pour consulter l'historique complet de complet de cette cha&icirc;ne et y effectuer des recherches, vous devez passer &agrave; l'un de nos forfaits payants.

Did anything worthwhile transpire?  Meanwhile I'll see whether I can cook anything up here.

nschloe:
[...]

PR #352, opened 2019-03-20
Read and write point/cell sets of a MED file

tianyikillua:
Based on #350, an attempt to read and write point/cell sets in a MED file.

Some characteristics of sets in a MED file are described in https://github.com/nschloe/meshio/issues/347#issuecomment-474247403, https://github.com/nschloe/meshio/issues/347#issuecomment-474251078 and https://github.com/nschloe/meshio/issues/290#issuecomment-474999314.

The current implementation will additionally provide if they are defined in the mesh
- a new `point_data` called `point_tags`, that assigns a unique node set id (positive integers, defined by MED) to each node
- a new `cell_data` called `cell_tags`, that assigns a unique element set id (negative integers, defined by MED) to each element

Since currently with gmsh you have `gmsh:physical`, so reading a MSH file with tags and then writing it to MED loses tags information, and vice versa. Maybe we can rename it also to `cell_tags`? We need to inform users by updating the change log.

Reading the example [Cylinder.zip](https://github.com/nschloe/meshio/files/2982723/Cylinder.zip), and then exporting to VTU gives under ParaView the expected behavior (here `cell_tags = -6` corresponds to the `Top circle` edge element sets).

The information about these `tags` is stored in `mesh.point_tags` and `mesh.cell_tags`: the unique set id → list of set name(s) defined in the mesh. Additionally, intersections of sets internally created by MED are decomposed in the list of set name(s).

The difference between gmsh `mesh.field_data` and `mesh.point/cell_tags` being that
1. Manifold dimension of sets is unavailable
2. The key of the dictionary is the unique set id defined by MED, and not the set name. In MED we can give the same name to different sets corresponding to different manifold dimensions (so `Top` can be used to group all 2d elements on the top, and all 3d elements sharing nodes on the top), internally MED will just assign another unique set id.
3. Allowing overlapping sets, by using a list of subset names for each set id.

In this example we have
``` python
mesh.point_tags =
{2: ['Side'],
 3: ['Side', 'Top'],
 4: ['Top']}

mesh.cell_tags =
{-6: ['Top circle'],
 -7: ['Top', 'Top and down'],
 -8: ['Top and down'],
 -9: ['A', 'B'],
 -10: ['B'],
 -11: ['A', 'B', 'C'],
 -12: ['B', 'C'],
 -13: ['C']}
```

So for node sets,
- `Side` set defined by the user is in fact obtained by combining nodes from 2 and 3
- `Top` can be obtained by combining 3 and 4

For element sets,
- `A` can be obtained by grouping elements tagged with -9, -11
- `B`, with -9, -10, -11, -12
- `C`, with -11, -13

Note that with MED, point sets are internally tagged with a positive number, while cell sets with a negative number.

Another important information concerning writing a mesh to MED. If the mesh contains `point_tags` and `cell_tags` in the `point_data` and `cell_data`, then they are interpreted as node/elements tags (sets, groups, ...) and not normal data. They will be written under `FAS` of MED and not under `CHA` where we put normal point/cell fields.

The reason is that if initially we have a mesh with no (other) fields but only tags `point_tags` and `cell_tags`, we don't want a MED file with both tags under `FAS` and also represented by two additional normal fields.

PS: the tags under `FAS` can be well interpreted by salome, both under the mesh module where user-defined (possibly overlapping) sets can be seen, but also under ParaVis (ParaView + MED support), where these two tags are named `FamilyidNode` and `FamilyIdCell`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,0 @@
-# Change log
-
-## Unreleased
-
-- MED: support reading and writing point and cell sets #347 #352
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,0 +1,5 @@
+# Change log
+
+## Unreleased
+
+- MED: support reading and writing point and cell sets #347 #352
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -221,24 +221,58 @@ def write(filename, mesh, add_global_ids=True):
     coo.attrs.create("CGT", 1)
     coo.attrs.create("NBR", len(mesh.points))
 
+    # Point tags
+    if "point_tags" in mesh.point_data:  # works only for med -> med
+        fam = noe_group.create_dataset("FAM", data=mesh.point_data["point_tags"])
+        fam.attrs.create("CGT", 1)
+        fam.attrs.create("NBR", len(mesh.points))
+
     # Cells (mailles in french)
     mai_group = ts.create_group("MAI")
     mai_group.attrs.create("CGT", 1)
-    for key, med_type in meshio_to_med_type.items():
-        if key in mesh.cells:
-            mai = mai_group.create_group(med_type)
-            mai.attrs.create("CGT", 1)
-            mai.attrs.create("CGS", 1)
-            mai.attrs.create("PFL", pfl.encode("ascii"))
-            nod = mai.create_dataset("NOD", data=mesh.cells[key].T.flatten() + 1)
-            nod.attrs.create("CGT", 1)
-            nod.attrs.create("NBR", len(mesh.cells[key]))
+    for cell_type, cells in mesh.cells.items():
+        med_type = meshio_to_med_type[cell_type]
+        mai = mai_group.create_group(med_type)
+        mai.attrs.create("CGT", 1)
+        mai.attrs.create("CGS", 1)
+        mai.attrs.create("PFL", pfl.encode("ascii"))
+        nod = mai.create_dataset("NOD", data=cells.T.flatten() + 1)
+        nod.attrs.create("CGT", 1)
+        nod.attrs.create("NBR", len(cells))
 
-    # Sets (familles in french)
+        # Cell tags
+        if "cell_tags" in mesh.cell_data[cell_type]:  # works only for med -> med
+            fam = mai.create_dataset("FAM", data=mesh.cell_data[cell_type]["cell_tags"])
+            fam.attrs.create("CGT", 1)
+            fam.attrs.create("NBR", len(cells))
+
+    # Information about point and cell sets (familles in french)
     fas = f.create_group("FAS")
     fm = fas.create_group(mesh_name)
     fz = fm.create_group("FAMILLE_ZERO")  # must be defined in any case
     fz.attrs.create("NUM", 0)
+
+    # For point tags
+    try:
+        if len(mesh.point_tags) > 0:
+            noeud = fm.create_group("NOEUD")
+            _write_families(noeud, mesh.point_tags)
+    except AttributeError:
+        pass
+
+    # For cell tags
+    try:
+        if len(mesh.cell_tags) > 0:
+            eleme = fm.create_group("ELEME")
+            _write_families(eleme, mesh.cell_tags)
+    except AttributeError:
+        pass
+
+    # Ignore point_tags and cell_tags that are already
+    # written under FAS
+    mesh.point_data.pop("point_tags", None)
+    for cell_type in mesh.cell_data:
+        mesh.cell_data[cell_type].pop("cell_tags", None)
 
     # Write nodal/cell data
     if mesh.point_data or mesh.cell_data:
@@ -334,3 +368,26 @@ def _comp_nom(nco):
     the data, we just use V1, V2, ...
     """
     return "".join(["V%-15d" % (i + 1) for i in range(nco)])
+
+
+def _family_name(num, nom):
+    """
+    Return the FAM object name corresponding to
+    the unique set id and a list of subset names
+    """
+    return "FAM" + "_" + str(num) + "_" + "_".join(nom)
+
+
+def _write_families(fm_group, tags):
+    """
+    Write point/cell tag information under FAS/[mesh_name]
+    """
+    for num, nom in tags.items():
+        fam = fm_group.create_group(_family_name(num, nom))
+        fam.attrs.create("NUM", num)
+        gro = fam.create_group("GRO")
+        gro.attrs.create("NBR", len(nom))  # number of subsets
+        dataset = gro.create_dataset("NOM", (len(nom),), dtype="80int8")
+        for i in range(len(nom)):
+            nom_80 = nom[i] + "\x00" * (80 - len(nom[i]))  # make nom 80 characters
+            dataset[i] = [ord(x) for x in nom_80]
--- a/meshio/med_io.py
+++ b/meshio/med_io.py
@@ -33,27 +33,27 @@ def read(filename):
     f = h5py.File(filename, "r")
 
     # Mesh ensemble
-    ens_maa = f["ENS_MAA"]
-    meshes = ens_maa.keys()
+    mesh_ensemble = f["ENS_MAA"]
+    meshes = mesh_ensemble.keys()
     assert len(meshes) == 1, "Must only contain exactly 1 mesh"
     mesh_name = list(meshes)[0]
-    mesh = ens_maa[mesh_name]
+    mesh = mesh_ensemble[mesh_name]
 
     # Possible time-stepping
     if "NOE" not in mesh:
         # One needs NOE (node) and MAI (french maillage, meshing) data. If they
         # are not available in the mesh, check for time-steppings.
-        ts = mesh.keys()
-        assert len(ts) == 1, "Must only contain exactly 1 time-step"
-        mesh = mesh[list(ts)[0]]
+        time_step = mesh.keys()
+        assert len(time_step) == 1, "Must only contain exactly 1 time-step"
+        mesh = mesh[list(time_step)[0]]
 
     # Read nodal and cell data if they exist
     try:
-        cha = f["CHA"]  # champs (fields) in french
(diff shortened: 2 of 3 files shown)


tianyikillua:
This PR provides a new way of defining boundary conditions markers under FEniCS.

The previous MED example (3d cylinder) doesn't work due to the presence of mixed elements (incredible FEniCS !!!!!). So I created a shell cylinder with only triangular elements.

[Cylinder_only_faces.zip](https://github.com/nschloe/meshio/files/2991907/Cylinder_only_faces.zip)

1. First reading and converting the MED file containing node/element sets (using Salome for example) to a XDMF file.
```python
import meshio

mesh = meshio.read("Cylinder_only_faces.med")

# In order to use MeshFunction of FEniCS
# The tag must be a positive number (size_t)
mesh.cell_data["triangle"]["cell_tags"] *= -1

meshio.write("Cylinder_only_faces.xdmf", mesh)
```

2. Then reading this file under FEniCS
``` python
from fenics import *

mesh = Mesh()
mesh_file = XDMFFile("Cylinder_only_faces.xdmf")
mesh_file.read(mesh)

boundaries = MeshFunction("size_t", mesh, mesh.topology().dim())
mesh_file.read(boundaries, "cell_tags")

V = FunctionSpace(mesh, "CG", 1)
u = Function(V)

# mesh.tags = {6: ['Top', 'Top and down'],  -> Top by intersection
#              7: ['Top and down']} -> Down in fact

with XDMFFile("boundaries.xdmf") as f:
    f.write(boundaries)
```

[boundaries.zip](https://github.com/nschloe/meshio/files/2991906/boundaries.zip)

RemiTheWarrior:
That sounds ace! If I understood correctly, we can't tag at the same time tets and triangles ?

nschloe:
Is this ready to be reviewed or still WIP? Just ping me when you're ready.

tianyikillua:
> That sounds ace! If I understood correctly, we can't tag at the same time tets and triangles ?

Due to #260, #265, there is no trivial way of doing this.

Workaround https://gist.github.com/michalhabera/bbe8a17f788192e53fd758a67cbf3bed.

RemiTheWarrior:
> Workaround https://gist.github.com/michalhabera/bbe8a17f788192e53fd758a67cbf3bed.

Ah yes I remember using `MeshValueCollection` in FEniCS in [a previous project](https://github.com/RemiTheWarrior/3-MOT/blob/master/Main/3-MOT.py). The XDMF mesh was generated from the meshing software Trellis and then passed to FEniCS. It worked quite well for tagging tets and triangles.

tianyikillua:
@nschloe Ok I'm happy now. Do you want me to update a reference MED file with node/elements sets for unit test? Or in another PR?

nschloe:
> Or in another PR?

It's got to be another PR because it targets another branch (gh-pages). Best do that first, then add a test like in VTK.

tianyikillua:
So it's passing also the new unit test. I've updated the PR description.

meshio/med_io.py
```diff
@@ -46,11 +47,29 @@ def read(filename):
         assert len(ts) == 1, "Must only contain exactly 1 time-step"
         mesh = mesh[list(ts)[0]]
 
+    # Read nodal and cell data if they exist
+    try:
+        cha = f["CHA"]  # champs (fields) in french
+        point_data, cell_data, field_data = _read_data(cha)
```
nschloe: Better: `try`, `except`,
```
else:
    point_data, cell_data, field_data = _read_data(cha)
```
(This way, we don't accidentally intercept `KeyError`s from `_read_data`.)

meshio/med_io.py
```diff
@@ -46,11 +47,29 @@ def read(filename):
         assert len(ts) == 1, "Must only contain exactly 1 time-step"
         mesh = mesh[list(ts)[0]]
 
+    # Read nodal and cell data if they exist
+    try:
+        cha = f["CHA"]  # champs (fields) in french
```
nschloe: I can't get over the French naming scheme. :laughing:

meshio/med_io.py
```diff
@@ -180,52 +221,86 @@ def write(filename, mesh, add_global_ids=True):
...
+    # Information about point and cell sets (familles in french)
     fas = f.create_group("FAS")
     fm = fas.create_group(mesh_name)
     fz = fm.create_group("FAMILLE_ZERO")  # must be defined in any case
     fz.attrs.create("NUM", 0)
 
+    # For point tags
+    try:
+        if len(mesh.point_tags) > 0:
+            noeud = fm.create_group("NOEUD")
```
nschloe: Let's not use French variable names ourselves. `nodes` would be fine. (Or `knoten`, since I'm from Germany. Your choice. :smile:)

ChangeLog.md
```diff
@@ -0,0 +1,5 @@
+# Change log
```
nschloe: Not sure if this PR is a good starting point for a changelog. I can see myself being too lazy to maintain one. @gdmcbain What do you think?

meshio/med_io.py
RemiTheWarrior: That's really silly!

tianyikillua:
> That sounds ace! If I understood correctly, we can't tag at the same time tets and triangles ?

Rethinking your question, oh we can! In `mesh.cell_tags` we can tag both volumetric elements and surface elements. Then using https://github.com/nschloe/meshio/pull/352#issuecomment-475204271 it should can be read by FEniCS.

Previously I was referring to a mesh of which the volumetric elements are "mixed", composing of for example tetra and hexa elements.

meshio/med_io.py
tianyikillua: Ok, I will let Molière speak Shakespeare (or Goethe).

ChangeLog.md
gdmcbain: I don't use change logs myself. Do they add anything over `git log`, `git blame`, &c.?

tianyikillua: The main objective is to inform users of (major) API changes...

nschloe: I got to say that I always find it very convenient to have changelog in other pieces of software since I'm curious about new things and too lazy to dig through the git history of other projects.

For us to maintain properly, I think it's too much work atm. Let's not start this as part of this PR.

tianyikillua: Yes maybe for this (currently relatively not so big) project git history is enough..

nschloe: Let's remove the changelog then.

nschloe:
Thanks @tianyikillua for the great work!

PhoenixNewbie:
This is not working for me. The code fails at `mesh.cell_data["triangle"]["cell_tags"] *= -1` with error `KeyError: 'triangle'`. In case you are wondering I did not remake the MED, I just downloaded the file from the thread. 

> [Cylinder_FEniCS.zip](https://github.com/nschloe/meshio/files/2996797/Cylinder_FEniCS.zip)
> 
> Based on https://gist.github.com/michalhabera/bbe8a17f788192e53fd758a67cbf3bed but also https://github.com/RemiTheWarrior/3-MOT/blob/master/Main/3-MOT.py.
> 
> 1. First read and convert the MED file containing node/element sets to a XDMF file.
> 
> ```python
> import meshio
> 
> mesh = meshio.read("Cylinder_FEniCS.med")
> meshio.write("Cylinder_FEniCS.xdmf", mesh)  # won't work for FEniCS, mixed
> 
> # In order to use MeshFunction of FEniCS
> # The tag must be a positive number (size_t)
> mesh.cell_data["triangle"]["cell_tags"] *= -1
> mesh.cell_data["tetra"]["cell_tags"] *= -1
> # mesh.cell_tags = {-6: ['Down'], -7: ['Top'], -8: ['Lying on Top']}
> 
> # Export mesh that contains only tetra cells
> # along with tags
> meshio.write_points_cells(
>     "Cylinder_FEniCS_tetra.xdmf",
>     mesh.points,
>     {"tetra": mesh.cells["tetra"]},
>     cell_data={"tetra": {"cell_tags": mesh.cell_data["tetra"]["cell_tags"]}},
> )
> 
> # Export mesh that contains only triangular faces
> # along with tags
> meshio.write_points_cells(
>     "Cylinder_FEniCS_triangle.xdmf",
>     mesh.points,
>     {"triangle": mesh.cells["triangle"]},
>     cell_data={"triangle": {"cell_tags": mesh.cell_data["triangle"]["cell_tags"]}},
> )
> ```
> 
> 2. Then read this file under FEniCS
> 
> ```python
> from fenics import *
> 
> # Read volumetric mesh
> mesh = Mesh()
> # XDMFFile("Cylinder_FEniCS.xdmf").read(mesh)  # ERROR: Unknown value "mixed"
> XDMFFile("Cylinder_FEniCS_tetra.xdmf").read(mesh)
> 
> # Read tags for volume elements
> cell_markers = MeshFunction("size_t", mesh, mesh.topology().dim())
> XDMFFile("Cylinder_FEniCS_tetra.xdmf").read(cell_markers, "cell_tags")
> with XDMFFile("cell_markers.xdmf") as f:
>     f.write(cell_markers)
> 
> # Read tags for surface elements (can also be used for applying DirichletBC)
> # boundaries = MeshFunction("size_t", mesh, mesh.topology().dim() - 1)  # wont' work
> boundaries = MeshValueCollection("size_t", mesh, mesh.topology().dim() - 1)
> XDMFFile("Cylinder_FEniCS_triangle.xdmf").read(boundaries, "cell_tags")
> boundaries = MeshFunction("size_t", mesh, boundaries)
> 
> V = FunctionSpace(mesh, "CG", 1)
> u = Function(V, name="Test")
> 
> # mesh.cell_tags = {-6: ['Down'], -7: ['Top'], -8: ['Lying on Top']}
> bc_down = DirichletBC(V, Constant(-1), boundaries, 6)
> bc_top = DirichletBC(V, Constant(1), boundaries, 7)
> for bc in [bc_down, bc_top]:
>     bc.apply(u.vector())
> 
> with XDMFFile("u.xdmf") as f:
>     f.write(u)
> ```

PR #364, opened 2019-03-22
Gmsh: Various fixes for mesh format v4.1 

dalcinl:


dalcinl:
Oh, you where too fast! I'll make a new one to fix some weird issue with numpy dtypes

nschloe:
Never mind that. This is what happens if the tests are missing. :)

Do you have a file with periodic data handy? You could add it to the `gh-pages` branch in a separate PR and use it for testing as in `test_vtk.py`.

Issue #358, opened 2019-03-22
Issue with reading .msh file in meshio version 2.3.6


athulshaji:
I am not able to read in the file "cube.msh" while using the meshio version 2.3.6. The mesh is attached here as a zip file. This worked fine in meshio versions 2.3.5 and 2.3.4
[cube.zip](https://github.com/nschloe/meshio/files/2994956/cube.zip)

gdmcbain:
Do you know what version of Gmsh generated this?

The current stable documentation for [MSH file format] says:

> The MSH file format version 4 (current revision: version 4.1)

so this file is from the short unhappy period between the introduction of MSH4 and the clearing up of its teething troubles in MSH4.1.

Any chance of your regenerating it with current stable Gmsh? (Gmsh 4.2.2)

gdmcbain:
Wow, even Gmsh (4.2.3-git-7bc72f5de) can't read this mesh.

> Info    : Reading 'cube.msh'...
> Error   : Can only read MSH 4.0 format in ASCII mode
> Error   : Error loading 'cube.msh'
> Info    : Done reading 'cube.msh'

gdmcbain:
I have an old meshio 2.3.5 lying around and it does not read this file either:
> OverflowError: Python int too large to convert to C ssize_t

gdmcbain:
Ditto for meshio 2.3.3.

athulshaji:
Hello,

I am currently using meshio 2.3.5 and I am able to read in the file. Screenshot of my desktop is given below. I am using Windows 10, Python 3.7.2.
am also not able to read in the .msh using GMSH 4.2.2.  Can you help with the command for converting the mesh from one format to the other? The command given in the readme file is not working.

![image](https://user-images.githubusercontent.com/40489289/54802160-1bfbd000-4c90-11e9-9105-1ec37e9edf20.png)

gdmcbain:
> Can you help with the command for converting the mesh from one format to the other?

Yes, normally it's [`gmsh -convert cube.msh`](http://gmsh.info/doc/texinfo/gmsh.html#index-_002dconvert-files), but that's not working with Gmsh 4.2.2 or 4.2.3-git-7bc72f5de and this cube.msh.  

Or do you mean the `meshio-convert` command-line utility?  I don't think that'll work; it's going to rely on `meshio.read` internally.

athulshaji:
Hello, I was pointing to the command meshio-convert.

gdmcbain:
[...]

Issue #332, opened 2019-02-27
write MSH4.1


gdmcbain:
The reading of MSH4.1 #330 is implemented in #331 but not the writing.

Writing is less pressing than reading for MSH4.1 since it is generated by Gmsh by default but not read by much except Gmsh and meshio (with #331).

gdmcbain:
This isn't going so well.  I don't understand how to write the `$Entities` block.  The spec says that
>  The $Entities section is now optional. 

but this isn't always the case, Gmsh 4.2.0-git-0d5d49207 sometimes complaining 
> Error   : Unknown entity 2 of dimension 1

nschloe:
Best prepare and example and file a bug over at gmsh's gitlab. Tag me, too.

nschloe:
Fixed by #435.

gdmcbain:
An error like that reported above:

```
Error : Unknown entity 2 of dimension 1
```

arose again, after the closing of this issue by #435, in #470:

```
Error   : Unknown entity 0 of dimension 3
```

PR #365, opened 2019-03-22
Gmsh: Various fixes and refactor

dalcinl:
* Fix read/write for periodic section.
* Fix bad data types for v4.0 reader.
* Platform independent (32/64 bits) v4.1 reader.
* Get rid of ctypes data types in favor of NumPy.

### DIFF
--- a/meshio/msh_io/common.py
+++ b/meshio/msh_io/common.py
@@ -1,10 +1,17 @@
 # -*- coding: utf-8 -*-
 #
-from ctypes import c_int, c_double
 import logging
 import shlex
 
 import numpy
+from ..common import (  # noqa F401
+    num_nodes_per_cell,
+    cell_data_from_raw,
+    raw_from_cell_data,
+)
+
+c_int = numpy.dtype("i")
+c_double = numpy.dtype("d")
 
 
 def _read_physical_names(f, field_data):
@@ -20,33 +27,7 @@ def _read_physical_names(f, field_data):
     return
 
 
-def _read_periodic(f):
-    periodic = []
-    num_periodic = int(f.readline().decode("utf-8"))
-    for _ in range(num_periodic):
-        line = f.readline().decode("utf-8")
-        edim, stag, mtag = [int(s) for s in line.split()]
-        line = f.readline().decode("utf-8").strip()
-        if line.startswith("Affine"):
-            transform = line
-            num_nodes = int(f.readline().decode("utf-8"))
-        else:
-            transform = None
-            num_nodes = int(line)
-        slave_master = []
-        for _ in range(num_nodes):
-            line = f.readline().decode("utf-8")
-            snode, mnode = [int(s) for s in line.split()]
-            slave_master.append([snode, mnode])
-        slave_master = numpy.array(slave_master, dtype=int).reshape(-1, 2)
-        slave_master -= 1  # Subtract one, Python is 0-based
-        periodic.append([edim, (stag, mtag), transform, slave_master])
-    line = f.readline().decode("utf-8")
-    assert line.strip() == "$EndPeriodic"
-    return periodic
-
-
-def _read_data(f, tag, data_dict, int_size, data_size, is_ascii):
+def _read_data(f, tag, data_dict, data_size, is_ascii):
     # Read string tags
     num_string_tags = int(f.readline().decode("utf-8"))
     string_tags = [
@@ -177,21 +158,6 @@ def _write_physical_names(fh, field_data):
             fh.write('{} {} "{}"\n'.format(*entry).encode("utf-8"))
         fh.write("$EndPhysicalNames\n".encode("utf-8"))
     return
-
-
-def _write_periodic(fh, periodic):
-    fh.write("$Periodic\n".encode("utf-8"))
-    fh.write("{}\n".format(len(periodic)).encode("utf-8"))
-    for dim, (stag, mtag), transform, slave_master in periodic:
-        fh.write("{} {} {}\n".format(dim, stag, mtag).encode("utf-8"))
-        if transform is not None:
-            fh.write("{}\n".format(transform).encode("utf-8"))
-        slave_master = numpy.array(slave_master, dtype=int).reshape(-1, 2)
-        slave_master = slave_master + 1  # Add one, Gmsh is 0-based
-        fh.write("{}\n".format(len(slave_master)).encode("utf-8"))
-        for snode, mnode in slave_master:
-            fh.write("{} {}\n".format(snode, mnode).encode("utf-8"))
-    fh.write("$EndPeriodic\n".encode("utf-8"))
 
 
 def _write_data(fh, tag, name, data, write_binary):
--- a/meshio/msh_io/main.py
+++ b/meshio/msh_io/main.py
@@ -18,8 +18,15 @@ def read_buffer(f):
     # The various versions of the format are specified at
     # <http://gmsh.info/doc/texinfo/gmsh.html#File-formats>.
 
-    line = f.readline().decode("utf-8")
-    assert line.strip() == "$MeshFormat"
+    line = f.readline().decode("utf-8").strip()
+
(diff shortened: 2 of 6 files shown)


dalcinl:
@nschloe It was not fun, many small issues here and there. I confirm the readers work for v2.2, v4.0, and v4.1 both ASCII and binary (for meshed generated in Linux 64 bits).

I do have a set of of very small meshes with periodic section, in all formats and binary 32/64 within the [PETSc repository](https://github.com/petsc/petsc/tree/master/share/petsc/datafiles/meshes), look for `gmsh-3d-{ascii|binary}-[{32|64}].msh[24]`. However, I have no clue how your `gh-pages` trick works.

Eventually, It would be nice to add a proper writer for v4.1, but the handling Gmsh entities requires some refactor (e.g. add a `gmsh_entity` attr to `Mesh`), plus figuring out how to synthesize the entity section from alien formats.

dalcinl:
@nschloe The version map `{..., "4": msh4_1}` is OK for the writer. However, IMHO, for the reader it should be `{..., "4": msh4_0}`. Suppose you download a legacy Gmsh app version that dumps the outdated v4.0, it writes a plain `4` in the header, so if you try to read, boom! Would you agree to such a fix? 

Or maybe, given than v4.0 is outdated, and it was superseded by v4.1 within a few months (half a year), we should just kill the v4.0 reader/writter for good and move forward. Who is going to waste its time to implement a writer for the defunct v4.0?

nschloe:
> The version map {..., "4": msh4_1} is OK for the writer. However, IMHO, for the reader it should be {..., "4": msh4_0}. Suppose you download a legacy Gmsh app version that dumps the outdated v4.0, it writes a plain 4 in the header, so if you try to read, boom! Would you agree to such a fix?

I'm not sure right now: Does Gmsh 4.2 (with it's output format 4.1, _sheesh_) also put a plain 4 in the header?

> Or maybe, given than v4.0 is outdated, and it was superseded by v4.1 within a few months (half a year), we should just kill the v4.0 reader/writter for good and move forward. Who is going to waste its time to implement a writer for the defunct v4.0?

One big problem is that gmsh 4.1 is in Ubuntu, and people will use it for a while and produce meshes with it, even if it's outdated. And those meshes will be around forever, so we'll have to have a _reader_ for it. We can perhaps ditch the writer in the the future, but since we have a reader anyways, it doesn't hurt too much.

nschloe:
The gh-pages thing isn't a trick really: It's just a file storage that can be used to fetch meshes from. The tests (in master) then just download them from there, try to read them, and assert some checksums on the point data (or whatever you want). It was @gdmcbain's idea and I like it a lot since now for the first time we're actually testing that the reader does the right thing. Could be handy in the case of gmsh especially.

nschloe:
Btw the `square_bin.msh` in https://github.com/petsc/petsc/tree/master/share/petsc/datafiles/meshes has 755 permissions.

gdmcbain:
> It was @gdmcbain's idea 

Actually I was meaning to ask for the rationale behind storing the authoritative examples in a nonmaster branch and then having the tests download them from the WWW each time the test is run. My idea was simply having a set of genuinely generated meshes; I hadn't specified where to store them. I had been imagining keeping them in the master branch but in a directory that would not be installed by a regular pip-install, i. e. only by those git-cloning, presumably developers. I assume the test directory is like this? I'm not really worried about this but was curious.

dalcinl:
> Does Gmsh 4.2 (with it's output format 4.1) also put a plain 4 in the header?

This is the last version released to produce outdated format v4.0 (writes **4** in header):
```
$ gmsh-4.1 --version
4.1.5
$ gmsh-4.1 -v 0 -3 tmp.geo && head -n 3 tmp.msh 
$MeshFormat
4 0 8
$EndMeshFormat
```

This is the LATEST version, which produces format v4.1 (writes **4.1** in header):
```
$ gmsh-4.2 --version
4.2.2
$ gmsh-4.2 -v 0 -3 tmp.geo && head -n 3 tmp.msh 
$MeshFormat
4.1 0 8
$EndMeshFormat
[dalcinl@kw60439 tmp]$ 
```

So, you see, we really need to map `{"4" : mesh4_0}` for the reader. So you mentioned Ubuntu providing gmsh-4.1, assuming this version writes "4" in the header (didn't check, I do not use Ubuntu, maybe they patched gmsh to write "4.0"?),  then `meshio` will try to read as format v4.1, and fail.

dalcinl:
> Btw the `square_bin.msh` in https://github.com/petsc/petsc/tree/master/share/petsc/datafiles/meshes has 755 permissions.

Noted, thanks. Pull-request submitted in Bitbucket.

dalcinl:
> > It was @gdmcbain's idea
> 
> Actually I was meaning to ask for the rationale behind storing the authoritative examples in a nonmaster branch and then having the tests download them from the WWW each time the test is run.

Oh, I see now. It makes simply no sense! These tests meshes should just be there in master under let say `test\meshes\msh`.

gdmcbain:
> Eventually, It would be nice to add a proper writer for v4.1, but the handling Gmsh entities requires some refactor (e.g. add a `gmsh_entity` attr to `Mesh`)

While I did myself stumble at the hurdle of entities on trying to write MSH4.1 #332, I do hope that it will be possible to avoid introducing another format-specific attribute into `meshio.Mesh`.

dalcinl:
@nschloe OK, at this time I gently request move forward. Either explicitly accept my proposal for mapping 4->4.0 on read, and then I'll update this PR, or reject it [and live with the consequences :smiley:] and merge this PR as is.

gdmcbain:
> mapping 4->4.0 on read

I had prepared a PR for this on Friday while working on #358. I didn't end up pushing it as the OP was able to wipe the defective Gmsh 4.1, upgrade to Gmsh 4.2, and regenerate the mesh as MSH4.1. This is the preferred solution, but unhappily it does seem that Gmsh 4.1 has some currency; that's really unfortunate news that Ubuntu are shipping it. I don't use that distro much myself but it's the basis for many containers and cloud images.

dalcinl:
@gdmcbain Anyway, the 4.1 reader had some minor issues here and there. The fixes in this PR should resolve them all, hopefully.

dalcinl:
> > mapping 4->4.0 on read
> 
> I had prepared a PR for this on Friday while working on #358.

@gdmcbain So do do you agree that mapping 4->4.0 on read is the right thing to do?

nschloe:
> So do do you agree that mapping 4->4.0 on read is the right thing to do?

I think we all agree here, [this](https://github.com/nschloe/meshio/pull/365#issuecomment-475858119) is very good evidence. If you could update the PR, I'll merge right away.

> Oh, I see now. It makes simply no sense! These tests meshes should just be there in master under let say test\meshes\msh.

The idea of not storing meshes in the master branch comes from bad experiences I've have with Trilinos. There, all test data is stored in master, and the meshes used there are _huge_. Not only does everyone who ever uses Trilinos have to download the meshes, even if having no intention of running the tests. What's worse is that the meshes are forever engraved in git history, so pulls are terribly slow even if the meshes are removed now.
Lesson from that: No heavy data in code branches. (The same applied to images etc.)

I do realize that this is a bit inconvenient, e.g., in having to file two separate PRs for master and gh-branches. Also, _so far_, our meshes are relatively small. I might change my mind in the future about this, but let's be conservative for now.

dalcinl:
@nschloe OK, two commits for unrecognized sections, do not merge just yet. I'm starting to work in the 4->4.0 mapping on read.

nschloe:
> do not merge just yet. 

Best say `[WIP]` in the title.

dalcinl:
@nschloe Ready to go from my side. I've tested the reader for the cross product {2.2,4.0.,.1}x{ascii,binary} plus some exported meshes (from a commercial mesh) generator having `$Comments` sections everywhere (even at the very beginning).

nschloe:
Thanks for a great PR, @dalcinl!

PR #371, opened 2019-03-29
Abaqus: using a Python `dict` to accelerate cell reading

tianyikillua:
In Abaqus (but also for NASTRAN for which I developed also a reading script, maybe to be included into meshio?) each point is also identified with a grid id. The cell definition relies on these grid id's, and not on the 0-based Python indices.

This PR suggests that using a Python `dict` as a map `node grid id -> Python index of the point` is more efficient than the current implementation.

[disk.zip](https://github.com/nschloe/meshio/files/3023799/disk.zip)

Current implementation
```
%timeit meshio.read("disk.inp")
32.2 s ± 898 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This PR
```
%timeit meshio.read("disk.inp")
5.91 s ± 54 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/meshio/abaqus_io.py
+++ b/meshio/abaqus_io.py
@@ -120,7 +120,7 @@ def read_buffer(f):
             elif word.startswith("NODE"):
                 points, point_gids = _read_nodes(f)
             elif word.startswith("ELEMENT"):
-                key, idx = _read_cells(f, word)
+                key, idx = _read_cells(f, word, point_gids)
                 cells[key] = idx
             elif word.startswith("NSET"):
                 params_map = get_param_map(word, required_keys=["NSET"])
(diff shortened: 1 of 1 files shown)


tianyikillua:
Currently this PR will fail if the element block is defined before the node block. Is that possible?

nschloe:
> but also for NASTRAN for which I developed also a reading script, maybe to be included into meshio

Definitely, why not.

> Currently this PR will fail if the element block is defined before the node block. Is that possible?

LGTM. If you'd like to make other improvements, feel free. Abaqus mesh I/O hasn't been touched in a while.

PR #389, opened 2019-05-27
Fix for 2D meshes converted with gmsh

j-zimmermann:
When creating a mesh for instance in SALOME and exporting it in MED format, it can be converted to msh format by using gmsh. Then all the physical and geometrical tags are kept.
If the geometry is a 2D geometry, there will be still 3 coordinates (with z being equal to 0). That makes meshio think, the mesh is 3D. I do not know whether this is also the case with STEP geometries meshed in gmsh or native gmsh geometries and meshes, but it could well be.

With the fix here, this is circumvented and the points are returned in the right format, such that the information is also correct in other export formats (e.g. xmdf).
I use FEniCS with xdmf and there without the fix the workflow I prefer (CAD + mesh: SALOME, conversion MED to msh: gmsh, conversion to xdmf: meshio) would not work.

### DIFF
--- a/meshio/msh_io/common.py
+++ b/meshio/msh_io/common.py
@@ -212,3 +212,13 @@ def _write_data(fh, tag, name, data, write_binary):
 
     fh.write("$End{}\n".format(tag).encode("utf-8"))
     return
+
+
+def check_whether_2D(points):
+    # determine, whether 2D data
+    # compare to z component of first element. if z is always the same, the mesh is 2D
+    reference = points[0][2]
+    if all(point == reference for point in points[:, 2:]):
+        # make it 2D
+        points = points[:, :2]
+    return points
--- a/meshio/msh_io/msh2.py
+++ b/meshio/msh_io/msh2.py
@@ -20,6 +20,7 @@ from .common import (
     _write_physical_names,
     _read_data,
     _write_data,
+    check_whether_2D
 )
 
 c_int = numpy.dtype("i")
@@ -105,6 +106,7 @@ def _read_nodes(f, is_ascii, data_size):
 
     line = f.readline().decode("utf-8")
     assert line.strip() == "$EndNodes"
+    points = check_whether_2D(points)
     return points
 
 
--- a/meshio/msh_io/msh4_0.py
+++ b/meshio/msh_io/msh4_0.py
@@ -21,6 +21,7 @@ from .common import (
     _write_physical_names,
     _read_data,
     _write_data,
+    check_whether_2D
 )
 
 c_int = numpy.dtype("i")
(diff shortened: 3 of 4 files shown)


nschloe:
Not a huge fan of this.

> If the geometry is a 2D geometry, there will be still 3 coordinates (with z being equal to 0). That makes meshio think, the mesh is 3D.

It _is_ 3D then.

meshio should not have any logic in it that tinkers with the input data. If the input data has three coordinates (even if they all happen to sit in one plane), then that's what you want in the output. If you remove the third component, you're stepping on the toes of everyone who _do_ want that component.

What you could do instead:

 * Make Salome create a mesh with 2 spatial coordinates.
 * File a bug against Salome if this isn't possible
 * Put that bit of extra logic into your own converter (just use `meshio.read` and `write`)

j-zimmermann:
Well, I probably forgot to mention sth. I just came across when looking at the CircleCI result:
I captured gmsh's default behaviour. See e.g. the warning in the write routine:
* meshio/msh_io/msh2.py, line 270
* meshio/msh_io/msh4_0.py, line 271


This clearly shows that it is not intentionally 3D.
IMO, it makes a difference. If you do not want to merge the PR, I'm also fine with that.
Currently, I do not see a simpler workaround. If you find one, I'd be happy if you could let me know :)

gdmcbain:
Those warnings and the appending of the _z_ are correct and appropriate behaviour for writing in Gmsh's MSH format, which always has to have exactly three coordinates for each node.   As @nschloe writes, this is the right thing for meshio to do; it means that one- or two-dimensional applications can instantiate a `meshio.Mesh` without having to append the _z_ themselves, which may be unnecessary, e.g. if the subsequent output is XDMF.  See, e.g., 

https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/skfem/mesh/mesh.py#L244-L274

For a workaround, along the lines of the suggestion 'put that bit of extra logic into your own converter', see, e.g.

https://github.com/kinnala/scikit-fem/blob/9d088f003502405db55c259e5036e1c9d9ca567b/skfem/mesh/mesh.py#L336-L357

and, e.g.,

https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/skfem/mesh/mesh2d/mesh2d.py#L186-L188

scikit-fem does want to read Gmsh MSH but doesn't want to carry around zero _y_ or _z_ coordinates in lower dimensional problems.

Longer term, for your workflow, wouldn't it be better to be able to use meshio to go directly from MED to XDMF, bypassing MSH?  I had thought that that might need a bit of tweaking here (e.g. for keeping the physical tags, as you mentioned; see #290) but in #347 it was reported that &lsquo;it couldn't work better&rsquo;.  I don't have SALOME installed myself.

j-zimmermann:
Thank you very much, @gdmcbain!
I agree that this is definitely the better solution.

Issue #388, opened 2019-05-20
MSH 4.1: nodes have wrong tags in some cases


mrambausek:
When reading the msh-file attached, the node labels (tags) are ignored, leading to silently corrupted data. In the example attached, this means that, e.g., node 11 ends up as node 6.
[reading_failure.msh.zip](https://github.com/nschloe/meshio/files/3196818/reading_failure.msh.zip)

mrambausek:
The issue is related to these lines of code in ``msh_io/msh4_1.py`` in function ``_read_nodes``.
```
if is_dense:  
    fromfile(f, c_size_t, num_nodes)  
else:  
    tags[ixx] = fromfile(f, c_size_t, num_nodes) - 1  
```
The file attached in the original description is identified as dense, but doe not have uniformly increasing tags.

gdmcbain:
Thank you for providing a nice small example.  I'm looking into it now.  Could you tell us how it was generated?  From a .geo file?  By which version of Gmsh?  Thanks.

gdmcbain:
I see, yes, the nodes in the MSH file are numbered from `minNodeTag == minElementTag == 1` to `numNodes == maxNodeTag == 29`, but not in order, rather 1, 2, 3, 4, 5, 11, 12, 6, 13, 14, 7, 15, 16, 8, 17, 18, 9, 10, and then 19 through 29.  I note that this is permitted in the [specification](http://gmsh.info/dev/doc/texinfo/gmsh.html#MSH-file-format):
> All the node, element and entity tags (their global identification numbers) should be strictly positive. (Tag 0 is reserved for internal use.) Important note about efficiency: tags can be "sparse", i.e., do not have to constitute a continuous list of numbers (the format even allows them to not be ordered).

however, it does go on to say that

> By default, for non-partitioned, single file meshes, Gmsh will create files with a continuous ordering of node and element tags, starting at 1.

Do you know how it came about that the node tags aren't ordered here in this non-partitioned single-file mesh?  Coping with this might require a substantial change in `meshio.msh_io.msh4_1`.

I suppose it's required that what Gmsh calls node 11 turns up as `.points[11 - 1]` in the `Mesh` returned by `meshio.read`?  Rather than merely that the elements connected to node 11 turn up as cells connected to a point at the same coordinates, viz. `[.25, 0., 0.]`)?

mrambausek:
Okay, it is a very nice subtle thing:
The .geo file consists of three instructions:
[...]

PR #390, opened 2019-05-28
read MSH 4.1 node tags and populate points accordingly

gdmcbain:
Previously, the node tags in MSH 4.1 were just discarded, on the assumption, or rather following the [MSH file format](http://gmsh.info/dev/doc/texinfo/gmsh.html#MSH-file-format) specification, that

> By default, for non-partitioned, single file meshes, Gmsh will create files with a continuous ordering of node and element tags

As reported in #388, this isn't quite true, failing if in Gmsh's graphical user interface one generates a mesh and then increases its order (e.g. adding nodes on the midpoints for P2 elements): the first order nodes keep their old numbers and the new ones are given numbers in a higher range, but then they are listed in the output .msh file out of order.

The simple remedy proposed here is to read the tags and use them as indices of where to store them in the `points` array.

### DIFF
--- a/meshio/msh_io/msh4_1.py
+++ b/meshio/msh_io/msh4_1.py
@@ -128,7 +128,6 @@ def _read_nodes(f, is_ascii, data_size):
     num_entity_blocks, total_num_nodes, min_node_tag, max_node_tag = fromfile(
         f, c_size_t, 4
     )
-    is_dense = min_node_tag == 1 and max_node_tag == total_num_nodes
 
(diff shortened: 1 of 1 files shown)


nschloe:
Can you add a link to the issue in the code? Otherwise we might overlook in the future why we did it this way.

gdmcbain:
Good idea.  I'll take the opportunity to document the `_read_nodes` method.

nschloe:
Thanks @gdmcbain!

Issue #328, opened 2019-02-12
Read in Vtk


iagyonwu:
Hello I am using meshio to read a large vtu or vtk file.    My vtu file contains multiple Piece, reading it with paraview is successful. But when reading with meshio, the following error comes:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/iagyongx/.local/lib/python3.6/site-packages/meshio/helpers.py", line 174, in read
    return format_to_reader[file_format].read(filename)
  File "/home/iagyongx/.local/lib/python3.6/site-packages/meshio/vtu_io.py", line 274, in read
    reader = VtuReader(filename)
  File "/home/iagyongx/.local/lib/python3.6/site-packages/meshio/vtu_io.py", line 147, in __init__
    assert piece is None, "More than one Piece found."
AssertionError: More than one Piece found.
```
Then I converted the vtu data into vtk, i.e export it from paraview as vtk. The following error comes:
```
 File "<stdin>", line 1, in <module>
  File "/home/iagyongx/.local/lib/python3.6/site-packages/meshio/helpers.py", line 174, in read
    return format_to_reader[file_format].read(filename)
  File "/home/iagyongx/.local/lib/python3.6/site-packages/meshio/vtk_io.py", line 113, in read
    out = read_buffer(f)
  File "/home/iagyongx/.local/lib/python3.6/site-packages/meshio/vtk_io.py", line 226, in read_buffer
    assert section == "FIELD", "Unknown section '{}'.".format(section)
AssertionError: Unknown section 'METADATA'.
```
Can you give me any hint, how should I proceed?  

Here attached is an example vtk file:
https://drive.google.com/file/d/1RhfqMS6FKOyazuhuwbL3Rkt19ZX0vATf/view

gdmcbain:
The offending example linked to is quite large.

> All.vtk (5,8G) dépasse la taille maximale que Google peut analyser. Voulez-vous vraiment télécharger ce fichier ?

Any chance of coming up with a [minimal, complete, and verifiable example](https://stackoverflow.com/help/mcve)?

nschloe:
+1 to what Geordie says. I cannot even open the example file.

nschloe:
Closing for lack of feedback. Feel free to reopen if you have more info though!

iagyonwu:
Hi nschloe,
[...]

Issue #130, opened 2017-09-27
Vtk file can't be read if there is only mesh points (no associate data)


L4student:
I have a structured mesh as a [vtk file](https://github.com/nschloe/meshio/files/1338416/mesh.tar.gz) that contains only mesh points and there is no associated data in that file besides the nodes. Using `meshio.read('mesh.vtk')` gives the following error that ties to line 123 in `vtk_io.py` because it seeks `GetData()` in a file that does not have any:
```
# Explicitly extract points, cells, point data, field data
    points = numpy.copy(numpy_support.vtk_to_numpy(
            vtk_mesh.GetPoints().GetData()
            ))
```

Any suggestion? Could you include the capability to read just the mesh file? I was interested in converting vtk mesh to gmsh...so, if you think there will be further problems during conversion because of missing associated data then I would be interested to know. Thanks.

nschloe:
All of meshio's data types revolve around _unstructured_ meshes, so no wonder this doesn't work. 

> Could you include the capability to read just the mesh file? 

I guess it would be possible to get something to work, but that part would be completely disconnected from everything else in meshio since there is no other structured mesh type I know of. Given that, I guess a reader could be another standalone program. Your best bet for now would be to take some inspiration from `vtk_io.py` in meshio and write it yourself.

Issue #128, opened 2017-09-22
Converting a structured mesh with nodes and cells to vtk?


L4student:
Hi Nico,

May I know how to convert a structured mesh with an array of cells and nodes to a vtk file?
Let's say I have a mesh with `30x30x20` cells, so the arrays of cells and nodes are given as follows:
```
np.size(cellsarray) = (18000, 3)
np.size(nodesarray) = (20181, 3)
```
Thanks.

nschloe:
The only data type for structured meshes I know of is VTK, that's why there is no such capability (yet?) in meshio. I could imagine that this would be useful though for I/O in Python.

If your mesh is structured though, why do you have a `cellsarray`?

L4student:
That's true that for a structured mesh you don't really need, but I do have that information for any further mesh processing. I don't know the exact reason that information is being provided by the code I am looking at, but my guess is probably for use in some other part of that code. Anyways, I just want to be able to develop a vtk file if the above information is given for a structured grid. Thanks.

nschloe:
> but I do have that information for any further mesh processing.

That's good then: You can just store it as an unstructured mesh. That's what meshio is all about anyways.

L4student:
What command is used to do that? Any example that I can refer to? Thanks.

nschloe:
Basically [the first example in the README](https://github.com/nschloe/meshio).

L4student:
It's not clear what the pattern of the arguments should be. I saw that when I first came across meshio, and I believe, because of not knowing what pattern/shape those arguments actually follow led me to ask you this question. To figure that out I tried looking at the repository files, but I did not find the appropriate code that defines those arguments. Perhaps you can help clarify the definition and shape of those arguments? I have defined the shape of cells/nodes array above, but I am not sure whether I should just give them as it is. Also, are `point data` supposed to mean the `node data`?

L4student:
Well, there is no help on the parameters inside the `meshio.write` function. I tried looking through the code and implemented the following, but it's giving me errors (whether I give the `cells` and `points` data as `numpy` arrays or `lists`)

```
points_float = np.reshape(np.array(grid.nodelist), np.product(np.array(grid.nxyz) + 1)*3, order='C')
cells_float = np.reshape(np.array(grid.celllist), np.product(grid.nxyz)*3, order='C')

points = np.array([int(i) for i in points_float])
cells = np.array([int(i) for i in cells_float])
meshio.write(name_pfmesh_vtk, points, cells)
[...]

PR #399, opened 2019-07-09
VTK: add ability to read structured meshes and metadata

MuellerSeb:
In this pull request I add a number of structured DATASETs to the VTK-reader:

* STRUCTURED_POINTS
* STRUCTURED_GRID
* RECTILINEAR_GRID

they will be automatically converted internally to UNSTRUCTURED_GRID.

In addition, METADATA now doesn't anymore violate the reading process. Everything in METADATA will be ignored.

This tackles:
https://github.com/nschloe/meshio/issues/328
https://github.com/nschloe/meshio/issues/130
https://github.com/nschloe/meshio/issues/128

This could also be done for POLYDATA (but there, a lot of section come up: VERTICES, LINES, POLYGONS, TRIANGLE_STRIPS, that need to be converted to valid meshio cell types)

The same approach could be used for structured VTK XML files, but it get a bit more complicated and there, different file endings for every type exists (vti, vtp, vtr, vts), so the question arises if there should be one reader for each or not.

### DIFF
--- a/meshio/vtk_io.py
+++ b/meshio/vtk_io.py
@@ -291,7 +291,6 @@ def read_buffer(f):
         ]
         points = _generate_points(axis)
         c, ct = _generate_cells(dim=dataset["DIMENSIONS"])
-        pass
     elif dataset["type"] == "RECTILINEAR_GRID":
         axis = [
             dataset["X_COORDINATES"],
--- a/meshio/vtk_io.py
+++ b/meshio/vtk_io.py
@@ -276,15 +276,15 @@ def _read_sub_section(f, info):
         if info.section[1:] == "_COORDINATES":
             info.num_points = int(info.split[1])
             data_type = info.split[2].lower()
-            info.dataset[info.section] = _read_coords(
+            d[info.section] = _read_coords(
                 f, data_type, info.is_ascii, info.num_points
             )
         else:
-            info.dataset[info.section] = list(map(int, info.split[1:]))
+            d[info.section] = list(map(int, info.split[1:]))
             assert (
-                len(info.dataset[info.section]) == 3
+                len(d[info.section]) == 3
             ), "Wrong number of info in section '{}'. Need 3, got {}.".format(
-                info.section, len(info.dataset[info.section])
+                info.section, len(d[info.section])
             )
     elif info.section == "SCALARS":
         d.update(_read_scalar_field(f, info.num_items, info.split))
--- a/meshio/vtk_io.py
+++ b/meshio/vtk_io.py
@@ -278,6 +278,15 @@ def read_buffer(f):
                 assert section == "FIELD", "Unknown section '{}'.".format(section)
                 d.update(_read_fields(f, int(split[2]), is_ascii))
 
+    points, c, ct = _check_mesh(dataset, points, c, ct)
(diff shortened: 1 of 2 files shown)


meshio/vtk_io.py
```diff
@@ -208,8 +278,30 @@ def read_buffer(f):
...
+        dim = dataset["DIMENSIONS"]
+        ori = dataset["ORIGIN"]
+        spa = dataset["SPACING"] if "SPACING" in dataset else dataset["ASPECT_RATIO"]
+        axis = [
+            numpy.linspace(ori[i], ori[i] + (dim[i] - 1.0) * spa[i], dim[i])
+            for i in range(3)
+        ]
+        points = _generate_points(axis)
+        c, ct = _generate_cells(dim=dataset["DIMENSIONS"])
+        pass
```
MuellerSeb: This is a typo..

MuellerSeb:
Another question is, if it would be convenient to store the information about the structured mesh as field data, so that the information is not lost.

nschloe:
Nice addition. You'll need to invest more a little work to make the linter happy. Also, this definitely needs tests. You could just append some SMALL reference files here and I'll put them in the gh-pages branch from where you could use them for reading in tests.

MuellerSeb:
[struct_vtk.zip](https://github.com/nschloe/meshio/files/3378015/struct_vtk.zip)

Here are some test files. One for structured points, one for structured grid and three for rectilinear grid with different dimensions. All files contain meta_data.

I will try to satisfy the linter and add tests.

nschloe:
Okay, flake8 complains about
```
meshio/vtk_io.py:151:1: C901 'read_buffer' is too complex (20)
```
I kind of agree, but maybe this shouldn't matter too much now.  If you want, you can put
```
# noqa: C901
```
there.

MuellerSeb:
Do I need to update my branch, or is it ready for merging?

nschloe:
It's ready for _review_ now. :smile: Since it's substantial, I want to have a closer look. Will do so today or tomorrow.

MuellerSeb:
There is still a bug, since I assume ``ORIGIN`` and ``SPACING`` to be integer, but they have to be float.

MuellerSeb:
[00_image.vtk.zip](https://github.com/nschloe/meshio/files/3381136/00_image.vtk.zip)

here is a testfile for that case.

nschloe:
lgtm

nschloe:
Thanks for the fine addition, I'll release it with 3.0.0 later today.

MuellerSeb:
Yeah thanks!

Some last thoughts: 
Maybe it would be nice to store the ``dataset`` info dictionary as an attribute to the returned mesh container, so one could retrieve that information.

Maybe I can adopt the presented workflow for vtk-xml files.. that would require a reader for vti, vtr and vts each.

The POLYDATA type would be hard to adopt, since the required cell-types are commented out in the vtk_io.py file.. so this one will remain unsupported for a while.

Cheers,
Sebastian

nschloe:
Sure, if you find it useful then PR (or write an issue so this idea doesn't get lost).

MuellerSeb:
See: https://github.com/nschloe/meshio/issues/404

PR #400, opened 2019-07-10
Fix issue 34, reading .msh when it contains binary sections.

goodok:
Fix https://github.com/nschloe/pygalmesh/issues/34

When reading a .msh file version 4.1, when the file contains a section
in the form of binary lines (for example, $Parametrizations"), an error
occurs when reading them in utf-8 encoding.

This case is processed in this PR to fix it.
The same is probably for 4.0 MeshFromat, so it is fixed too.
Now example runs without errors^
```
import meshio
fn = 'ellipsoid.msh'
mesh = meshio.read(fn)
print(mesh)
```
Output:

```
<meshio mesh object>
  Number of points: 290
  Number of elements:
    vertex: 7
    line: 76
    triangle: 396
    tetra: 785
```

file: https://goodok.github.io/data/ellipsoid.msh

The flake8 check gave the result without errors in the style

### DIFF
--- a/meshio/msh_io/msh4_0.py
+++ b/meshio/msh_io/msh4_0.py
@@ -72,7 +72,14 @@ def read_buffer(f, is_ascii, data_size):
             # ```
             # skip environment
             while line != "$End" + environ:
-                line = f.readline().decode("utf-8").strip()
+                line = f.readline()
+                # Skip binary strings, but try to recognize text strings
+                # to catch the end of the environment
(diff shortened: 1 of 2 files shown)


nschloe:
Fair enough. Perhaps you could link this PR or the issue in the code as comment so I don't forget what this was for.

goodok:
Comments and references added.
(commit rebased)
Thank you.

PR #424, opened 2019-08-07
Update performance and add plotting

tianyikillua:
performance

Surely something should/could be done for Abaqus reading...

### DIFF
--- a/test/performance.py
+++ b/test/performance.py
@@ -1,3 +1,4 @@
+import tempfile
 import time
 
 import meshio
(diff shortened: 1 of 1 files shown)


nschloe:
I love the plots!!

nschloe:
> Surely something should/could be done for Abaqus reading...

The Abaqus format isn't well designed; the problem is that you don't know how many nodes you're reading unless go go through it line by line until you meet a `*`. That makes things slow, and I don't know how to fix it other than to update the format.

PR #425, opened 2019-08-07
Add methods to read and write field_data to XDMF

iitrabhi:
When creating a mesh in the meshing software gmsh we can tag different regions of the mesh with string tags.  This information is stored in the `.msh` file along with the int tags created by gmsh. When converting from `.msh` to `XDMF` the information regarding the string tags is lost. 

This PR adds methods to preserve the mapping between the string tags and int tags when converting from `.msh` to `XDMF`. The idea is to store this mapping inside the `<Information>` element of XDMF. 

This PR is according to the [XDMF Model and Format](http://www.xdmf.org/index.php/XDMF_Model_and_Format) and the [Xdmf.dtd](https://gitlab.kitware.com/xdmf/xdmf/blob/master/Xdmf.dtd#L62) and the results work well with paraview. As per the [XDMF schema](http://www.xdmf.org/index.php/XDMF_Model_and_Format#Information), the information is stored in CDATA inside the information element. Also, the XDMF reader of PARAVIEW enforces the use of “Name” and “Value” attributes. Thus, the information element is given the name “Information” and value is the number of string tags present in the model.


nschloe:
The changes are only in XDMF2, not XDMF3. Any reason for that?

iitrabhi:
I was testing it with XDMF3, so missed it. 
I have now made the required the changes to read_xdmf2() also. 
There is no difference in the definition of Information tag in both versions.

nschloe:
Nicely done, thank you!

Issue #409, opened 2019-07-26
Problem reading Abaqus Mesh correctly (`tetra10`)


bhaveshshrimali:
Is the `Abaqus` format fully supported in meshio ? 
I tried converting the attached mesh (comprised of 10-noded tets in 3D) using meshio but it spits the following error: 

```
  File "C:\Users\Bhavesh\AppData\Local\Programs\Python\Python37\Scripts\meshio-convert.exe\__main__.py", line 9, in <module>
  File "c:\users\bhavesh\appdata\local\programs\python\python37\lib\site-packages\meshio\cli.py", line 32, in main
    mesh = read(args.infile, file_format=args.input_format)
  File "c:\users\bhavesh\appdata\local\programs\python\python37\lib\site-packages\meshio\helpers.py", line 174, in read
    return format_to_reader[file_format].read(filename)
  File "c:\users\bhavesh\appdata\local\programs\python\python37\lib\site-packages\meshio\abaqus_io.py", line 95, in read
    out = read_buffer(f)
  File "c:\users\bhavesh\appdata\local\programs\python\python37\lib\site-packages\meshio\abaqus_io.py", line 122, in read_buffer
    key, idx = _read_cells(f, keyword, point_gids)
  File "c:\users\bhavesh\appdata\local\programs\python\python37\lib\site-packages\meshio\abaqus_io.py", line 182, in _read_cells
    entries = [int(k) for k in filter(None, line.split(","))]
  File "c:\users\bhavesh\appdata\local\programs\python\python37\lib\site-packages\meshio\abaqus_io.py", line 182, in <listcomp>
    entries = [int(k) for k in filter(None, line.split(","))]
ValueError: invalid literal for int() with base 10: ' \n'
```
Consequently I read the mesh using meshio in python and tried to remove the trailing newlines by naively replacing : 
```
entries = [int(k) for k in filter(None, line.split(","))]
```
with 
```
entries = [int(k) for k in filter(None, line.strip().split(","))]
```
inside `_abaqus.py` script. 

However this results in an incorrect `cells` dictionary. It is unable to read in fact element connectivity spanning multiple lines. Any ways to fix this ? I have not checked it with element connectivity that is printed "one line per element" though, but it should be fine after birefly peeking at the `_read_cells` function inside `_abaqus.py`

[mesh_trial.zip](https://github.com/nschloe/meshio/files/3437246/mesh_trial.zip)

keurfonluu:
[...]

PR #436, opened 2019-08-08
Abaqus: reconsider cells defined by multiple lines

tianyikillua:
I'm afraid 08497a6a3cfd44dc33ea63289077e0fb3142907e breaks again #409.

In case of continuation, the first chunk is part of the current cell definition.

Some continuation cases are added in the test file.


nschloe:
The new test file does indeed break master. Thanks so much @tianyikillua!

PR #438, opened 2019-08-08
FLAC3D I/O

keurfonluu:
This PR adds support to mesh grids generated by the commercial software FLAC3D (Itasca).

#### Features
1. Read and write FLAC3D grid in ASCII,
2. Similarly to Gmsh, zone groups are stored in ``cell_data`` under the keyword ``flac3d:zone`` with their names in ``field_data``,
3. Zone groups are exported if keyword ``flac3d:zone`` is present in ``cell_data``, which means that renaming ``gmsh:physical`` keywords (e.g. using dict's ``pop`` method) in ``cell_data`` to ``flac3d:zone`` allows to export physical groups defined in Gmsh as FLAC3D zone groups, for instance.

#### Missing features
1. Support for FLAC3D binary grid files (no documentation available),
2. Support for multiple group slots,
3. So far, only support simple tetra, pyramid, wedge and hexahedron.


nschloe:
Looks good to me except one thing: The example file was committed to git without adding it to GitLFS first, so it'll always be in git history. To avoid that, you'll first have to add the new extension to `.gitattributes` and _then_ commit the file. This way, it gets handled properly.

Would you mind creating a _new_ branch off of master and add your changes there?

nschloe:
Also, a link to the format description (if it exists) would be nice in the doc string.

keurfonluu:
Close this PR in favor of PR #439.

> Also, a link to the format description (if it exists) would be nice in the doc string.

Unfortunately, the documentation is only available within the software. I did not find any external resource documenting the format...

Issue #315, opened 2018-10-09
Meshio 2.3.1 fails while importing


jorgensd:
Installing meshio with the following command `pip install meshio==2.3.1` yields the following error when importing:
```

  File "/home/dokken/V_envs/dolfin_lfs/env/lib/python3.5/site-packages/meshio/cli.py", line 13, in <module>
    from .helpers import read, write, input_filetypes, output_filetypes

  File "/home/dokken/V_envs/dolfin_lfs/env/lib/python3.5/site-packages/meshio/helpers.py", line 185
    **kwargs,
            ^
SyntaxError: invalid syntax

```

This error is not present in `meshio==2.3.0`

TormodLandet:
Also seeing this with Python 3.5

```
> python3
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import meshio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/meshio/__init__.py", line 7, in <module>
    from . import cli
  File "/usr/local/lib/python3.5/dist-packages/meshio/cli.py", line 13, in <module>
    from .helpers import read, write, input_filetypes, output_filetypes
  File "/usr/local/lib/python3.5/dist-packages/meshio/helpers.py", line 185
    **kwargs,
            ^
SyntaxError: invalid syntax
>>> print(open('/usr/local/lib/python3.5/dist-packages/meshio/__about__.py').readlines()[2])
__version__ = "2.3.1"

```

PR #446, opened 2019-08-31
make work with python 3.5

tjhei:
should fix the problem in #315


nschloe:
Alright. Note though that we don't test against 3.5, so if the linter or so decides to revert this, I won't notice. Better upgrade to 3.6 at some point.

Issue #470, opened 2019-10-19
Wrong mesh total volume with MSH4 format


keurfonluu:
Hi,

I think there might be an error when reading Gmsh v4 format.

I have a box split as 6 horizontal boxes with different physical groups. A vertical box (a fault) cuts the 4 middle boxes. No issue whatsoever while meshing. I still check that everything is fine by calculating the total volume of my tetrahedra cells (it should equal 6e10 in my case).

With a mesh file exported as MSH4, I get a bigger volume. Nevertheless, the volume is correct when I export my mesh as MSH2. Both ``meshio.Mesh`` objects have the same number of tetra, but I noticed that the one for MSH2 has fewer points (5333 instead of 5337 for MSH4). I don't know if it helps, though.

I attach an archive containing the Gmsh script (actually generated by pygmsh), the two Gmsh mesh files (v2 and v4) and a script to check the total volume of the mesh:
[mesh4.zip](https://github.com/nschloe/meshio/files/3747262/mesh4.zip)

Note that everything is alright if I do not define physical groups.

nschloe:
Really interesting! For my own reference:
```python
import meshplex
import numpy

mesh = meshplex.read("mesh2.msh")
print(numpy.sum(mesh.cell_volumes))

mesh = meshplex.read("mesh4.msh")
print(numpy.sum(mesh.cell_volumes))
```
```
60000000000.00001
60004555829.20669
```
There seem to be differences in both cells and points. Can you produce a small-scale example?

keurfonluu:
Here is the downscaled example. I exported the file in ASCII with meshio too.
[mesh4_2.zip](https://github.com/nschloe/meshio/files/3748141/mesh4_2.zip)

I don't exclude the possibility that I am using Gmsh in the wrong way, but still, the error should be the same independently of the format...

gdmcbain:
I get an error reading `mesh4.msh` from `mesh4_2.zip` in Gmsh 4.5.0-git-0c1b63c8d on Ubuntu 19.04:

```
Error   : Unknown entity 0 of dimension 3
Error   : Could not read elements
Error   : Error loading 'mesh4.msh'
Info    : Done reading 'mesh4.msh'
```
How was this file generated?  The first `mesh4.msh` from `mesh4.zip` was O. K.

keurfonluu:
The mesh files in the second zip were generated by meshio. The original mesh file was generated using pygmsh.

[...]

PR #479, opened 2019-10-24
fix issue #470

keurfonluu:
Fixed issue #470.

``points`` is already sorted using ``tags``, there is no need to also sort ``data``. Not even sure whether ``tags`` is useful in ``_read_elements``.

### DIFF
--- a/meshio/_gmsh/msh4_1.py
+++ b/meshio/_gmsh/msh4_1.py
@@ -210,7 +210,7 @@ def _read_elements(f, point_tags, physical_tags, is_ascii, data_size):
         itags[point_tags] = numpy.arange(len(point_tags))
 
     # Note that the first column in the data array is the element tag; discard it.
-    data = [(physical_tag, tpe, itags[d[:, 1:] - 1]) for physical_tag, tpe, d in data]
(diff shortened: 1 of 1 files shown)


nschloe:
I don't understand the solution here. As far as I understand, the element list _does_ refer to the point tags, while in meshio it refers to their index in the array. That's why we have to use the tag inversion via `itags`. Can you let me know again what the point tags are, and how they are referred to in the elements list?

keurfonluu:
Thats's correct, the element list refers to the point tags. However, in ``_read_nodes``, ``points`` array is sorted by these tags.

```python
tags[ixx] = fromfile(f, c_size_t, num_nodes) - 1
points[tags[ixx]] = fromfile(f, c_double, num_nodes * 3).reshape((num_nodes, 3))
```

Thus, the element list also refers directly to ``meshio.Mesh.points`` point indices. I guess that it has been done this way to preserve the tag information of the points, but I could be wrong. Otherwise, we could simply read ``points`` as is and then remap ``data`` according to point tags (as it is done for ``abaqus`` format).

nschloe:
Closing in favor of #484 (which handles point tags correctly, even with sparse tags).

PR #481, opened 2019-10-24
Fix a problem with element sets

nilswagner:


nschloe:
I can merge when the linter is happy. `make lint`

nschloe:
Why do you try-catch and then raise, just to print the set_ids from the previous step?

nilswagner:
[tutorial.inp.txt](https://github.com/nschloe/meshio/files/3767635/tutorial.inp.txt)
mv tutorial.inp.txt tutorial.inp
meshio_convert tutorial.inp tutorial.dato
ValueError: invalid literal ...

nschloe:
I understand, but you'll get the same error message with the patch, don't you? After all, you `raise`  the exception again.

nilswagner:
My patch resolves the problem.
[tutorial.txt](https://github.com/nschloe/meshio/files/3767665/tutorial.txt)

nschloe:
Okay, then it's the
```
        if not line:
            break
```
part that does the trick. If the try-catch block is a debugging relic (and it looks like one), please remove it.

Issue #353, opened 2019-03-21
field_data for two sets of different dimensions but with a same name


tianyikillua:
Reading a `.msh` file containing two sets of different dimensions but with a same name leads to one being eaten by the other.

![](https://user-images.githubusercontent.com/4027283/54740343-65c6b680-4bbb-11e9-9b27-9a03feda2cc4.png)

Here `Top` is the name given to a 1d element set, and a 2d element set, however in `field_data` only that appears later will win.

Issue #493, opened 2019-10-30
Abaqus parser


dbeurle:
Thanks for the work on meshio!  Quite the task you've taken on ;-)

I'm trying to convert an Abaqus input to a gmsh file format but I am unable to get the element sets (`Elset`) and node sets (`Nset`) transferred into physical groups in the gmsh format.  I kinda need these for prescribing my boundary conditions.

What would you suggest to add in order to get these sets across?  The `Mesh` class doesn't seem to provide a mechanism for passing the element sets but does for `node_sets`.  However this does not translate to a physical group (of dimension 0) in gmsh.

Another issue:  the `node_sets` and `element_sets` do not retain the casing of the set name.  For example `applied_displacements` becomes `APPLIED_DISPLACEMENTS`, which is unexpected.  Are you happy to accept a PR to revert this behaviour?

Ideas on how to move forward?  I'm happy to do the grunt work.

nschloe:
As far as I remember, there is no support for sets in meshio yet. See https://github.com/nschloe/meshio/issues/249 which asks about the same.

What's a set (in abaqus terms) anyway? A collection of nodes, or are the nodes stores in one big array with an additional dictionary?

dbeurle:
They look like this:
```
*Nset, nset=applied_displacement
     2,     3,    60,    61,    62,    63,   728,   729,   786,   787,   788,   789,  1454,  1455,  1512,  1513
  1514,  1515,  2180,  2181,  2238,  2239,  2240,  2241,  2906,  2907,  2964,  2965,  2966,  2967,  3632,  3633
  3690,  3691,  3692,  3693,  4358,  4359,  4416,  4417,  4418,  4419,  5084,  5085,  5142,  5143,  5144,  5145
  5810,  5811,  5868,  5869,  5870,  5871,  6536,  6537,  6594,  6595,  6596,  6597,  7262,  7263,  7320,  7321
  7322,  7323,  7988,  7989,  8046,  8047,  8048,  8049,  8714,  8715,  8772,  8773,  8774,  8775,  9440,  9441
...
 58543,
*Elset, elset=applied_displacement
     2,     4,     6,     8,    10,   484,   486,   488,   490,   492,   966,   968,   970,   972,   974,  1448
  1450,  1452,  1454,  1456,  1930,  1932,  1934,  1936,  1938,  2412,  2414,  2416,  2418,  2420,  2894,  2896
[...]

PR #494, opened 2019-10-30
Support for named node and element sets for gmsh and Abaqus

dbeurle:
- Fix unexpected case change for named sets, see #493
- Add support for different dimensions in physicalNames, see #353

### DIFF
--- a/meshio/_mesh.py
+++ b/meshio/_mesh.py
@@ -10,6 +10,7 @@ class Mesh:
         cell_data=None,
         field_data=None,
         node_sets=None,
+        element_sets=None,
         gmsh_periodic=None,
         info=None,
     ):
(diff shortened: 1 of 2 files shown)


meshio/_mesh.py
```diff
@@ -10,6 +10,7 @@ def __init__(
         cell_data=None,
         field_data=None,
         node_sets=None,
+        element_sets=None,
```
dbeurle: @nschloe Why isn't `node_sets={}` used here instead of `None`.  I'm not a great python programmer.

gdmcbain: Mutable default arguments are a no-no in Python; essentially there's only one of them that gets shared by all calls to the function. For the classic explanation see:

https://effbot.org/zone/default-values.htm

nschloe:
How does this fix #353?

meshio/_mesh.py
dbeurle: Ha the more you know - thanks!

dbeurle:
> How does this fix #353?

It doesn't in its current state.  However, hopefully it will as both node and element sets have the same name and their equivalent in gmsh would have to accommodate multiple sets with different dimensions and the same name.

nschloe:
Fair enough.

PR #495, opened 2019-10-30
Improve FLAC3D writer performance

keurfonluu:
- Improved: vectorize calculation of scalar triple products to determine order of points (much faster),
- Fixed: issue when ``meshio.Mesh`` contains cell types not supported by FLAC3D,
- ~~Added: ``zone_data`` option to define ``cell_data`` to export as zone group (e.g. ``meshio.write("filename.f3grid", mesh, zone_data = "gmsh:physical")`` to export Gmsh physical groups as FLAC3D zone groups).~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
           command: pytest --cov meshio
           working_directory: test/
       # submit to codecov
-      # - run: bash <(curl -s https://codecov.io/bash)
+      - run: bash <(curl -s https://codecov.io/bash)
 
 workflows:
   version: 2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,5 +40,5 @@ jobs:
       run: |
         pip install pytest pytest-cov
         pytest --cov meshio
-    - name: Submit to codecov
-      run: bash <(curl -s https://codecov.io/bash)
+    # - name: Submit to codecov
+    #   run: bash <(curl -s https://codecov.io/bash)
--- a/meshio/__about__.py
+++ b/meshio/__about__.py
@@ -1,4 +1,4 @@
-__version__ = "3.2.5"
+__version__ = "3.2.6"
 __author__ = "Nico Schlömer"
 __author_email__ = "nico.schloemer@gmail.com"
 __copyright__ = "Copyright (c) 2015-2019 {} <{}>".format(__author__, __author_email__)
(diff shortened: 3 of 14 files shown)


nschloe:
> meshio.write("filename.f3grid", mesh, zone_data = "gmsh:physical")

Can't merge this. The `zone_data` keyword is too specific to the flac3d writer, it wouldn't do anything else for any other mesh type. Can you check in `mesh.cell_data` for suitables zones and translate those?

Also `np.cross` is really slow, and if you're looking for speed you almost always want to try and replace it with dot products. There is an expression for `(np.cross(p1 - p0, p2 - p0) * (p3 - p0)).sum(axis=1)` (the scalar triple product) in terms of _dot_ products only, see https://math.stackexchange.com/q/1954789/36678. I can't recall if that's any faster than the cross product now, but it's certainly worth trying.

tianyikillua:
Thanks for your contribution!

For performance (*much faster*), could you post the figure generated by https://github.com/nschloe/meshio/blob/master/test/performance.py, before and after the PR? You need to add `flac3d` into the list.

keurfonluu:
@nschloe Okay I removed the ``zone_data`` option and improved interoperability with Gmsh and Medit (based on what's already done for Medit format). I also improved the way labels were handled.

Concerning the scalar triple product, I need to know its sign to correctly reorder the points (that determines the direction of the normal vector in FLAC3D), so I cannot use the expression in your link... Actually, the main purpose of this PR is to avoid calling ``np.cross`` for each cell by replacing it with a single call to bigger ``numpy.array``s which significantly improves the performance of the writer (see below). ``np.cross`` seems to be quite optimized in the latest version of ``numpy`` and I think the only way to improve a bit more the performance is to avoid creating intermediate ``numpy.array``s when broadcasting (e.g. use ``cython`` or ``numba``).

@tianyikillua Unfortunately, I cannot run the test since FLAC3D does not support triangles. The only way would be to transform triangles to wedges and quads to hexahedra. Nevertheless, here are some benchmark results for a custom mesh with 573642 tetrahedra (new below).

```
25.6 s ± 139 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
7.07 s ± 49.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

nschloe:
> np.cross seems to be quite optimized in the latest version of numpy

Yeah, it's good enough. It's about five times slower than a `dot`, though, which is why I'd once looked into this. It's alright to use `cross`.

nschloe:
LGTM. FLAC3D is a rather obscure format, so I'm fine with an incomplete test coverage. Would be nice to have tests for this though.

PR #499, opened 2019-11-06
Allow XDMF meshes to be 1D.

cianwilson:
I'd like to use meshio to convert 1D gmsh meshes to xdmf.  This super simple change allows that but let me know if there are implications I haven't considered.

Thanks!


nschloe:
Thanks for the PR!

cianwilson:
Many thanks!

PR #505, opened 2019-11-11
pad locally rather than mutating 2D writing VTK

### LINKED ISSUE #504: write VTK mutates 2D Mesh argument to 3D
When VTK pads a two-dimensional `meshio.Mesh` with zero third coordinate to `write` it, it mutates the argument, leaving it three-dimensional thereafter in the caller.  This can be confusing and unwanted if the caller is using the two-dimensional `Mesh` data structure for other things.

```python
import meshio
import numpy

mesh = meshio.Mesh(
    numpy.array([[0, 0], [1, 0], [1, 1], [0, 1]]),
    {"triangle": numpy.array([[0, 1, 2], [0, 2, 3]])},
)

print('Before:', mesh.points.shape[1])
meshio.write('mutate23.vtk', mesh)
print('After:', mesh.points.shape[1])
```

gdmcbain:
Fixes #504

### DIFF
--- a/meshio/_vtk.py
+++ b/meshio/_vtk.py
@@ -603,18 +603,24 @@ def write(filename, mesh, binary=True):
             "VTK requires 3D points, but 2D points given. "
             "Appending 0 third component."
         )
-        mesh.points = pad(mesh.points)
+        points = pad(mesh.points)
+    else:
+        points = mesh.points
 
     if mesh.point_data:
+        point_data = {}
         for name, values in mesh.point_data.items():
             if len(values.shape) == 2 and values.shape[1] == 2:
                 logging.warning(
                     "VTK requires 3D vectors, but 2D vectors given. "
                     "Appending 0 third component to {}.".format(name)
                 )
-                mesh.point_data[name] = pad(values)
+                point_data[name] = pad(values)
+            else:
+                point_data[name] = values
 
     if mesh.cell_data:
+        cell_data = {}
         for t, data in mesh.cell_data.items():
             for name, values in data.items():
                 if len(values.shape) == 2 and values.shape[1] == 2:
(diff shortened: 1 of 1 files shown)


gdmcbain:
I think this is failing because of `test.helpers.write_read` on the two-dimensional examples.  They get written as three-dimensional and then read back as three-dimensional and since here they haven't been themselves mutated to three-dimensional, they don't pass the test.  I'm not sure whether this is the desired behaviour.

nschloe:
In `test/helpers.py`, use `in_mesh = copy.deepcopy(input_mesh)` at the beginning of `write_read` and use `in_mesh` instead of `input_mesh` in the rest of the file (except for the `writer` call).

gdmcbain:
I'm not sure that that will help. Under the PR, mutation of the input mesh on writing will be avoided, so it won't diverge from its deep copy which will therefore be redundant.

If a mesh has to be distorted in some way to be writable in some output format (e.g. embedding planar data in 3-space for VTK) perhaps the idea should be that either
* the input mesh is similarly distorted before comparison with its reread version or
* the reread version has the distortion undone.

I imagine the latter isn't always possible, but here what about only comparing those columns which are present in the input mesh?

Alternatively, the deep copy could be applied in the VTK writing function; however, that might slow all uses of an important function.

nschloe:
Not sure I understand what's going on. I'll take a stab at it tomorrow.

nschloe:
Ah okay. Well, add the deepcopy nonetheless and add an assertion that the input mesh isn't changed after write. This is something that should fail _before_ this PR.

For the output test, you could just take the `[:input_mesh.shape[1]]` columns of the output mesh. Btw I was hoping to be able to remove the column stacking altogether; see https://gitlab.kitware.com/paraview/paraview/issues/19058.

nschloe:
Closing in favor of #506.

Issue #509, opened 2019-11-13
Paraview crashes after importing converted .vtu/.vtk files from .f3grid (FLAC3D)


kuateric:
I converted a .f3grid file from FLAC3D to .vtu and .vtk. After important the file to Paraview to visualise the mesh, Paraview crashes and closes. Now I don't know if the problem is in the converted files or it is a issue with Paraview. 
Thank you

nschloe:
@keurfonluu Best append the meshes here. (I think GitHub wants you to zip them first though.)

kuateric:
@nschloe @keurfonluu here the mesh files: 
[testmodell3.zip](https://github.com/nschloe/meshio/files/3841653/testmodell3.zip)
Thank you for your help

tianyikillua:
The problem is with FLAC3D itself (either the mesh file, or reading this file).

Using https://github.com/tianyikillua/paraview-meshio-reader to read the original FLAC3D file, ParaView also crashes.

```
<meshio mesh object>
  Number of points: 68796
  Number of elements:
    wedge: 38
    hexahedron: 61142
  Cell data: flac3d:zone
```
So we only have (read) 68796 points, however
```
np.any(mesh.cells["hexahedron"] >= mesh.points.shape[0], axis=1).sum() = 1253
```
so we have 1253 hexahedron elements that refer to non-existing nodes!

nschloe:
Perhaps that's something we should assert against when writing a file.

kuateric:
Thank you @tianyikillua for your remark. I will check the file and try again. It looks like i found my problem. I will let you know soon.

nschloe:
Alright, there now is `meshio-info` which prints some details and issues if it found some.
```
meshio-info /tmp/testmodell3.f3grid 
```
```
<meshio mesh object>
  Number of points: 68796
  Number of elements:
    wedge: 38
    hexahedron: 61142
  Cell data: flac3d:zone

ATTENTION: Inconsistent mesh. Cells refer to nonexistent points.
```
Let's call this fixed.

tianyikillua:
> Let's call this fixed.

Not necessarily.

Like Abaqus / Nastran, each cell is defined using the unique but not **necessarily naturally ordered** id's of the nodes.

In this case, the last node is referred to with the id 69773, even though it is the 68796th and the last point of the mesh.

![](https://user-images.githubusercontent.com/4027283/68779223-9240ae80-0634-11ea-8b6d-416f9e3771bf.png)

[...]

PR #511, opened 2019-11-13
Fix issue #509

keurfonluu:
- Fixed: handle non-natural order point IDs.

### DIFF
--- a/meshio/_flac3d.py
+++ b/meshio/_flac3d.py
@@ -83,12 +83,15 @@ def read_buffer(f):
     zgroups = {}
     count = {k: 0 for k in meshio_to_flac3d_type.keys()}
 
+    index = 0
     line = f.readline()
     while line:
         if line.startswith("ZGROUP"):
             _read_zgroup(f, line, zgroups)
         elif line.startswith("G"):
-            _read_point(line, points, point_ids)
+            pid = _read_point(line, points)
+            point_ids[pid] = index
+            index += 1
         elif line.startswith("Z"):
             _read_cell(line, zones, count, point_ids)
         line = f.readline()
@@ -102,13 +105,13 @@ def read_buffer(f):
     )
 
 
-def _read_point(line, points, point_ids):
+def _read_point(line, points):
     """
     Read point coordinates.
     """
     line = line.rstrip().split()
     points.append([float(l) for l in line[2:]])
-    point_ids[int(line[1])] = len(point_ids)
+    return int(line[1])
 
 
 def _read_cell(line, zones, count, point_ids):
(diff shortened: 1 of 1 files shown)


nschloe:
`point_ids` should be _returned_ from `read_points`.

keurfonluu:
> `point_ids` should be _returned_ from `read_points`.

It's not possible. FLAC3D format has no separate point and cell sections (``* GRIDPOINTS`` and ``* ZONES`` are dummy comments). You are totally allowed to define a set of points, then a zone element, followed by another set of points, e.g.

```
G 1   2.10000000000000e+00   0.00000000000000e+00   3.10000000000000e+00
G 2   2.30564263324093e+00   0.00000000000000e+00   3.24739811913093e+00
G 3   2.10000000000000e+00   3.57711954132911e-01   3.10000000000000e+00
G 4   1.79145372108365e+00   0.00000000000000e+00   3.32724180720029e+00
G 5   2.30564263324093e+00   3.57711954132911e-01   3.24739811913093e+00
G 6   1.79145372108365e+00   3.57711954132911e-01   3.32724180720029e+00
G 7   1.96230278624208e+00   0.00000000000000e+00   3.47179049928151e+00
G 8   1.96230278624208e+00   3.57711954132911e-01   3.47179049928151e+00
Z B8 1 1 2 3 4 5 6 7 8
G 9   2.54644975339359e+00   0.00000000000000e+00   3.42000102750863e+00
G 10   2.54644975339359e+00   3.57711954132911e-01   3.42000102750863e+00
G 11   2.16236670641845e+00   0.00000000000000e+00   3.64105673417323e+00
G 12   2.16236670641845e+00   3.57711954132911e-01   3.64105673417323e+00
```

is a valid FLAC3D mesh file.

nschloe:
I see. I still don't like how the `point_ids` dict is manipulated "behind the scenes". I'd like it better if `read_points` returned something, and that something is then incorporated into `point_ids` explicitly in `read`.

nschloe:
Also, does it have to be a `dict`? A list of arrays (which are concatenated at the end) would be more efficient I guess. Not a must though.

keurfonluu:
> I see. I still don't like how the `point_ids` dict is manipulated "behind the scenes". I'd like it better if `read_points` returned something, and that something is then incorporated into `point_ids` explicitly in `read`.

Got you.

> Also, does it have to be a `dict`? A list of arrays (which are concatenated at the end) would be more efficient I guess. Not a must though.

It might be more complicated to implement, I guess. The problem is that the format does not provide any indication about the total number of points (or cells)...

nschloe:
Thanks!

Issue #466, opened 2019-10-15
[Request] Add support for `unv` and plot3d formats


s1291:
Hello,
Is it possible to add support for Ideas `unv` format and Plot3D `p3d` ?
Thank you

nschloe:
I've never heard of either of the formats. Please post links to the specifications.

s1291:
Hello, 
For Plot3D:
- From wikipedia: In computational fluid dynamics, the PLOT3D file format is a standard file format used for storing grid and results data. PLOT3D was created by NASA Ames Research Center in 1982 by Pieter Bunin.
- You can find the Plot3D user manual [here](https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19900013774.pdf) 

For UNV I could't find much information, But:
- It is supported in [Salome SMESH](https://docs.salome-platform.org/7/gui/SMESH/index.html) module.
- It is also available in [GMSH](http://gmsh.info/doc/texinfo/gmsh.html).

nilswagner:
All datasets wrt *.unv are available here. Just replace the corresponding number

http://sdrl.uc.edu/sdrl/referenceinfo/universalfileformats/file-format-storehouse/universal-dataset-number-2412

http://sdrl.uc.edu/sdrl/referenceinfo/universalfileformats/file-format-storehouse/universal-dataset-number-15
http://sdrl.uc.edu/sdrl/referenceinfo/universalfileformats/file-format-storehouse/universal-dataset-number-2411
http://sdrl.uc.edu/sdrl/referenceinfo/universalfileformats/file-format-storehouse/universal-dataset-number-18
http://sdrl.uc.edu/sdrl/referenceinfo/universalfileformats/file-format-storehouse/universal-dataset-number-2420

s1291:
Another source for `unv` is the `unvHandler` class from [PyGeM](https://github.com/mathLab/PyGeM/blob/master/pygem/unvhandler.py) library.
A tutorial that demonstrate the usage of [Free Form Deformation on a unv mesh on cylinder](https://github.com/mathLab/PyGeM/blob/master/tutorials/tutorial-3-unv.ipynb).

qrendu:
Hello,

It would be very nice to have plot3d format available for CFD community. As far as I understood, @s1291 proposed some code but it was rejected because an ASCII mesh file was too big? If it's still open I can try to start from there and push something.

3rav:
Please add .unv

2411 nodes
2412 elements
[...]

PR #515, opened 2019-11-17
Add basic support for PLOT3D mesh format.

s1291:
Adding `_plot3d.py` file to support the PLOT3D mesh format. Related #466


nschloe:
This PR introduces a huge file to the tree so I'll need to close this.

Best redo the PR by first adding the format to https://github.com/nschloe/meshio/blob/master/.gitattributes and _then_ adding the file. The file should be a lot smaller, too.

s1291:
@nschloe 
Which format? the p3d is an ascii file.

nschloe:
Yes, but it's HUGE. We use git-lfs for such files.

PR #517, opened 2019-11-17
[WIP] Add basic support for PLOT3D format

s1291:
Very basic support for PLOT3D file format.


meshio/_plot3d.py
```diff
@@ -0,0 +1,94 @@
+"""
+PLOT3D format is for structured grids only
+For more info, check out: PLOT3D User's Manual, chapter 8: Data File Formats
+Author: Samir OUCHENE
```
nschloe: We don't put author names in the files. Everyone can read up in the Git history who contributed what.

meshio/_plot3d.py
```diff
@@ -0,0 +1,94 @@
...
+"""
```
nschloe: Please put a link to the PLOT3D format specification.

meshio/_plot3d.py
```diff
@@ -0,0 +1,94 @@
...
+        cells = {"quad": cells_}
+        return Mesh(pts, cells)
+    else:
+        pts = np.loadtxt(filename, skiprows=1).reshape(
+            (imax * jmax * kmax, 3), order="F"
+        )
+        cells_ = []
+        for k in range(kmax - 1):
+            cell = []
+            for j in range(jmax - 1):
```
nschloe: All of these loops look like C code, and this will be SUPER slow.

meshio/_plot3d.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return Mesh(pts, cells)
+
+
+def write(filename, mesh):
+    """
+    Write the mesh to filename in PLOT3D format
+    """
+    points = mesh.points
+    imax = np.unique(points[:, 0]).size
```
nschloe: Calling `unique` is very slow, and it's not clear what it does for float values. We'll need to find another solution here.

meshio/_plot3d.py
```diff
@@ -0,0 +1,94 @@
...
+            print(imax, jmax, kmax, file=p3dfile)
+            for value in points.flatten(order="F"):
+                print(value, file=p3dfile)
+        else:
+            print(imax, jmax, file=p3dfile)
+            for value in points[:, 0:2].flatten(order="F"):
+                print(value, file=p3dfile)
+
+
+def test_plot3d():
```
nschloe: Tests are in `test/`.

nschloe:
How large is the test file?

meshio/_plot3d.py
s1291: Yes. Sorry, It was a mistake.

meshio/_plot3d.py
s1291: The wikipedia article does not contain much information. However, the format specification can be found in the user's manual: https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19900013774_1990013774.pdf

meshio/_plot3d.py
s1291: I can't find a way to get `imax`, `jmax`, and `kmax` other than using `np.unique`

meshio/_plot3d.py
s1291: Yes, I know, unfortunately I am not yet familiar with meshio project structure.

s1291:
I think PR is not yet ready. I have to make more tests.

meshio/_plot3d.py
s1291: @keurfonluu 
Yes, PLOT3D format supports only structured grids.

nschloe:
> I think PR is not yet ready. I have to make more tests.

I modified the title to say [WIP] (work in progress). You can edit it out once you think it's ready.

The code quality doesn't meet meshio's standards yet, but since you seem like a diligent programmer, I'll make some remarks and perhaps you can fix things accordingly. :)

meshio/_plot3d.py
nschloe: Most meshio tests do only one thing: Create a mesh, write it using the writer, read the created file using the reader and check if the result is the original mesh again. Check `test/test_off.py` for a basic example; you can take that over. It only uses `helpers.tri_mesh`; in other test files, you'll see more meshes.

meshio/_plot3d.py
nschloe: Oh, this can be quite difficult then. @MuellerSeb, do we have proper support for structured meshes in meshio yet? As far as I remember not.

meshio/_plot3d.py
nschloe: We'll need the link in the file, not here.

nschloe:
Unfortunately, I'll have to close this one again. You added the test files _before_ putting the respective files into .gitattributes. This means they will be in git  history, so everyone will have to download the mesh when cloning the repo _forever_. You'll have to redo the PR.

nschloe:
In fact, I'd recommend _not_  to use a test file for now. It's okay for testing locally on your machine, but rather use the write-read-tests of meshio as explained above.

PR #523, opened 2019-11-19
WIP: extend AttributeType to XdmfTimeSeries

### LINKED ISSUE #522: write vectors in XDMF time series
Previously vectors were not written as such to XDMF #327 but that was fixed in #333 but not for `XdmfTimeSeriesWriter`.  I think it should be a simple matter of extending #333 to the latter.

gdmcbain:
This is intended to fix #522.

### DIFF
--- a/meshio/_cli.py
+++ b/meshio/_cli.py
@@ -134,7 +134,7 @@ def info(argv=None):
 
 def _get_info_parser():
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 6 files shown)


gdmcbain:
I noticed that the instance method `XdmfWriter._attribute_type` doesn't refer to its first argument and so could be redefined as static; however, since that class is unrelated to `XdmfTimeSeriesWriter`, I think they can only share the code by both importing it as a module function from `common`.

PR #531, opened 2019-11-22
Use open instead of codecs.open

s1291:
codecs.open is obsolete in python 3.x. open in python 3.x has an
encoding argument. For more info, check out this [question](https://stackoverflow.com/questions/5250744/difference-between-open-and-codecs-open-in-python).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,3 @@
-import codecs
 import os
 
 from setuptools import find_packages, setup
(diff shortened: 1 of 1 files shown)


nschloe:
Actually you can remove the entire `read()` function and just put `open("README.md").read()` inline.

s1291:
> Actually you can remove the entire `read()` function and just put `open("README.md").read()` inline.

Yes of course. I just thought it might be more clear to let it as it is.

nschloe:
Cool, thanks.

PR #540, opened 2019-11-28
handle XDMF lines consistently and without mutation

### LINKED ISSUE #539: write XDMF mutates cell['line'] if other cells present
`write` to format XDMF mutates its argument `mesh`, changing the shape & content of `mesh.cells['line']`, but only if there are also other elements of `mesh.cells`; e. g.:

```python
from numpy import array

from meshio import Mesh, write

mwe =  Mesh(array([[0., 0., 0.],
                   [1., 0., 0.]]),
                   {'line': array([[0, 1]])})

print(mwe.cells['line'].shape)
write('mwe0.xdmf', mwe)
print(mwe.cells['line'].shape)
mwe.cells['vertex'] = array([[0]])
write('mwe1.xdmf', mwe)
print(mwe.cells['line'].shape)
```

The output:
```
(1, 2)
(1, 2)
(1, 3)
```

gdmcbain:
i.e. regardless of whether `Mesh.cells` has keys other than `'line'` and not prepending the required 2 to the `cells['line']` in the caller.

Fixes #539.


nschloe:
`make black` please.

nschloe:
Thanks, Geordie!

PR #551, opened 2019-11-28
Buffers

clbarnes:
Ensure that pathlikes can be read in python 3.5, and allow `read`/`write` methods to accept buffers as well as filenames.

Could use some tests and docs.


meshio/_files.py
```diff
@@ -0,0 +1,22 @@
...
+        "w" in mode and hasattr(obj, "write")
+    )
+
+
+@contextmanager
+def open_file(path_or_buf, mode="r"):
+    if is_buffer(path_or_buf, mode):
+        yield path_or_buf
+    elif sys.version_info < (3, 6) and isinstance(path_or_buf, os.PathLike):
+        # todo: remove when 3.5 is EoL
```
nschloe: Best explicitly say 2020-09-13 here and link to https://devguide.python.org/#status-of-python-branches

nschloe:
Very nice PR. I hadn't expected it to be so unintrusive. Good job!

You still need to run `make format` to make the linter happy.

meshio/_helpers.py
```diff
@@ -242,23 +250,18 @@ def write(filename, mesh, file_format=None, **kwargs):
...
-                raise WriteError()
-        elif key in num_nodes_per_cell:
-            if value.shape[1] != num_nodes_per_cell[key]:
-                raise WriteError()
-        else:
-            # we allow custom keys <https://github.com/nschloe/meshio/issues/501> and
-            # cannot check those
-            pass
+    if is_buffer(filename, "r"):
+        assert (
```
nschloe: assert? :smile_cat:

clbarnes: Haha, started the changes before The Exceptioning, clearly didn't get all of them in the merge.

meshio/_files.py
```diff
@@ -0,0 +1,22 @@
...
+        with open(str(path_or_buf), mode) as f:
+            yield f
+    else:
+        with open(path_or_buf, mode) as f:
```
nschloe: Not `yield open(path_or_buf, mode)`? (Not familar with `@contextmanager`. I thought its purpose was that you can use the class with `with`.)

clbarnes: `@contextmanager` means that code before the `yield` is run on `__enter__`, then code after the `yield` is run on `__exit__`. Wrapping the `yield`s in `with` statements means that on `__exit__`, the file's `__exit__` is also (semi-explicitly) run, where if you just yield the result, then you're relying on whatever it is that closes file handles when they go out of scope, which I don't trust as much.

meshio/_abaqus.py
```diff
@@ -90,7 +91,7 @@
 def read(filename):
     """Reads a Abaqus inp file.
     """
-    with open(filename, "r") as f:
+    with open_file(filename, "r") as f:
         out = read_buffer(f)
```
nschloe: Do we still need the separation between `read` and `read_buffer` now? (Also flac3d etc.)

clbarnes: Some code somewhere might be relying on it existing, for cases where people were already reading out of HTTP responses and so on. It might also be nice to have an "accepts anything" function and then the actual reading logic in a more type-safe function, as it is now.

nschloe:
> Could use some tests and docs.

Tests, yes. Could you create a new file (`test/test_read.py` or so) where you read a file and a buffer of your favorite mesh format? Perhaps also a string etc, just for the purpose of demonstration. (And that I don't accidentally destroy the functionality later.)

For documentation, just add a word in the main readme in
```
mesh = meshio.read(filename)  # optionally specify file_format
```
Something like
```
mesh = meshio.read(
    filename,  # or PathLike or buffer
    # file_format="vtk"   # optionally specify the format
    ) 

```

meshio/_files.py
nschloe: Got it.

requirements.txt
```diff
@@ -0,0 +1,13 @@
+# runtime
```
nschloe: What's this file for?

clbarnes: The setup.py and requirements.txt do different things. The setup.py is package metadata used in the installation of the package: it contains hard runtime dependencies, and optional runtime dependencies under "extras", with loose-as-possible version constraints (if it's a library or has to coexist with other python packages). The requirements.txt is for developers and testers of the package: it contains all possible runtime dependencies, ideally with fixed versions to test against, and all test, doc, format, and build dependencies.

Without a requirements.txt, someone interested in contributing has to dig through the setup.py (and hope the extras are all well-described), all of the CI config, the Makefile, and potentially all of the test files to find what they actually need to install to make sure it works. With a requirements.txt, it's `git clone...; pip install -r requirements.txt` and they're away.

It doesn't have to stay, it's just something I would have found useful a couple of hours ago.

clbarnes: It can also help shrink CI configs as all of the dependencies can be listed in one place.

meshio/_abaqus.py
nschloe: Alright.

requirements.txt
nschloe: I see where you're coming from. It's just that now, if one adds or removes a dependency one has to do it in two files, and I can already see how this gets messed up. Let's keep it for now, but I might remove it in the future.

clbarnes: Agreed, it feels counterintuitive. I haven't had any issues with this pattern myself, but the possibility is there.

nschloe:
Alright, thank you!

PR #552, opened 2019-11-28
Format refactor

clbarnes:
This started off as a great idea and then spiralled in complexity.

## Problem

- Adding new formats is hard. Adding a new reader/writer meant modifying several different variables in `_helpers`; these dicts ended up dominating those files with hardcoded values.
- Third parties can't patch their own IO functions in at runtime
- If optional dependencies were missing, they would still be listed as available formats
- The functionality depended on there being exactly one function called `read` and one function called `write` in each module, without a type system to guarantee this; this constrained the use of additional arguments (see the necessity of `_writer_map`)

## Solution (sort of)

There are now a couple of dicts containing the name/ext mappings for readers and writers. Every module containing IO functions knows how to register itself in those readers and writers (there are helper functions making it very easy), in a manner dependent on the available dependencies. Third-parties can just call those registration helpers on their own functions and then use meshio's conversion functionality. Currying is used to pass extra arguments to read/write at registration time, in the same file as they are defined (easier to maintain).

## Remaining problems

- One import is made for its side-effects. I'm not happy with it but it is at least minimal, and it's got to happen somewhere.
- I'm not 100% certain that all the registrations end up at the same mappings we had before. I did my best but there were some inconsistencies to start with.


nschloe:
Thanks for the PR! Again, some great ideas here. The PR is quite intrusive though which is why I won't accept it. Large changes like this first need to be discussed with the other devs in an issue, otherwise people won't recognize the code anymore after one day.

I like the idea of the PR though. The fact that one has to edit multiple helper files to add a new format is a bit weird. Why don't you start by moving just the viable extensions into the mesh files? This seems like a small enough change that everyone can live with.

clbarnes:
That's fair, I started just trying to add a new (very simple) format and it got away from me :D

> Why don't you start by moving just the viable extensions into the mesh files?

Could you clarify this point?

nschloe:
> Could you clarify this point?

Each format file has a line
```
extensions = [".msh", ".lel"]
```
(or `filename_extensions` or so) at the top. In `_helpers.py` those are `import`ed and those arrays `input_filetypes`, `_extension_to_filetype` etc are constructed from it. (Perhaps we need more than `extensions`, I'm thinking of `name` etc.)

clbarnes:
Got it! Yes, that's a more incremental change. I think each would need:

- a dict of name to reader
- a dict of name to writer (currying rather than storing arts/kwargs saves on some overhead at the call site)
- a dict of extension to reader (or name)
- a dict of extension to writer (or name)

Then those can also exist in `helpers`, but be populated from the imported files. If we're going to enforce some structure on the modules (i.e. expecting variables to be there) it might be nice to implement them as classes, even if they're static classes so that there is a guarantee of that structure. But maybe I'm just getting carried away again.

What do you think about moving the format files into their own subpackage? I think it makes the layout more clear, and for refactors where there are repetitive tasks across all formats, it's nice to have them clearly delineated.

clbarnes:
If you haven't checked out the imageio project, I recommend it; that's where most of my inspiration is coming from!

clbarnes:
> it might be nice to implement them as classes

Halfway house: every module "exports" a NamedTuple containing all the registration information. That way all the functions stay as they are, there's only one module-level variable to enforce, but the other member variables have the structure imposed on them as well.

nschloe:
Closing as explained above. (Too intrusive for a single PR.)

PR #554, opened 2019-11-29
Neuroglancer format

clbarnes:
Before anything major changes :wink: 

https://github.com/google/neuroglancer/tree/master/src/neuroglancer/datasource/precomputed#legacy-single-resolution-mesh-format

One query: why do the tests check whether the arrays in a mesh are writeable (i.e. owned memory)? Meshio doesn't do any mesh mutation itself (besides pruning, which is a copy operation anyway). It's not a big deal, it would just save a copy in some circumstances to not have that requirement.


nschloe:
Thanks!

PR #555, opened 2019-11-29
WKT (TIN) format

clbarnes:
Well-Known Text representation of Triangulated Irregular Networks, useful for IO with spatial databases.


nschloe:
The format seems so simple; can't we just parse with `re`?

clbarnes:
Yes, probably. I sort of wanted to play with pyparsing, and also liked that it could read buffers directly rather than having to get the whole string out first.

nschloe:
> But that's probably not worth the dependency.

That'd be my first impulse as well. If you can do it without pyparsing, I'll merge it without hesitation.

gdmcbain:
I really like pyparsing.  It enables one to describe the format and then use that description to parse the file.  This is an elegant approach and leads to clear compact code.

However, I have found that it's too slow for reading meshes.  I'm routinely using meshio for meshes with 10⁶–10⁷ nodes.

At the moment I am using pyparsing to begin reading [polyMesh](https://cfd.direct/openfoam/user-guide/v7-mesh-description/#x24-1740005.1.2) meshes from OpenFOAM.  The parser runs slowly (on small meshes with 10³–10⁴ nodes) but it is quick to develop with pyparsing, which is helpful given the complexity of the polyMesh format and the early stage of my work.  I think that if I have to apply this to the aforementioned routine larger problems, I'll have to rewrite the parser using more of the idioms found in meshio.  (I'd eventually like to contribute this to meshio but it's awkward since polyMesh is intrinsically polygonal and polyhedral, not restricted to particular shapes of cells.)

Of course this slowness might well just be a limitation of how I'm designing my pyparsing parsers; I'd be interested to read of contrary experiences.

clbarnes:
> I really like pyparsing. It enables one to describe the format and then use that description to parse the file. This is an elegant approach and leads to clear compact code.
>
> However, I have found that it's too slow for reading meshes. I'm routinely using meshio for meshes with 10⁶–10⁷ nodes.

I suspect raw regex is also not going to be very efficient, either. But short of writing a custom "real" parser it may be the best option.

Any opinions on this design query: it would be much easier and probably a bit faster to write a parser on the basis of "if it is a valid TIN WKT, this will parse it correctly", with undefined behaviour for strings which are not valid WKT. It would be more complicated to write a parser which *only* works with valid WKT, and errors otherwise. Which implementation would you prefer to target?

gdmcbain:
Interesting question.  I don't know anything about the TIN WKT ecosystem or community, but for other formats in general, I think it often happens that mesh files are generated by different programs and don't necessarily adhere exact to the specification.  STL and MEDIT spring to mind.  Given that, I don't think it's meshio's job to be a mesh format compliance cop, better to act like web browsers that do their best to render as much of the HTML they're asked to, regardless of how well-formed it is.  Validation could be a useful extra though, if it's not too complicated, but it would want to be optional, or just in the manner of a warning, like 

https://github.com/nschloe/meshio/blob/590b5cdd2dd761192cb44f352e706c5050d38b84/meshio/_vtk.py#L618-L622

nschloe:
First off, I don't think efficiency should be our main concern here. The format is just badly designed and not suitable for fast computer consumption. If you want fast I/O, use some binary format like VTK, VTU, XDMF. When having to parse numbers, your code _will_ be slow, and I don't really care if 10 or 20 times slower than binary I/O.

As for how strict meshio can be, I think we should start of quite strict. That way, people take notice if they have a malformed file, and have the chance the fix it upstream. Sometimes we get issues with malformed files where the error isn't "too grave" (like a missing, but redundant, attribute in an XML file). In this case we can adapt the parser later.

clbarnes:
> The format is just badly designed and not suitable for fast computer consumption. If you want fast I/O, use some binary format like VTK, VTU, XDMF.

Agreed. There is a binary form (WKB) which is *also* very badly designed :man_shrugging: Also, for a format with "well-known" in the name it's actually nontrivial to get the spec.

> As for how strict meshio can be, I think we should start of quite strict. 

Sure. I'm working on a `re`-based implementation which makes sure the entire string is a valid TIN, and then pulls all the triangles out. Because of the way the indexing works you need to cache the whole structure a couple of times  anyway.

clbarnes:
One brand spanking new re-based parser. My regex-fu is not great; I'm sure groups could be used better for pulling out numbers and triangles rather than having to regex every element of the string several times. I was struggling a bit with matching groups in a way that they could be pulled out, but not capturing them (so that you can match 2 different floats with the same regex).

nschloe:
Could you reset setup.py and requirements.txt?

clbarnes:
My bad, I'll get that after lunch.

nschloe:
Thanks! If you want to improve upon the i/o, you can always submit another PR.

PR #558, opened 2019-12-03
Support CPS3

### LINKED ISSUE #557: AssertionError: Element type not available: CPS3
`meshio_convert cutout.inp cutout.dato` 
fails with the above mentioned error message.
[cutout.inp.txt](https://github.com/nschloe/meshio/files/3914911/cutout.inp.txt)

nilswagner:
Fixes #557.


nschloe:
CPS3 in Abaqus is a plane stress element, but I must say I don't really get what that means with respect to the geometry of the domain. Since it works for you and disturbs no other tests, I'll merge it.

PR #559, opened 2019-12-03
Support of CPS4

### LINKED ISSUE #560: Support of CPS4 element
meshio_convert fails with the attached Abaqus model.
[rect.inp.txt](https://github.com/nschloe/meshio/files/3917189/rect.inp.txt)
If you're having problems converting from or to a mesh format, remember to attach a mesh that shows the problem. This will enable other people to reproduce and fix the problem.

nilswagner:
Fixes #560.

### DIFF
--- a/meshio/_abaqus.py
+++ b/meshio/_abaqus.py
@@ -30,6 +30,7 @@ abaqus_to_meshio_type = {
     "B33": "line3",
     "B33H": "line3",
     # surfaces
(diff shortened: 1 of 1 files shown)


nschloe:
Thanks.

By the way you don't need to open up a bug report _and_ and PR for such simple things. Just use the text field in the PR to describe what you mean.

PR #564, opened 2019-12-04
only read XDMF Polyline of 2 points

### LINKED ISSUE #563: read line XDMF
In the example for #561, notice the output line for the reading of XDMF (when the file was named by a `str`):
```
xdmf str {'line': array([[2, 0, 1]], dtype=uint64)}
```
The `dict` is the `.cells` attribute.  It should have `.cells['line'].shape[1] == 2`.  

I think the 2 has been prepended as the number of points on the XDMF POLYLINE and that this should be checked and stripped on reading.

gdmcbain:
fixes #563


gdmcbain:
There are no `cells`of type `'line'` in [test/helpers.py](test/helpers.py).  Should some be added so that this path is covered?

nschloe:
Yes, add it!

gdmcbain:
Oh, it's passing the checks now…

I haven't added the line_mesh yet, that's in #569 which is yet to be merged.

PR #565, opened 2019-12-04
Black-compatible isort configuration

clbarnes:
Currently, `make format` causes `isort` to reformat imports, and then `black` to undo those changes. This configuration should prevent isort making black-incompatible changes.

See "a compatible \`.isort.cfg\`" on the Black website https://black.readthedocs.io/en/stable/the_black_code_style.html

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,9 +6,10 @@ jobs:
       - image: circleci/python:3.7
     steps:
       - checkout
-      - run: pip3 install -U black flake8 --user
+      - run: pip3 install -U black flake8 isort --user
       - run: black --check .
       - run: flake8 .
+      - run: isort --check-only -rc .
   build:
     working_directory: ~/work
     docker:
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,4 @@
 [flake8]
-ignore = E203, E266, W503
-max-line-length = 88
+ignore = E203, E266, E501, W503
 max-complexity = 18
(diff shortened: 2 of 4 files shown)


nschloe:
Can you add a link to where this comes from into the file? (Not sure if comments are supported in this format.)

Also, with this change we can isort-check as part of our tests in .github and .circleci. Can you add it there?

tianyikillua:
According  to https://github.com/psf/black, it would be `line_length=88`

clbarnes:
> Can you add a link to where this comes from into the file? (Not sure if comments are supported in this format.)

It's INI, there's no spec so it's basically thunderdome. I believe most implementations have some kind of comments though. And yes, will do.

> According to https://github.com/psf/black, it would be line_length=88

Yes, but meshio's flake8 config gives the max line length at 80, so I went with that. However, looking at it, line length errors are ignored in the flake8 config. If we're going with black (88) and having flake8 ignore line length (E501), flake8 shouldn't define a max-line-length.

nschloe:
>  flake8 shouldn't define a max-line-length.

Yep, remove it please.

clbarnes:
The import sort checks are failing on CI but work locally. Any thoughts? Maybe the fact that isort is user-installed means it's missing the config file?

nschloe:
Hm, weird. Perhaps you can SSH into the circleci box and see what's wrong there.

nschloe:
Closing in favor of #570.

PR #566, opened 2019-12-04
Move formats into subpackage

clbarnes:
Also, exclude tests from setup.py install.

I feel that this makes the codebase easier to navigate, especially as more formats are added. Some of the imports have been switched from relative to absolute; I can change that as required.


nschloe:
Thanks Chris! Got to say that I don't feel like merging this right now. Navigating the code base simply doesn't cause enough pain for me right now to move all the files around. It may be the case in the future though, so I'll definitely keep that idea in mind.

As for the installation of test files: What's the problem here? (I don't think they actually get installed, do they?)

clbarnes:
> As for the installation of test files: What's the problem here? (I don't think they actually get installed, do they?)

Seems you're right - I thought the zero-argument `find_packages` would pick up the test directory but I guess it doesn't recognise it as a module because it doesn't have an `__init__.py`.

clbarnes:
> Navigating the code base simply doesn't cause enough pain for me right now to move all the files around.

As someone who came into the codebase more recently, I feel it would have been helpful at the point of entry to distinguish "structure" files from the format implementations - less to search through when picking it up. Your show though!

nschloe:
I've improved a bunch of things in #572, including the format and extension registration, so let's close this one here.

PR #569, opened 2019-12-05
simple line_mesh for the test suite

gdmcbain:
PR #564 ‘only read XDMF Polyline of 2 points’ failed a coverage test.  There aren't any meshes with cells of type `'line'` in `test.helpers`, so here's one.


nschloe:
Fantastic, thanks!

Issue #553, opened 2019-11-29
read mixed XDMF


gdmcbain:
It's possible to `meshio.write` an XDMF file with more than one type of `cell` and the file appears to open correctly in ParaView but `meshio.read` raises a `KeyError`.

For example, the m.w.e. from #223, minimally updated, and with a `read` appended:
```python
from numpy import array
from meshio import Mesh, write, read

points = array([[0.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]])
cells = {"line": array([[0, 1], [1, 2], [2, 0]]), "triangle": array([[0, 1, 2]])}
write("foo.xdmf", Mesh(points, cells))
bar = read("foo.xdmf")
```
raises
```
…
  File "/home/geordie/.local/share/miniconda3/envs/scikit-fem/lib/python3.7/site-packages/meshio/_xdmf/common.py", line 118, in translate_mixed_cells
    r += xdmf_idx_to_num_nodes[data[r]] + 1
KeyError: 2
```
And similarly with the example [mwe1.xdmf.gz](https://github.com/nschloe/meshio/files/3903664/mwe1.xdmf.gz) from #539.

gdmcbain:
Actually the problem isn't with `TopologyType="mixed"`, since
```python

**Comment by gdmcbain** [comment]
Since #564 #569, `read` in the top (mixed line–triangle) example is still failing but with a different `KeyError`:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/geordie/src/meshio/meshio/_helpers.py", line 234, in read
    return format_to_reader[file_format].read(filename)
  File "/home/geordie/src/meshio/meshio/_xdmf/main.py", line 27, in read
    return XdmfReader(filename).read()
  File "/home/geordie/src/meshio/meshio/_xdmf/main.py", line 53, in read
    return self.read_xdmf3(root)
  File "/home/geordie/src/meshio/meshio/_xdmf/main.py", line 239, in read_xdmf3
    cells = translate_mixed_cells(data)
  File "/home/geordie/src/meshio/meshio/_xdmf/common.py", line 119, in translate_mixed_cells
    r += xdmf_idx_to_num_nodes[data[r]] + 1
KeyError: 0
```

gdmcbain:
Yes, we just need to increment `r` one more to skip the number of points in the XDMF Polyline.

Issue #568, opened 2019-12-05
multiple XDMF Grids


gdmcbain:
I was able to propose a solution to the question [‘multiple topology blocks?’ (xdmf#15)](https://gitlab.kitware.com/xdmf/xdmf/issues/15) by using multiple `Grid` elements.  The XDMF ([multitop.xdmf.txt](https://github.com/nschloe/meshio/files/3924729/multitop.xdmf.txt)) posted there is accepted by ParaView 5.6.0 but not `meshio.read`.

```
meshio._exceptions.ReadError: XDMF reader: Only supports one grid right now.
```

gdmcbain:
Although the example referred to above (with the `Xdmf Version` manually bumped from `"2.0"` to `"3.0"`) 
```xml
<?xml version="1.0" ?>
<!DOCTYPE Xdmf SYSTEM "Xdmf.dtd" []>
<Xdmf Version="3.0" xmlns:xi="http://www.w3.org/2001/XInclude">
    <Domain>
        <DataItem Name="Point Data" Format="XML" Dimensions="6 3">
            0.0     1.0    0.0
	    1.0	    0.0	   0.0
	    2.0	    0.0	   0.0
	    2.0	    2.0	   0.0
	    1.0	    2.0	   0.0
	    0.0	    3.0	   0.0
	</DataItem>
        <Grid Name="One Quadrilateral">
            <Topology Type="Quadrilateral" NumberOfElements="1">
                <DataItem Format="XML"
                          DataType="Int"
                          Dimensions="1 4">
                    1 2 3 4
                </DataItem>
            </Topology>
            <Geometry GeometryType="XYZ">
                <DataItem Reference="XML">
                    /Xdmf/Domain/DataItem[@Name="Point Data"]
                </DataItem>
            </Geometry>
        </Grid>
        <Grid Name="Two Triangles">
            <Topology Type="Triangle" NumberOfElements="2">
                <DataItem Format="XML"
                          DataType="Int"
                          Dimensions="2 3">
                    0 1 4
		    0 4 5
                </DataItem>
            </Topology>
            <Geometry GeometryType="XYZ">
                <DataItem Reference="XML">
                    /Xdmf/Domain/DataItem[@Name="Point Data"]
                </DataItem>
            </Geometry>
        </Grid>
    </Domain>
</Xdmf>
```
[...]

PR #571, opened 2019-12-05
skip over the number 2 of points in an XDMF polyline in mixed topology

gdmcbain:
As discussed in #553.


nschloe:
Do we want any tests for this?

gdmcbain:
Yes of course.  Just a sec.

(Actually I imagine one might want to avoid writing `TypologyType="mixed"` once having more than one `Grid` is possible #568, but they are valid XDMF and so might be found in the field.)

nschloe:
Yeah, it's a nice idea to have multiple grids. The devil invented mixed topologies.

Issue #583, opened 2019-12-10
f-string in WKT


keurfonluu:
The WKT format uses an f-string

```python
tin_pattern = rf"TIN\s*\((?:\s*{triangle_pattern}\s*,?)*\s*\)"
```

that has only been introduced in Python 3.6.

I don't know if it's necessary, but I guess the ``f`` could be removed to make it compatible with Python 3.4 and 3.5.

nschloe:
Yep, I guess.

clbarnes:
Not exactly necessary, just deeply preferred. Python 3.4 is EOL anyway, and it might be nice to change `python_requires` to reflect that.

There are shims elsewhere in the codebase for supporting 3.5, so I'll fix this one, registering my protests in so doing :stuck_out_tongue:

clbarnes:
See #585

PR #585, opened 2019-12-10
Python 3.5 compatibility

clbarnes:
Removes a couple of f-strings which had snuck in (because they're great).

Also updates package metadata to only support python >= 3.5 (because 3.4 is EOL).


clbarnes:
Resolves #583

nschloe:
Thanks!

Issue #584, opened 2019-12-10
improve memory efficiency of flac3d


nschloe:
The current FLAC3D is the worst of all when it comes to memory efficiency. https://github.com/nschloe/meshio#maximum-memory-usage

@keurfonluu You might want to look into this.

keurfonluu:
Don't know what else to do to improve the performances after PR #588.

By the way, aren't the I/O speed figures affected by ``tracemalloc``?

nschloe:
> By the way, aren't the I/O speed figures affected by tracemalloc?

Not really.

PR #588, opened 2019-12-11
Improve FLAC3D read and write performances (Issue #584)

keurfonluu:
- Changed: refactor reader code to rely less on dictionaries,
- Changed: zone groups are no longer written if not recognized or ``cell_data`` is empty,
- Added: handle non-naturally ordered cells.

**Remarks**
1. I still need to use some dictionaries to handle non-natural order of points and cells since the format does not provide how many points and cells are in the mesh,
2. The conversion of  ``cells`` to ``numpy.ndarray`` in the reader uses some memory. I have no idea how to append directly to a ``numpy`` array in place.

**Results**
On an example of mine (100K points and 570K tetra), the old reader used 300 MB down to 200 MB with the new reader. The new reader is also slightly faster compared to the old one.

### DIFF
--- a/meshio/flac3d/_flac3d.py
+++ b/meshio/flac3d/_flac3d.py
@@ -17,13 +17,14 @@ meshio_only = {"tetra", "pyramid", "wedge", "hexahedron"}
 meshio_data = {"flac3d:zone", "gmsh:physical", "medit:ref"}
 
 
-flac3d_to_meshio_type = {
-    "T4": "tetra",
-    "P5": "pyramid",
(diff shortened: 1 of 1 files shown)


keurfonluu:
Don't know why the last commit didn't pass the build test....

nschloe:
> ERROR: /home/circleci/project/meshio/flac3d/_flac3d.py Imports are incorrectly sorted.

`make format`

nschloe:
Thanks! I've updated the images in the readme.

PR #589, opened 2019-12-11
Multi-version testing configuration

clbarnes:
Run tests against all supported versions, as well as linting, with a single command.

Things I'm unsure of:

1. Explicitly installing the dependencies in the `requirements.txt` in each environment. This reduces the number of times we have to specify our development dependencies, but possibly is not a fair test of the normal installation procedure.
2. The circleci configuration got a lot simpler, but we lose parallelism between the lint and build steps. How best to test multiple python versions in parallel is not documented (issue raised here https://github.com/circleci/circleci-docs/issues/4005 ).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,7 @@ shared: &shared
     - checkout
     - run: git lfs pull
     - run: pip install .[all] --user
+    - run: pytest --cov
     - run: |
         if [ $UPLOAD_COV ]; then
           bash <(curl -s https://codecov.io/bash)
@@ -37,9 +38,8 @@ jobs:
       - image: circleci/python:3.5
 
   lint:
-    <<: *shared
     docker:
-      - image: circleci/python:3.7
+      - image: circleci/python:3.8
     steps:
       - checkout
       - run: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     docker:
       - image: circleci/python:3.8
     steps:
-      - run: pip install tox tox-pyenv
+      - run: pip install tox tox-pyenv --user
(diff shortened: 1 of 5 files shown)


nschloe:
Wait, what does this do? Why do we need a local config? Can't we simply do all these things on circleci/gh-actions?

clbarnes:
Tox means that you can run the tests and lints against all supported python versions, in exactly the same way CI does, on your local machine. It saves lots of pushing, waiting, and history rewriting. We can do those things on CI, but being able to do it locally in an ergonomic fashion can also be helpful.

nschloe:
I think the benefit of having it run locally is outweighed by the disruption this PR causes in the config. (That's another way of saying I don't understand what's going on. :smile: tox, tox-pyenv, pyenv global, envlist... phew.)

Compatibility issues with older Python version have never been a big problem for meshio, and in case someone _did_ notice something (happens about once a year), the fix is trivial.

I'm inclined to close this.

clbarnes:
OK, I can scrub tox from the CI config and just use CircleCI's docker images to cover the different python versions. Shall I remove the tox config file entirely or leave it there for the benefit of developers?

clbarnes:
P.S. I'll squash down all of these commits... testing CI configs is always a pain.

clbarnes:
I think this will require you to change the repo's github actions configuration, as there is no longer a job called "build".

The failure on py35 is an actual test failure due to getting the wrong result which is fine on other versions. That's an implementation problem, and a good argument for testing against all python versions :wink:

nschloe:
So, I just added multiple python version to gh-actions with about 5 lines of code. https://github.com/nschloe/meshio/pull/590

No tox necessary.

clbarnes:
OK, so does that mean that `.circleci/` is another vestigial CI configuration which needs to be removed?

Issue #579, opened 2019-12-10
OBJ to PLY >> ValueError: invalid literal for int() with base 10: '1/1/1'


loicroybon:
Hello, 
I'm trying to convert an OBJ mesh (georeferenced, so with high coordinates values) to PLY output format.
I've this issue : 
```
Traceback (most recent call last):
  File "/home/loicroybon/bin/conversion_scripts/meshio_converter.py", line 6, in <module>
    mesh = meshio.read(input_file)
  File "/usr/local/lib/python3.7/dist-packages/meshio/_helpers.py", line 67, in read
    return reader_map[file_format](filename)
  File "/usr/local/lib/python3.7/dist-packages/meshio/obj/_obj.py", line 18, in read
    mesh = read_buffer(f)
  File "/usr/local/lib/python3.7/dist-packages/meshio/obj/_obj.py", line 49, in read_buffer
    faces.append([int(item.split("//")[0]) for item in split[1:]])
  File "/usr/local/lib/python3.7/dist-packages/meshio/obj/_obj.py", line 49, in <listcomp>
    faces.append([int(item.split("//")[0]) for item in split[1:]])
**ValueError: invalid literal for int() with base 10: '1/1/1'**
```
Is there a workaround / solution ?

loicroybon:
To complete, those are the first lines of my input OBJ file
`# File produced by Open Asset Import Library (http://www.assimp.sf.net)
# (assimp v4.0.0)

mtllib Mesh.mtl

# 305007 vertex positions
v  555813.1512337965 5385888.544480518 -185.8601684570313
v  555813.1708465158 5385888.744605773 -185.9606170654297
v  555813.0237576538 5385888.617726117 -185.9500885009766
v  555795.3996238171 5385883.850832564 -182.0402526855469
v  555795.4913905058 5385883.833155329 -182.123046875
v  555795.4713974552 5385883.993770655 -182.1911010742188
v  555821.1034882489 5385867.785865381 -162.9013366699219
v  555821.0701044361 5385867.60030769 -162.8728485107422
`

nschloe:
Please you triple fences for code multiline highlighting.

loicroybon:
> Please you triple fences for code multiline highlighting. (You can edit your posts.)

Done

nschloe:
Closing via #591.

PR #591, opened 2019-12-11
Fix issue #579

keurfonluu:
- Changed: handle case for vertices with textures and normals (e.g. ``1/1/1``).

Fixes #591.


nschloe:
Thanks!

PR #594, opened 2019-12-13
Improve FLAC3D writer

keurfonluu:
- Fixed: wrong cell ID after commit 0e06b48c68bcc82c5cbd1db20de86f65068a9870,
- Fixed: use double quotes when writing zone group names,
- Changed: use [``slicing_summing``](https://stackoverflow.com/questions/42158228/fast-scalar-triple-product-in-numpy/42386330#42386330) to calculate scalar triple product (a bit faster but lower memory usage),
- Changed: preserve field data original order,
- Added: partial support for higher order cells (converted to lower order cells).


nschloe:
This pumped the FLAC3D mem requirements back up. I'll try to bring it down again (while keeping the tests clean of course). If I accidentally do something wrong then, we'll need to add more tests.

Issue #600, opened 2019-12-19
read mixed XDMF 2


gdmcbain:
In #599, distinguishing between how Polylines are written in XDMF depending on whether they're in `TopologyType` `Polyline` or `Mixed`, a simple test XDMF file generated by ParaView could not be read by meshio.

To reproduce: first generate the VTK 
```python
from numpy import array
from meshio import Mesh, write

mesh =  Mesh(array([[0., 0., 0.],
                    [1., 0., 0.],
                    [1., 1., 0.],
                    [0., 1., 0]]),
             {'line': array([[0, 1],
                             [2, 3],
                             [3, 0]]),
              'quad': array([[0, 1, 2, 3]])})
write('mixed.vtk', mesh)
```
then open and `Save Data` in ParaView (here 5.6.0).  This gives:
```xml
<?xml version="1.0" ?>
<!DOCTYPE Xdmf SYSTEM "Xdmf.dtd" []>
<Xdmf xmlns:xi="http://www.w3.org/2003/XInclude" Version="2.2">
  <Domain>
    <Grid GridType="Uniform">
      <Topology TopologyType="Mixed" Dimensions="4">
        <DataItem Dimensions="17" NumberType="Int" Precision="8" Format="XML">
2 2 0 1 2 2 2 3 2 2 
3 0 3 0 1 2 3 
</DataItem>
      </Topology>
      <Geometry GeometryType="XYZ">
        <DataItem Dimensions="4 3" NumberType="Float" Precision="8" Format="XML">
0 0 0 
1 0 0 
1 1 0 
0 1 0 
</DataItem>
      </Geometry>
    </Grid>
  </Domain>
</Xdmf>
```
Attempting to read this XDMF in meshio raises a `KeyError`.
```
 File "/home/geordie/.local/share/miniconda3/envs/scikit-fem/lib/python3.7/site-packages/meshio/_helpers.py", line 67, in read
    return reader_map[file_format](filename)
  File "/home/geordie/.local/share/miniconda3/envs/scikit-fem/lib/python3.7/site-packages/meshio/xdmf/main.py", line 27, in read
    return XdmfReader(filename).read()
  File "/home/geordie/.local/share/miniconda3/envs/scikit-fem/lib/python3.7/site-packages/meshio/xdmf/main.py", line 48, in read
    return self.read_xdmf2(root)
[...]

PR #601, opened 2019-12-19
read TopologyType Mixed  in XDMF2

gdmcbain:
This is for #600.  Not extensively tested, since meshio and ParaView 5.7.0 only write XDMF 3.0, so I'm relying on ParaView 5.6.0 and that's buggy, but this is better than what's in master.

### DIFF
--- a/meshio/xdmf/main.py
+++ b/meshio/xdmf/main.py
@@ -147,8 +147,12 @@ class XdmfReader:
                 data_items = list(c)
                 if len(data_items) != 1:
                     raise ReadError()
-                meshio_type = xdmf_to_meshio_type[c.attrib["TopologyType"]]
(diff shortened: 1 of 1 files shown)


gdmcbain:
> 0% of diff hit 

Yep; meshio doesn't write XDMF 2.

https://github.com/nschloe/meshio/blob/1f5e725decbabbf7de1c9f03daaa8c36e7a63331/meshio/xdmf/main.py#L335

nschloe:
> Yep; meshio doesn't write XDMF 2.

There are some mesh formats that I'd love to get rid of, like XDMF2 which nobody uses, but the reader should still stay put in case someone runs into an old file someday. It's okay that it's not too well tested.

PR #602, opened 2019-12-19
only prepend 2 2 to Polyline in XDMF Topology Type Mixed

### LINKED ISSUE #599: prepending number of vertices in XDMF TopologyType Polyline
In #539, @dajuno wrote:

> Prepending 2 to lines breaks compatibility with the FEniCS MeshValueCollection reader for reading physical line labels from XDMF (https://fenicsproject.discourse.group/t/transitioning-from-mesh-xml-to-mesh-xdmf-from-dolfin-convert-to-meshio/412/54).
> I don't know about other applications/XDMF readers, but is your recommendation to follow your decision here (i.e., to raise the issue with FEniCS devs)?

(The problem isn't related to mutation, so I've launched this separate new issue.   It's slightly more related to #553, though the XDMF files in question aren't of `TopologyType="Mixed"`.)  

The snippet at https://fenicsproject.discourse.group/t/transitioning-from-mesh-xml-to-mesh-xdmf-from-dolfin-convert-to-meshio/412/53  :

```xml
<Topology TopologyType="Polyline" NumberOfElements="40">
   <DataItem DataType="Int" Dimensions="40 3" Format="XML" Precision="8">
       2 0 4
       2 4 5
       2 5 6
       ...
```

@dajuno, could you provide a minimal working example?  Here's a simple one based on that for #539: [539.tar.gz](https://github.com/nschloe/meshio/files/3980907/539.tar.gz).  It's like the snippet but with one Polyline of two vertices, with that number two being prepended.  It reads fine in ParaView 5.6.0 and VisIt 3.0.2.

If the polylines aren't prepended with the number, how is the reader to infer it or know where the polylines break?

_Reference:—_ http://www.xdmf.org/index.php/XDMF_Model_and_Format#Linear

gdmcbain:
Fixes #599.

### DIFF
--- a/meshio/xdmf/main.py
+++ b/meshio/xdmf/main.py
@@ -428,7 +428,11 @@ class XdmfWriter:
                 NumberOfElements=str(total_num_cells),
             )
             total_num_cell_items = sum(numpy.prod(c.shape) for c in cells.values())
-            dim = str(total_num_cell_items + total_num_cells)
+            dim = str(
+                total_num_cell_items
+                + total_num_cells
+                + (cells["line"].shape[0] if "line" in cells else 0)
+            )
             cd = numpy.concatenate(
                 [
                     # prepend column with xdmf type index
--- a/meshio/xdmf/main.py
+++ b/meshio/xdmf/main.py
@@ -427,7 +427,7 @@ class XdmfWriter:
                 TopologyType="Mixed",
                 NumberOfElements=str(total_num_cells),
             )
-            total_num_cell_items = sum(c.size for c in cells.values())
+            total_num_cell_items = sum(numpy.prod(c.shape) for c in cells.values())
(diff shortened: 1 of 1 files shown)


gdmcbain:
Oops, need to make the corresponding fix in `XdmfReader`.

nschloe:
Thanks Geordie for the fix and the extensive research!

gdmcbain:
Of course I do hope eventually to make this redundant by having meshio not write `Topology Type="Mixed"` (drive out that [‘devil’](https://github.com/nschloe/meshio/pull/571#issuecomment-562356425)!) in case

https://github.com/nschloe/meshio/blob/57f32c77b1c08c43f07987e3accc84632b32bc5a/meshio/xdmf/main.py#L432

but instead write multiple XDMF Grids #568 ; however here I felt that the impact on FEniCS #599 required prompt action.

PR #610, opened 2020-01-02
WIP: read multiple Grids in XDMF

gdmcbain:
This is for #568.

### DIFF
--- a/meshio/xdmf/main.py
+++ b/meshio/xdmf/main.py
@@ -40,7 +40,7 @@ class XdmfReader:
         if root.tag != "Xdmf":
             raise ReadError()
 
-        version = root.attrib["Version"]
+        version = root.get("Version")
 
         if version.split(".")[0] == "2":
             return self.read_xdmf2(root)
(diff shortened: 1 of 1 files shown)


gdmcbain:
This includes #609.

gdmcbain:
Thus far, this does read multiple Grids well enough to read the MWE; but perhaps this should go further and:
* Distinguish the cells read from different grids.
* Write multiple grids instead of `TopologyType="Mixed"`.

The exact desired behaviour does depend on bit on the interpretation of tagged subsets of elements in meshio #290.

gdmcbain:
Given the current code

https://github.com/nschloe/meshio/blob/23fa4ef8862f8f45b4dfdf4da5f5cebc05e92b49/meshio/xdmf/main.py#L251

plus the new code in #610 to read multiple Grids, subsequent grids with the same `cell_type` will clobber those preceding them.  That can't be the desired behaviour.  I suppose the new cells should be appended but should they be distinguished from those already read?  How?  Like Gmsh MSH2
```python
cell_data[xdmf_to_meshio_type[cell_type]]["gmsh:physical"]
```
or something like MED for which I see 

https://github.com/nschloe/meshio/blob/23fa4ef8862f8f45b4dfdf4da5f5cebc05e92b49/meshio/med/_med.py#L105-L114

though I'm not familiar with the MED format so don't know what that attribute would usually be expected to contain.

…Ah, I've just noticed that `meshio.Mesh` has sprouted some new attributes! `node_sets` and `element_sets`.  I assume that these should be renamed `point_sets` and `cell_sets` (as per #493).  The `cell_sets` could be just what's wanted here (and for #290).

gdmcbain:
This now works again, following the switch from `lxml`, correctly reading the MWE of #568 with its two Grids sharing Geometry.  It also records the two Topologies to `cells` and `cell_sets` (the latter partially fulfilling #290).

It remains to also write multiple grids, thereby:
* avoiding `TopologyType="Mixed"`
* enabling testing via `test.helpers.write_read`

nschloe:
The smaller a PR, the better; we can do multiple grids later.

gdmcbain:
Yes, right, sorry; I'll split this up.

gdmcbain:
1. #621

gdmcbain:
1. #621 
2. #622

PR #618, opened 2020-01-05
Add methods to iterate over points and cells

keurfonluu:
- Added: ``Mesh.iterpoints`` to iterate over points, it returns a generator that returns a tuple with the point coordinates and a dict with the point data associated with the current point,
- Added: ``Mesh.itercells`` to iterate over cells, it returns a generator that returns a tuple with the cell corner points, a dict the cell data associated with the current cell, and the current cell type.

**Usage**

```python
for point in mesh.iterpoints():
    # Do something with point coordinates and/or data

for cell in mesh.itercells():
    # Do something with cell corner points, data and/or type
```

**Comment**
- Sometimes, we just want to iterate over point/cell and their data at the same time and it usually ends up with complicated nested for loops (especially with cells). These methods provide a simple way to iterate over points and cells with a single for loop,
- In ``python < 3.7``, dicts are iterated based on the hash value of their keys while in ``python >= 3.7``, dicts are iterated in the same order as they were defined. This results in inconsistent order of iteration when iterating over cells, depending on the Python environment. In ``Mesh.itercells``, the cells are iterated over depending on the number of nodes providing consistency across the different versions of Python.

### DIFF
--- a/meshio/_mesh.py
+++ b/meshio/_mesh.py
@@ -113,3 +113,27 @@ class Mesh:
         from ._helpers import read
 
         return read(path_or_buf, file_format)
+
+    def iterpoints(self):
+        for i, point in enumerate(self.points):
+            data = (
+                {label: data_array[i] for label, data_array in self.point_data.items()}
+                if self.point_data
+                else {}
+            )
+            yield (point, data)
+
+    def itercells(self):
+        from ._common import num_nodes_per_cell
+
+        for cell_type in sorted(self.cells.keys(), key=lambda x: num_nodes_per_cell[x]):
+            for i, corner in enumerate(self.cells[cell_type]):
+                data = (
+                    {
+                        label: data_array[i]
+                        for label, data_array in self.cell_data[cell_type].items()
+                    }
+                    if cell_type in self.cell_data
+                    else {}
+                )
+                yield (corner, data, cell_type)


nschloe:
Uuh I'm not sure about that. I'd always thought that you never want to explicitly loop over the points/cells because the numbers are just so big and Python loops are slow. Block-operations are typically what you want.

keurfonluu:
Yeah, I know and I completely agree with you. But similarly to ``pandas.itertuples`` and ``pandas.iterrows``, it should only be used as a last resort when we can't use vectorization (e.g. in the FLAC3D format writer, but I guess there should be other applications we didn't think of).

For instance (useless example, I don't know why someone would do that):

```python
import meshio
import pandas
mesh = meshio.read("mesh.msh")
df = pandas.DataFrame(columns = ["Corner", "Data", "Type"])
for i, cell in enumerate(mesh.itercells()):
    df.loc[i] = cell
```

It is not meant to replace the more efficient way to iterate over cells (i.e. ``mesh.cells.items()``), but rather provide an additional option to iterate over cells that would provide all the information we have for a given cell with a single ``for`` loop.

nschloe:
I thought about it some more and I'm still not convinced we need this. The biggest issue I have with it is that the loops are super slow. If the user wants to write slow code, she can always do that on her own. meshio should not encourage that by making it easy.

keurfonluu:
Fair enough. Feel free to close this PR.

keurfonluu:
I have been playing around with the new ``cells`` and ``cell_data``. This is no longer an issue and thus not needed anymore.

nschloe:
Never knew this would be related. Cool.

Issue #620, opened 2020-01-05
minor vestiges from lxml


gdmcbain:
There are a few minor changes needed following the switch back from `lxml` to the standard library #613; e.g.
* Instead of the `.xpath` method, there's [`.find`](https://docs.python.org/3/library/xml.etree.elementtree.html?highlight=xpath#xml.etree.ElementTree.Element.find).  This affects #608 and its fix, #609.
* One get [`.get`](https://docs.python.org/3/library/xml.etree.elementtree.html?highlight=xpath#xml.etree.ElementTree.Element.get) XML attributes.

The fix for this will be the first part of splitting up the cumbersome proposal #610 to address multiple XDMF grids #568.

gdmcbain:
An example, based on #608: 
[ref.xdmf.txt](https://github.com/nschloe/meshio/files/4024251/ref.xdmf.txt)
```python
import xml.etree.ElementTree as ET

from meshio.xdmf.main import XdmfReader

reader = XdmfReader('ref.xdmf')
xdmf = ET.parse(reader.filename, ET.XMLParser()).getroot()
data_item = xdmf.find('./Domain/Grid/Geometry/DataItem')
print(data_item.attrib, data_item.text)
print(reader._read_data_item(data_item))
```
The last line raises
```
  File "/home/geordie/src/meshio/meshio/xdmf/main.py", line 58, in _read_data_item
    data_item.xpath(
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'xpath'
```

PR #621, opened 2020-01-06
xml.etree.ElementTree usage

gdmcbain:
This is for #620, being the first part of the splitting up of #610.

### DIFF
--- a/meshio/xdmf/main.py
+++ b/meshio/xdmf/main.py
@@ -56,11 +56,7 @@ class XdmfReader:
         reference = data_item.get("Reference")
         if reference:
             return self._read_data_item(
-                data_item.xpath(
-                    data_item.text
-                    if reference == "XML"
-                    else reference
-                )[0]
+                data_item.xpath(data_item.text if reference == "XML" else reference)[0]
             )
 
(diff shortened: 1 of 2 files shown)


gdmcbain:
With this, the example from #620 gives:
```
{'Reference': 'XML'} 
                   /Xdmf/Domain/DataItem[@Name="Point Data"]
               {'Reference': 'XML'} 
                   /Xdmf/Domain/DataItem[@Name="Point Data"]
               
[[[0. 0. 0.]
  [1. 0. 0.]
  [1. 1. 0.]
  [0. 1. 0.]]

 [[0. 0. 2.]
  [1. 0. 2.]
  [1. 1. 2.]
  [0. 1. 2.]]]

[[[0. 0. 0.]
  [1. 0. 0.]
  [1. 1. 0.]
  [0. 1. 0.]]

 [[0. 0. 2.]
  [1. 0. 2.]
  [1. 1. 2.]
  [0. 1. 2.]]]
```
which is correct.

It's still not possible to `meshio.read('ref.xdmf')` because `meshio` mistakes the `DataItem Name="Point Data"` for a `Grid`; handling that requires the full #610 of which this is the first part.

PR #631, opened 2020-01-08
first stab at cells dict -> list of (named) tuples

### LINKED ISSUE #629: cells dict -> list of tuple/namedtuple/dataclasses?
Another idea for 4.0:

We could transform the cells (and cell data) from a dict to a list of tuples, namedtuples, or dataclasses ([as has been suggested before](https://github.com/nschloe/meshio/issues/615#issuecomment-570833763)), having the attributes `type` (e.g., `triangle6`) and `data` (a numpy array). With this, we could preserve cell ordering (@MuellerSeb is going to like that) and better represent data structures like Exodus's regions.

Opinions, please. @tianyikillua

nschloe:
Fixes #629.

### DIFF
--- a/meshio/_mesh.py
+++ b/meshio/_mesh.py
@@ -119,7 +119,7 @@ class Mesh:
         assert len(self.cells) == len(
             numpy.unique([c.type for c in self.cells])
         ), "More than one block of the same type. Cannot create dictionary."
-        return {cell_type: data for cell_type, data in self.cells}
+        return dict(self.cells)
 
     @classmethod
     def read(cls, path_or_buf, file_format=None):
--- a/meshio/_mesh.py
+++ b/meshio/_mesh.py
@@ -52,32 +52,28 @@ class Mesh:
         return "\n".join(lines)
 
     def prune(self):
-        prune_list = []
+        prune_list = ["vertex", "line", "line3"]
+        if any([c.type in ["tetra", "tetra10"] for c in self.cells]):
+            prune_list += ["triangle", "triangle6"]
 
-        for cell_type in ["vertex", "line", "line3"]:
-            if cell_type in self.cells:
-                prune_list.append(cell_type)
+        new_cells = []
+        new_cell_data = {}
+        for c in self.cells:
+            if c.type not in prune_list:
+                new_cells.append(c)
+                for name, data in self.cell_data:
+                    if name not in new_cell_data:
+                        new_cell_data[name] = []
+                    new_cell_data[name].append(data)
 
-        self.cells.pop("vertex", None)
-        self.cells.pop("line", None)
-        self.cells.pop("line3", None)
-        if "tetra" in self.cells or "tetra10" in self.cells:
-            # remove_lower_order_cells
-            for cell_type in ["triangle", "triangle6"]:
-                if cell_type in self.cells:
-                    prune_list.append(cell_type)
-
-        for cell_type in prune_list:
-            self.cells.pop(cell_type, None)
-            self.cell_data.pop(cell_type, None)
+        self.cells = new_cells
+        self.cell_data = new_cell_data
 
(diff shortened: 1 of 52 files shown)


nschloe:
Side note: It turns out h5py doesn't preserve the group ordering yet, cf. https://github.com/h5py/h5py/issues/1471, so a write-read might produce a different cell ordering.

tianyikillua:
For the MED format, I don't think it is possible to do
```
for cell_type, cells in mesh.cells:
    med_type = meshio_to_med_type[cell_type]
    med_cells = cells_group.create_group(med_type)
```
since the MED format doesn't support multiple same cell types. We have to convert `mesh.cells` to the former `dict` structure, that is to aggregate all cells with the same type.

nschloe:
This is basically done, I think I'll merge after dinner. I'm quite happy with the PR, especially since the new approach allowed to remove some costly `unique` calls on the cell types for VTK, VTU, XDMF.

meshio/_mesh.py
```diff
@@ -114,6 +114,13 @@ def write(self, path_or_buf, file_format=None, **kwargs):
 
         write(path_or_buf, self, file_format, **kwargs)
 
+    @property
+    def cells_dict(self):
+        assert len(self.cells) == len(
+            numpy.unique([c.type for c in self.cells])
+        ), "More than one block of the same type. Cannot create dictionary."
+        return {cell_type: data for cell_type, data in self.cells}
```
gdmcbain: An equivalent:
```python
return dict(self.cells)
```

nschloe: Much better, thanks.

tianyikillua:
Great work!

I suppose this PR should be accompanied by a another that updates `cell_data["name"]` also to a list of the same length as `cells`, that contains numpy cell values?

nschloe:
Cell data is also changed in this one, exactly to match `cells`. Now, `cell_data["name"]` is a list of arrays of the same length as `mesh.cells`, and `len(cell_data["name"][k]) == len(cells[k].data)`.

meshio/gmsh/_gmsh41.py
```diff
@@ -216,31 +216,15 @@ def _read_elements(f, point_tags, physical_tags, is_ascii, data_size):
...
-                    cell_data["gmsh:physical"][key] = {}
-                cell_data["gmsh:physical"][key] = physical_tag[0] * numpy.ones(
-                    len(values), int
-                )
+        cells.append((key, values))
+        if physical_tag:
+            if "gmsh:physical" not in cell_data:
+                cell_data["gmsh:physical"] = {}
+            cell_data["gmsh:physical"].append(
+                (key, physical_tag[0] * numpy.ones(len(values), int))
```
dalcinl: @nschloe Why you do `(key, tags)` in here? Using `tags` would be just fine, the cell type in here is redundant, and hardly useful if you have many blocks with the same cell type. The cell data ordering is implicit from the `mesh.cells` attribute. This broke compatibility with the gmsh 2.2 reader/writer, particularly if you want to read v4.1 and dump v2.2 (which I was using routinely before these changes). I'm working on fixing this issue. 

Would you accept this change (plus other stuff to make conversion from 4.1 to 2.2)?

```diff
--- a/meshio/gmsh/_gmsh41.py
+++ b/meshio/gmsh/_gmsh41.py
@@ -222,7 +222,7 @@ def _read_elements(f, point_tags, physical_tags, is_ascii, data_size):
             if "gmsh:physical" not in cell_data:
                 cell_data["gmsh:physical"] = []
             cell_data["gmsh:physical"].append(
-                (key, physical_tag[0] * numpy.ones(len(values), int))
+                physical_tag[0] * numpy.ones(len(values), int)
             )
```

nschloe: >  Using tags would be just fine

Yep, that's an oversight. PR welcome!

Issue #632, opened 2020-01-11
Implement AVS-UCD mesh file reader and writer


daniellivingston:
[AVS-UCD](https://dav.lbl.gov/archive/NERSC/Software/express/help6.1/help/reference/dvmac/UCD_Form.htm) is a very simple ASCII mesh file format that's used often in United States-based national labs. 

I would be happy to write a reader and writer if there are no plans to do so, as meshio support for this format would benefit my workflow quite a bit.

nschloe:
> I would be happy to write a reader and writer

Sure, go right ahead. Let me know if you need any help.

keurfonluu:
I was planning to implement it next week (end of my parental leave) as I am interested in using LaGrit. I'll look into that.

daniellivingston:
@keurfonluu Awesome, I'm one of the devs on LaGriT. I'd be happy to team up on this.

keurfonluu:
I've pushed the reader to my fork (https://github.com/keurfonluu/meshio/tree/avs). Would you mind taking a look at it @daniellivingston ?

It works for all the samples I've been able to find/produce, but I am far from being an expert with LaGriT.

There is also a few things I don't understand such as what is ``num_model_data`` and what to do with the first line of ``node_data`` and ``cell_data`` (e.g. ``002 1 1``).

nschloe:
Fixed by #633.

PR #633, opened 2020-01-14
Add AVS-UCD format

keurfonluu:
This PR adds support to AVS-UCD format following #632.

**Features**
- Reader and writer for AVS-UCD (ASCII) format,
- Support non-sequential ordering of nodes and cells,
- Material data is stored in ``avsucd:material`` and interoperability with formats ``gmsh``, ``medit`` and ``flac3d`` has been handled,
- Ready for next major patch of ``meshio`` (inverse cell data keys and use of f-strings).

**What's is missing?**
- Model data are skipped since it cannot be generalized to other formats,
- Data units are not saved.

**Note**
- White space in node and cell data labels are converted to underscores,
- No format extension registered since it is also ``inp`` which is reserved to Abaqus in ``meshio``.


nschloe:
This also needs an entry in the main README.

nschloe:
LGTM, thanks!

PR #650, opened 2020-01-30
Gmsh: various fixes and minor code refactor

dalcinl:
* Get rid of the `struct` module, just use numpy for I/O
* Fix v2.2 binary/ascii reader to return c_int arrays consistently
* Fix broken mesh.cells when reading binary v2.2
* Fix broken mesh.cell_data when reading v2.2
* Fix format for mesh.cell_data when reading v4.1
* Fix broken meshio <-> gmsh cell reordering and
  refactor code to use common utility routines

### DIFF
--- a/meshio/gmsh/_gmsh22.py
+++ b/meshio/gmsh/_gmsh22.py
@@ -3,7 +3,6 @@ I/O for Gmsh's msh format, cf.
 <http://gmsh.info//doc/texinfo/gmsh.html#File-formats>.
 """
 import logging
-import struct
 
 import numpy
 
@@ -11,7 +10,9 @@ from .._common import cell_data_from_raw, raw_from_cell_data
 from .._exceptions import ReadError, WriteError
 from .._mesh import Cells, Mesh
 from .common import (
+    _gmsh_to_meshio_order,
     _gmsh_to_meshio_type,
+    _meshio_to_gmsh_order,
     _meshio_to_gmsh_type,
     _read_data,
     _read_physical_names,
@@ -71,7 +72,9 @@ def read_buffer(f, is_ascii, data_size):
     for name, tag_dict in cell_tags.items():
         if name not in cell_data:
             cell_data[name] = []
-        cell_data[name] += list(tag_dict.values())
+        for cell_type, _ in cells:
+            tags = tag_dict.get(cell_type, [])
+            cell_data[name].append(tags)
 
     return Mesh(
         points,
@@ -116,6 +119,7 @@ def _read_cells(f, cells, is_ascii):
         _read_cells_ascii(f, cells, cell_tags, total_num_cells)
     else:
         _read_cells_binary(f, cells, cell_tags, total_num_cells)
+    cells[:] = _gmsh_to_meshio_order(cells)
 
     # Fast forward to $EndElements
     line = f.readline().decode("utf-8")
@@ -127,7 +131,6 @@ def _read_cells(f, cells, is_ascii):
     for cell_type in cell_tags:
         physical = []
         geometrical = []
-        # output_cell_tags[cell_type] = {"gmsh:physical": [], "gmsh:geometrical": []}
         for item in cell_tags[cell_type]:
             if len(item) > 0:
                 physical.append(item[0])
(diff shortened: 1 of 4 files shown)


dalcinl:
After these changes, I managed to run successfully my own exhaustive testing, which involves the tensor product of read high-order v2.2/v4.1 binary/ascii meshes (1D/2D/3D simplex/tensor cells and wedges), some of them with periodicity, and write a corresponding v2.2 binary/ascii "squeleton", low-order mesh extracted from the high-order one to feed PETSC's DMPlex, all that to run torture tests in a high-order, time-explicit, conservative, entropy-consistent hyperbolic+diffusion PDE solver. So I'm confident all these changes are right (although Python's duck typing bites from time to time).

meshio/gmsh/_gmsh22.py
```diff
@@ -205,13 +201,13 @@ def _read_cells_binary(f, cells, cell_tags, total_num_cells):
         count = shape[0] * shape[1]
         data = numpy.fromfile(f, count=count, dtype=c_int).reshape(shape)
 
-        if len(cells) == 0 or t != cells[-1][1]:
+        if len(cells) == 0 or t != cells[-1][0]:
             cells.append((t, []))
-        cells[-1][1].append(data[:, -num_nodes_per_elem:])
+        cells[-1][1].append(data[:, -num_nodes_per_elem:].astype(int))
```
nschloe: Why do we need `astype` here?

dalcinl: It is not strictly needed, but I did it this way to be consistent with the v4.1 reader.

> *A foolish consistency is the hobgoblin of little minds ...*

Unless you really hate it, I humbly request to keep it that way.

Or, maybe in a followup PR, we can rework this stuff. I believe you are worried about conversion warnings. IMHO, these warnings are annoying. I would rather emit these warnings ONLY if the conversion actually change values (let say testing with `np.all(old==new)` if `old.dtype != new.dtype`).

```python
n [1]: import numpy as np                                                      
In [2]: a = np.array([1<<40])                                                   
In [3]: b = a.astype(np.int32)                                                  
In [4]: if b.dtype != a.dtype: assert np.all(a==b)                              
---------------------------------------------------------------------------
AssertionError                         


n [5]: a = np.array([1<<30])                                                   

In [6]: b = a.astype(np.int32)                                                  

In [7]: if b.dtype != a.dtype: 
   ...:      assert np.all(a==b) 
   ...:      print("OK") 
   ...:                                                                         
OK
```

nschloe: Ah, I now see that we're talking about gmsh 2.2 here, which we'll see less and less, so I'm not too worried about that at all.

Type conversion can be really costly though, so we should always strive to avoid it. That's why there are warnings in place if the user provides the wrong type and we need to convert.

Question is: What's the data type before conversion?

dalcinl: Before conversion, the dtype is `np.int32`.

PS1: About v2.2 being phased out, let me say that our group is using Pointwise for mesh generation, and they export high order meshes in v2.2 format. I don't see them upgrading to v4.1. I'm still struggling to figure out how to write the `$Entities` section (otherwise I would have already contributed a PR to meshio!). I know this section is optional, but without it, how can we specify region tags for elements? Maybe we should ask Gmsh folks about this issue.

PS2: I understand that type conversions are costly. However, given that you are in the business of I/O in here, I would argue that we can hardly mesure the impact in performance.

nschloe: > Before conversion, the dtype is np.int32.

Aha, and `int` is probably `numpy.int64` on your machine.

I don't want to push too hard here, but what's the reason for the conversion again? Gmsh provides the values in `numpy.int32`, why should we change it?

dalcinl: Just for consistency with the output of the v4.1 reader. On 64bit archs, the v4.1 reader outputs `int64` natively (if the mesh was generated with a 64bit build of gmsh). 

PS: What machine are you running on??? I assume you are not on Windows, macOS is 64 bits, and it is quite hard to see out there Linux 32 systems (though I may be wrong, I'm a Fedora user, so I'm always in the bleeding edge).

nschloe: > and it is quite hard to see out there Linux 32 systems 

Also 64 bit of course. The thing is that gmsh 2.2 needs 32-bit int data, see the warning
```
"Binary Gmsh needs 32-bit integers (got %s). Converting.",
```
in the writer. That's perhaps probably what causes the issues for you.

I'd say, if you need 64-bit integer, you'll have to convert it yourself after reading the mesh. I agree that it's inconsistent between 2.2 and 4.1, but that's not meshio's fault but gmsh's.

dalcinl: Well, yes, if you read v22 and write v22, you get the warning. But not if you read v22 and write back as v41. On the other hand, if we maintain int32 on v22 read and write back to v41, now you get the warning! This is an impossible situation! That's the reason I advocated to silence the warnings, and emit them only if conversion actually changed values.

Let me add one extra point. Look at the v22 reader again, but the ASCII path. I certainly did not mess with conversions that code. And in current master, if you read v22 ASCII, then you return int64 values! Isn't that enough inconsistency for you? I suddenly got a headache just thinking about it :laughing:.

nschloe: > Let me add one extra point. Look at the v22 reader again, but the ASCII path. [...] And in current master, if you read v22 ASCII, then you return int64 values!

That's not very good indeed. Since binary gmsh 2.2 requires int32, ASCII reads should return int32 as well. This should defeat some of the inconsistencies you mentioned. For gmsh 4.1, we should have the same kind of consistency between ascii and binary (I think it's `c_size_t` there).

dalcinl: @nschloe I think I've addressed all your concerns, implemented the required consistency fixes for v2.2, and a minor maintenance enhancement (get rid of the `struct` module). I ran again my torture testing, and everything seems to be OK.

nschloe:
Thanks for the nice PR!

Issue #651, opened 2020-02-03
XDMF writer with vertex cells crashes ParaView


adtzlr:
Hello,

the following MWE causes ParaView to crash. It happens if vertex cells use their own nodes (not connected to the grid used by quads):

```python
import numpy as np
import meshio

filename = 'test.xdmf'

nodes= np.array([0,0,0, 
                 1,0,0,
                 2,0,0,
                 0,1,0, 
                 1,1,0,
                 2,1,0,
                 0.5,0.5,0,
                 1.5,0.5,0], dtype=float).reshape(-1,3)

connectivity = np.array([[0,1,4,3],
                         [1,2,5,4]])

conn = np.array([[-2],[-1]])
# change to 
# conn = np.array([[-4],[-3]])
# and it works....

cells = {'quad': connectivity, 'vertex': conn}

with meshio.xdmf.TimeSeriesWriter(filename) as writer:
    writer.write_points_cells(nodes, cells)
    
    for time in np.linspace(0,1,10):

        cell_data = {
                    'hexahedron': 
                    {'stress': np.random.rand(len(connectivity),6)},
                    'vertex': 
                    {'stress': np.random.rand(len(conn),6)}
                    }
        
        disp = np.random.rand(*nodes.shape)
        disp[:-1] = np.nan
            
        point_data = {'Displacement': disp}
        
        writer.write_data(time, 
                          point_data=point_data, 
                          cell_data=cell_data)
```

However, if vertex cells re-use nodes from quad elements, everything is fine. My idea was to add vertex cells as fake quadrature points to export my stresses, strains, kinematic variables at the correct location from my FE code. Beside this export problem in meshio - do you have any better ideas for that?

Best regards,
Andreas

keurfonluu:
You can display your displacement vectors as arrow glyphs.

adtzlr:
@keurfonluu Thank you for your answer. I think however you misunderstood me. I know I can display displacements - or in general any node related data - as point data. I was thinking of exporting stresses for linear quadrilaterals at their calculated gauss points. The other way round would be to interpolate the gauss point data to the nodes, perform nodal averaging and export them also as point data with meshio. I thought I could use some magic of meshio or paraview for that :-) .

BR Andreas

gdmcbain:
[...]

PR #652, opened 2020-02-03
Polyvertex needs number of cells (1) in XDMF mixed topology

gdmcbain:
This is for #651 but only treats `XdmfWriter`, not `TimeSeriesWriter` as asked for; the fix for the latter should be similar.

Basically, although it's explicitly not stated in http://www.xdmf.org/index.php/File:MixedTopology.png, Polyvertex does need to specify the number of points in XDMF `TopologyType="Mixed"`.


nschloe:
Thanks, Geordie!

PR #663, opened 2020-02-05
Fixes and update to pass Sfepy tests

vlukes:
We would like use meshio in our SfePy project [1]. To pass some of our mesh reading/writing tests I have made several updates and fixes. We need to keep information about point and cell groups (subdomains, boundaries, etc.) when reading and writing various formats.

[1] http://sfepy.org

### DIFF
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -138,7 +138,7 @@ def read_buffer(f):
             name = params_map["ELSET"]
             if name not in cell_sets:
                 cell_sets[name] = []
-            cell_sets[name].append(setids)
+            cell_sets[name].append(setids - 1)
         else:
             # There are just too many Abaqus keywords to explicitly skip them.
             line = f.readline()
@@ -150,6 +150,7 @@ def read_buffer(f):
         cell_data=cell_data,
         field_data=field_data,
         point_sets=point_sets,
+        cell_sets=cell_sets,
     )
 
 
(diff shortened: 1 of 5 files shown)


nschloe:
Hi Vladimir,

thanks for a great PR! The linter isn't happy yet; you can just run `make format` in meshio for the necessary changes.

meshio/abaqus/_abaqus.py
```diff
@@ -245,17 +247,17 @@ def _read_set(f, params_map):
...
-        if line.startswith("*"):
+        if not line or line.startswith("*"):
             break
         if line.strip() == "":
             continue
 
         set_ids += [int(k) for k in line.strip().strip(",").split(",")]
 
-    if "generate" in params_map:
+    if "GENERATE" in params_map:
```
nschloe: Is this always capitalized? If not, we could check for either or the two. Don't know the abaqus spec by heart.

meshio/exodus/_exodus.py
```diff
@@ -141,8 +141,8 @@ def read(filename):  # noqa: C901
         # elif key == "eb_names":
         #     value.set_auto_mask(False)
         #     eb_names = [b"".join(c).decode("UTF-8") for c in value[:]]
-        elif key == "node_ns":
-            ns = value
+        elif key.startswith("node_ns"):
```
nschloe: Out of interest: Why `startswith`?

meshio/abaqus/_abaqus.py
vlukes: You are right, it does not have to be capitalized. So I added `upper()` when parsing the arguments.

meshio/exodus/_exodus.py
vlukes: It is because the keys for node sets are `node_ns1`, `node_ns2`, etc.

nschloe: Best add a comment about that.

vlukes: OK, the comment added.

tianyikillua:
For Abaqus, your current implementation of `cell_sets` only store the raw Abaqus cell id's...which is useless outside the INP file (since these point id's / cell id's are not exposed in the `Mesh` object). These id's need to be converted to cell ordering as is in `mesh.cells`. With the new cells / cell_data convention, if we have two types of cells
```
mesh.cells = [("triangle", ...), ("quad", ...)]
```
and a particular cell set `A` is composed of the 1st triangle element, as well as the 2nd and the 6th quad elements, we should have
```
mesh.cell_sets = [(0), (1, 5)]
```

vlukes:
I agree with @tianyikillua, this implementation is only for keeping `cell_sets` when converting via INP format. It should be improved. @tianyikillua, do you have any such Abaqus INP file that contains triangle and quad (or tetra and hex) elements and various cell sets?

tianyikillua:
> I agree with @tianyikillua, this implementation is only for keeping `cell_sets` when converting via INP format. It should be improved. @tianyikillua, do you have any such Abaqus INP file that contains triangle and quad (or tetra and hex) elements and various cell sets?

@vlukes 

Already fixed in https://github.com/nschloe/meshio/pull/666 :-)

I've added a test mesh (`nle1xf3c.inp`) containing two sets of triangles.

You can find lots of example files [here](http://dsk.ippt.pan.pl/docs/abaqus/v6.13/books/bmk/default.htm?startat=ch01s19ach135.html).

Issue #643, opened 2020-01-22
VTK output file crashes paraview (paraview version 5.7.0)


firstkingofrome:
Hello I have a number of gmesh files that I would like to convert to vtk. The vtk file I produce (I tried both ascii and binary) cause paraview to crash when I attempt to render them. I have attached an input gmesh file and output ascii vtk. file.
I converted by simply loading a test file and attempting to rewrite it as a vtk file:
import meshio
testFile = "gmshRes.mesh.0.msh"
msh = meshio.read(testFile) 
meshio.write("test.vtk",msh,file_format="vtk-ascii")

Based on this discussion https://discourse.paraview.org/t/paraview-immediately-crashing-upon-when-opening-a-legacy-vtk-file-created-using-python/1104 I thought that it might be a problem with the CELLS block?

Thank You for helping me look into this issue!

(Note I changed added a .txt extension to the gmesh and vtk files I was describing in order to get github to accept them)

[gmshRes.mesh.0.msh.txt](https://github.com/nschloe/meshio/files/4100637/gmshRes.mesh.0.msh.txt)

[test.vtk.txt](https://github.com/nschloe/meshio/files/4100643/test.vtk.txt)

gdmcbain:
That test.vtk crashes Gmsh 4.5.1 too.

I note that MSH 2.2 can also be converted to VTK by Gmsh with
```shell
gmsh -2 gmshRes.mesh.0.msh -o gmshRes.mesh.0.vtk
```
which gives [gmshRes.mesh.0.vtk.txt](https://github.com/nschloe/meshio/files/4100727/gmshRes.mesh.0.vtk.txt) which doesn't look too bad (opens in ParaView 5.6.0, e.g.).

This might be helpful for seeing where meshio is going wrong here.  I see that the original MSH 2.2 has 3784 nodes as does the meshio VTK but the Gmsh VTK has only 3272.

gdmcbain:
Gmsh 4.5.1 crashes pretty quickly on opening test.vtk, but I caught the error messages in 
[test.txt](https://github.com/nschloe/meshio/files/4100759/test.txt) with
```shell
gmsh -log test.txt test.vtk 
```
Looks like:
```
Info: Reading 'test.vtk'...
Info: Reading 3784 points
Info: Reading 2564 cells
Error: Wrong node index 6711
Error: Wrong node index 6710
…
```
and many more like the last two.

[...]

PR #665, opened 2020-02-05
Fixes for following formats: vtk, medit, med

vlukes:
### VTK

- saving call data was not correctly implemented, tags `CELL_DATA` and `FIELD FieldData` should be in VTK file only once
- limit to `FeildData` dimension has no reasons

### MEDIT

- `"medit:ref"` for cells must be a list of arrays containing group ids

### MED

- `FAS` object  can be located in the mesh object, not only in the root


meshio/vtk/_vtk.py
```diff
@@ -629,10 +629,10 @@ def pad(array):
             _write_field_data(f, mesh.point_data, binary)
 
         # write cell data
-        for name, data in mesh.cell_data.items():
-            total_num_cells = sum([len(c) for c in data])
+        if mesh.cell_data:
+            total_num_cells = sum(len(c.data) for c in mesh.cells)
             f.write(f"CELL_DATA {total_num_cells}\n".encode("utf-8"))
-            _write_field_data(f, raw_from_cell_data(mesh.cell_data), binary)
+            _write_field_data(f, mesh.cell_data, binary)
```
nschloe: Can we really just write `mesh.cell_data`? I should be a dict of lists of numpy.arrays, really. `raw_from` converts this to a dict of numpy.array by concatenation.

vlukes: Yes, `mesh.cell_data` is a dict of lists of arrays and the lists must be joined. This is done in  `_write_field_data()`, if `data` is a list. This produces just one `CELL_DATA` + `FIELD` section followed by the fields.

There was probably something strange in the commit, I didn't see the changes in `_write_field_data()`, after re-pushing, I see it again.

vlukes:
The fix for vtk should solve #643.

gdmcbain:
Oh, I thought the bug #643 was in reading MSH 2.2 rather than the subsequent write to VTK.

vlukes:
Yes, I see, there is another issue. In the MSH input file in #643, the points are not numbered in a continuous range.

meshio/vtk/_vtk.py
nschloe: Alright, since the tests pass, let's give it a go.

PR #667, opened 2020-02-05
Paraview plugin

nschloe:


tianyikillua:
Great! I will leave a message at https://github.com/tianyikillua/paraview-meshio saying that it's no longer maintained and users should directly use the version bundled with meshio.

PR #668, opened 2020-02-06
fix gmsh22: remap cell connectivity

vlukes:
Fix for #643 (tested). The cell connectivity must be remapped for MSH22. For MSH40 and MSH41, this seems to be done.


nschloe:
Very nice, thanks!

PR #671, opened 2020-02-07
Add Tecplot format

keurfonluu:
This PR adds support to Tecplot ASCII data format. Tecplot is a popular commercial technical visualization software widely used in the world, thus many simulators (e.g. ANSYS FLUENT) can export their results to Tecplot format.

**Features**
- Reading finite-element type (unstructured) grids (``FEPOINT`` and ``FEBLOCK``),
- Writing ``meshio.Mesh`` to ``FEBLOCK`` format (``FEPOINT`` does not appear to be able to store cell data).

**What is missing?**
- Only support one zone,
- Only support 2D and 3D cells,
- ``VARLOCATION`` cannot be defined as a list,
- Reading ordered (structured) grids (it should be quite straightforward to implement).

**Note**
- Sample data files can be found [here](https://people.sc.fsu.edu/~jburkardt/data/tec/tec.html),

**TODO**
Improve reading of zone record (lines 178-222): how to split string
``ZONE N = 1, E = 2, VARLOCATION = ([1-2]=NODAL, [3-4]=CELLCENTERED)``
into a dict
```python
zone = {
    "N": "1",
    "E": "2",
    "VARLOCATION": "([1-2]=NODAL, [3-4]=CELLCENTERED)",
}
```
with ``VARLOCATION`` preventing the use of ``.split()``. This does not impact much the speed, though, but it is a pretty ugly workaroung IMHO (although it works).


nschloe:
Very cool!

PR #673, opened 2020-02-08
UGRID: fix wrong tag ordering

ChristosT:
In ugrid reader order of reading the surface elements is important. 

1.  Added a test case that exposes this issue.
2.  fixed issue and updated ugrid tests

Note: Another way to visually validate the problem and the solution is to use meshio-convert to convert any ugrid that has both quads and triangles into vtk. Without this patch surface elements are not connected properly. Applying the patch produces the expected result. 

Also, I introduced a function in  `test_ugrid.py` (`first()`) to pick elements by type just for convenience. It can be removed if this is needed.


keurfonluu:
Run ``make format`` to make the linter happy.

test/test_ugrid.py
```diff
@@ -12,7 +12,7 @@
     [
         helpers.tri_mesh,
         helpers.quad_mesh,
-        helpers.quad_tri_mesh,
+        helpers.tri_quad_mesh,
```
nschloe: Let's test both tri_quad and quad_tri here.

ChristosT: ok no problem

ChristosT: it actually helped to find some issues and improve the code. Thanks for the idea.

nschloe:
LGTM, thanks!

Issue #619, opened 2020-01-05
Unification of point / cell sets among formats


tianyikillua:
## Objectives

Unify the names (`gmsh:physical`, `medit:ref`, `flac3d:zone`...) and implementations of point / cell sets among formats.

### Names

Currently these different names need to be hard-coded when the subsets are transferred among formats, like
```
if key in mesh.cell_data and "medit:ref" in mesh.cell_data[key]:
    labels = mesh.cell_data[key]["medit:ref"]
elif key in mesh.cell_data and "gmsh:physical" in mesh.cell_data[key]:
    # Translating gmsh data to medit is an important case, so treat it
    # explicitly here.
    labels = mesh.cell_data[key]["gmsh:physical"]
elif key in mesh.cell_data and "flac3d:zone" in mesh.cell_data[key]:
    labels = mesh.cell_data[key]["flac3d:zone"]
else:
    labels = numpy.ones(len(data), dtype=int)
```

**Two solutions**

1. Either provide a list of these names like
```
CELL_SETS_NAMES = [
    "gmsh:physical", "medit:ref", "flac3d:zone", ...
]
```

2. Adopt a unified name and object for point / cell sets, like
```
mesh.point_sets = ...
mesh.cell_sets = ...
```

### Implementations

An example of mesh composed of 5 line elements and 6 nodes. We have two point sets: "Red" and "Green", and two cell sets: "Orange" and "Blue".

![subsets](https://user-images.githubusercontent.com/4027283/71777726-2b8ca100-2fa4-11ea-8bac-b0da3efee579.png)

1. Use a `point_data` and `cell_data` as indicator functions. Using the current terminology for MED, we have
```
mesh.point_data["point_tags"] = [0, 1, 1, 2, 1, 0]
mesh.point_tags = {
    1: ["Red"],
    2: ["Green"],
}

mesh.cell_data["line"]["cell_tags"] = [-1, -2, -1, 0, -2]
mesh.cell_tags = {
    -1: ["Orange"],
    -2: ["Blue"],
}
```
Note that with this implementation, we need two lists: the indicator function list that assigns a unique value to each point / cell, as well as a list that indicates the actual names corresponding to these unique values. This implementation is suited when we want to visualize point / cell sets in ParaView, since they are modeled as a `point_data` / `cell_data`.

This implementation is currently adopted by (to be completed)
- gmsh
- medit
- med
- flac3d

2. Explicit enumerations of points / cell id's
```
mesh.point_sets = {
[...]

PR #691, opened 2020-02-13
Unifying point and cell tags

tianyikillua:
Related issue #619

**Already covered** (+ before this PR)
- abaqus: point and cell sets
- avsucd: cell tags
- exodus: point sets
- flac3d: cell tags
- gmsh22: cell tags
- med: point and cell tags
- medit: point and cell tags
- permas: point and cell sets
- ugrid: cell tags

**TODO**
- gmsh44: cell tags

Repr of mesh also now prints point / cell tags.
```
In [4]: meshio.read("flac3d/flac3d_mesh_ex.f3grid")                             
Out[4]: 
<meshio mesh object>
  Number of points: 160
  Number of cells:
    hexahedron: 45
    pyramid: 9
    hexahedron: 18
    wedge: 9
    hexahedron: 6
    wedge: 3
    hexahedron: 6
    wedge: 3
    pyramid: 6
    tetra: 3
  Cell data: flac3d:zone
  Cell tags: 1, 2, 3, 4

In [5]: meshio.read("flac3d/flac3d_mesh_ex.f3grid").cell_tags                   
Out[5]: 
[array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
        1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
        2]),
 array([2, 2, 2, 2, 2, 2, 2, 2, 2]),
 array([4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4]),
 array([4, 4, 4, 4, 4, 4, 4, 4, 4]),
 array([3, 3, 3, 3, 3, 3]),
 array([3, 3, 3]),
 array([3, 3, 3, 3, 3, 3]),
 array([3, 3, 3]),
 array([3, 3, 3, 3, 3, 3]),
 array([3, 3, 3])]
```

Now point / cell tags can be transferred among formats, through `mesh.point_tags` and `mesh.cell_tags`. Here the cell data `gmsh:physical` is transferred into `med:family`.

```
In [3]: meshio.read("msh/insulated-2.2.msh")                                    
Out[3]: 
<meshio mesh object>
  Number of points: 67
  Number of cells:
    line: 21
    triangle: 111
  Cell data: gmsh:physical, gmsh:geometrical
  Cell tags: 1, 2, 3

mesh = meshio.read("msh/insulated-2.2.msh")
mesh.write("test.med")
mesh = meshio.read("test.med")

In [14]: mesh                                                                   
Out[14]: 
<meshio mesh object>
  Number of points: 67
  Number of cells:
    line: 21
    triangle: 111
  Cell data: med:family
  Cell tags: 1, 2, 3
```


ChristosT:
> ugrid: cell tags (WARNING after reviewing the code, it seems that when writing it does not support cells that contain multiple blocks of a same cell type due to ugrid_counts[key] = data.shape[0], also should in fact "triangle" be before "quad"?)

Hi,

thank you for working on this. 
ugrid indeed does not support multiple blocks of the same type. Each block appears exactly once and in a specific order. As for the order or triangle and quad . I recently submitted a PR (#673) that handles this case the right way.

nschloe:
I'll wait for this to get ready before the 4.0 release.

tianyikillua:
I'm no expert on the MSH4 format. Maybe @gdmcbain can take the relay for this? As I understand we'll create a `cell_set` and not a standard `cell_tag` since cells can have to different tags.

nschloe:
Maybe we don't need the `self.{cell,point}_tags_key`. Since dicts are ordered, how about just sticking to the convention to write out the _first_ entry in the dict if the file format supports only one tag?

tianyikillua:
> Maybe we don't need the `self.{cell,point}_tags_key`. Since dicts are ordered, how about just sticking to the convention to write out the _first_ entry in the dict if the file format supports only one tag?

1. I saw Python 3.5 came back...in which cell keys are not ordered..
2. Explicit is better than implicit (PEP8?). Implicit conventions may become hidden and make debugging harder (in 1 month?).
3. At least I can combine these two into a single key `tags_key`. If `point_data` does not contain this key currently it also returns `None`.

nschloe:
> I saw Python 3.5 came back...in which cell keys are not ordered..

The reason for 3.5 to come back is to make it easier for downstream packages to use meshio. The further down the food chain, the more conservative one has to be. It'd say that this little feature shouldn't stop us here.

For me it really seems that, if not all `point_sets` can be written into a file, it's a shortcoming of the format and one should expect nothing. It's got nothing to do with the innocent `Mesh` object that carries all `point_sets` like an old, faithful donkey. It shouldn't have to carry that weird priority tag, too.

If a mesh format writer cannot write more than one `point_set`, let's allow a `point_set_priority` keyword to the writer, picking the set. If not provided, and if more than one set is present, write out _any_ (e.g., the first, it doesn't matter) and emit a warning (`Omitting sets "foo", "bar", and "bla".).

nschloe:
Even better:

No `point_set_priority`. Just force the user to provide on _one_ `point_set` if she means to write that one out. Otherwise it's the first.

tianyikillua:
I see your point.

The fact that the `Mesh` object contains this *useless* and strange attribute `mesh.tags_key` also disturbs me...If we want `mesh.point_tags` or `mesh.cell_tags` to just a shortcut to a certain value in `point_data` and `cell_data` we have to provide a key name.

Otherwise we need to convert these plain dicts to `OrderedDict` to enforce [dict ordering](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6). Then make sure the tag arrays always come first or last. Then we still need to have a bool attribute `mesh.tags_exist`. For me it becomes more complicated...than just providing a key name.

If we don't want neither `mesh.tags_key` nor `mesh.tags_exist`, we have to **copy** `mesh.cell_data[mesh.tags_key]` to `mesh.cell_tags`. Idem for point tags.

nschloe:
Maybe I misunderstand something here. Don't all values in the `point_sets` serve as tag sets?

I'll write that down for myself: The reason why I don't like additional attributes is that you write code like this:
```python
mesh = generate_somehow()
# now mesh contains multiple point_sets
print(mesh.point_sets)    # "aha, good!"
mesh.write("out.foo")
# inspect out.foo ... "what?"
```
where the `foo` format can only take _one_ point set. `write` chose the set `"barset"`, but why that?? Aha, because  `mesh.tags_key` is `"barset"`. This is not at all obvious though. It'd be much better if it went like this:
```python
mesh = generate_somehow()
# now mesh contains multiple point_sets
print(mesh.point_sets)    # "aha, good!"
mesh.write("out.foo")  # warning: cannot write more than one point set. choosing "barset`, skipping "fooset", "testset"`.
```

tianyikillua:
Currently `point_sets` and `cell_sets` are not converted to `point_tags` and `cell_tags`. This PR should only cover unification of tags (like `gmsh:physical`). Do you mean in fact `point_tags` and `cell_tags` in your previous comments?

These two tags are just a function that assigns an integer to each point / cell. And in general indeed only **one unique** point/ cell tag function exists inside a mesh file, except for GMSH 2 where two exist. In the latter case we just choose `gmsh:physical`.

gdmcbain:
>  And in general indeed only one unique point/ cell tag function exists inside a mesh file, except for GMSH 2 where two exist. In the latter case we just choose gmsh:physical.

Thank goodness MSH 2.2 is alone. I was worried someone was going to advance another pathological format. 

Quirkily, when Gmsh writes MEDIT it uses its 'elementary' tag for the single MEDIT ref of each cell. This is going back a decade or so. The motivation, I was told, was that a cell might have multiple physical tags but only ever one elementary tag. (This despite the fact that Gmsh erroneously never actually wrote MSH 2.2 cells with multiple physical tags, it multiplied the cells.) Since elementary tags are irrelevant to finite element applications, and MEDIT is good input for FreeFEM, I wrote the Gmsh list about this and the developers added an option [Mesh.SaveElementTagType](http://gmsh.info/dev/doc/texinfo/gmsh.html#index-Mesh_002eSaveElementTagType) which could be used to save the physical tag instead. (Or at least one of the physical tags, though the multiple physical tags question was moot. The stated behaviour was to save an arbitrary one of the physical tags. I haven't reinvestigated this since Gmsh 4 started handling multiple physical tags properly.)

tianyikillua:
The `tags_key` attribute is now removed, since in Python no copy is done when dong `point_tags = point_data["med:family"]`.

nschloe:
As outlined in https://github.com/nschloe/meshio/pull/691#issuecomment-586493507, closing in favor of https://github.com/nschloe/meshio/pull/704.

PR #702, opened 2020-02-17
Format config

nschloe:


meshio/abaqus/_abaqus.py
```diff
@@ -271,13 +271,13 @@ def _read_set(f, params_map):
     return set_ids, line
 
 
-def write(filename, mesh, translate_cell_names=True):
+def write(filename, mesh, float_fmt=".15e", translate_cell_names=True):
```
lrineau: Why 15 and not 17?

nschloe: The double-precision resolution is 2^-53 = 1.11e−16, so 15 is the number of guaranteed correct digits. Note that you can change it to whatever you like; it's a parameter.

ASCII formats are only for debugging anyway. They are slow and cut off or add information.

lrineau: I had not chosen 17 by chance. 17 digits is the minimal number of digits for which the chain of conversions binary->ascii->binary is guaranteed to be exact. See the the [definition of `DBL_DECIMAL_DIG` (C11/C++17)](https://en.cppreference.com/w/c/types/limits).

Your calculus is right:
```
  log10(2^53)=15.95...
```
That means that you need at least 16 digits to correctly represent in ASCII a binary `double`, otherwise one looses precision. But, for a technical reason (denormals), one more digit is needed. See also the Wikipedia page  [IEEE_754, section Character representation](https://en.wikipedia.org/wiki/IEEE_754#Character_representation).

lrineau: > ASCII formats are only for debugging anyway. They are slow and cut off or add information.

I agree. But meshio only supports the ASCII Abaqus format, for the moment. That is fine to exchange with users of Abaqus, but I need 17 digits of precision, to guarantee that the mesh is right.

nschloe: One has to make a choice here, right? A double-precision float does not have an exact representation in decimal format. You either lose information (when using 15 or fewer digits) or add information (when using 16 or more digits). By default, I prefer not adding info which isn't really there, even if that means that information is lost.

Is it a problem for your application to use the `float_fmt` parameter?

nschloe: > But meshio only supports the ASCII Abaqus format, for the moment.

That should probably be fixed. Do you happen to know where to get a format specification?

lrineau: > One has to make a choice here, right? A double-precision float does not have an exact representation in decimal format. You either lose information (when using 15 or fewer digits) or add information (when using 16 or more digits). By default, I prefer not adding info which isn't really there, even if that means that information is lost.
> 
> Is it a problem for your application to use the `float_fmt` parameter?

Actually, as I am not a Python develop, I use the `meshio-convert` utility, and, as far as I know, that utility does not expose that configuration possibility.

nschloe: Understood. I'll add a command-line parameter.

nschloe: https://github.com/nschloe/meshio/pull/703

lrineau: Great, thanks a lot!

PR #710, opened 2020-02-19
Add quad9 to meshio_types

jorgensd:
Adding missing type supported by XDMF


nschloe:
Those exist in XDMF? Well, I guess http://www.xdmf.org/index.php/XDMF_Model_and_Format#Topology isn't complete.

gdmcbain:
Not that the DTD is definitive either, but it is there: https://gitlab.kitware.com/xdmf/xdmf/blob/master/Xdmf.dtd#L34

nschloe:
Adding more at https://github.com/nschloe/meshio/pull/712.

PR #714, opened 2020-02-19
Replace ModuleNotFoundError by ImportError

keurfonluu:
- Fixed: ``ModuleNotFoundError`` is only available in ``python>=3.6``.


nschloe:
I want a comment somewhere saying that `ModuleNotFoundError` is to be used from Python 3.6 on.

PR #727, opened 2020-02-21
Read and write info from/to vtk, medit, xdmf

vlukes:
This PR allows to read and write application specific information (e.g. bounds, post-processing data, time steps, etc.) in VTK, MEDIT and XDMF formats.

The information is stored in:
- VTK: title line
- MEDIT: comment line
- XDMF: `Information` tag

### DIFF
--- a/meshio/medit/_medit.py
+++ b/meshio/medit/_medit.py
@@ -30,6 +30,7 @@ def read_buffer(f):
     cells = []
     point_data = {}
     cell_data = {"medit:ref": []}
+    spec_info = []
 
     meshio_from_medit = {
         "Edges": ("line", 2),
@@ -47,7 +48,10 @@ def read_buffer(f):
             break
 
         line = line.strip()
-        if len(line) == 0 or line[0] == "#":
+        if len(line) == 0:
+            continue
+        if line[0] == "#":
+            spec_info.append(line[1:])
             continue
 
         items = line.split()
@@ -97,7 +101,9 @@ def read_buffer(f):
             if items[0] != "End":
                 raise ReadError("Unknown keyword '{}'.".format(items[0]))
 
-    return Mesh(points, cells, point_data=point_data, cell_data=cell_data)
+    return Mesh(
+        points, cells, point_data=point_data, cell_data=cell_data, info=spec_info,
+    )
 
 
 def write(filename, mesh, float_fmt=".15e"):
--- a/meshio/medit/_medit.py
+++ b/meshio/medit/_medit.py
@@ -106,8 +106,12 @@ def read_buffer(f):
     )
 
 
-def write(filename, mesh, float_fmt=".15e"):
+def write(filename, mesh, float_fmt=".15e", info=None):
     with open_file(filename, "wb") as fh:
+        # write specific information as comments
+        if info:
+            fh.write(bytes("".join("# %s\n" % i for i in info), "utf-8"))
+
         version = {numpy.dtype(c_float): 1, numpy.dtype(c_double): 2}[mesh.points.dtype]
         # N. B.: PEP 461 Adding % formatting to bytes and bytearray
         fh.write(b"MeshVersionFormatted %d\n" % version)
--- a/meshio/vtk/_vtk.py
+++ b/meshio/vtk/_vtk.py
@@ -148,7 +148,7 @@ def read_buffer(f):
 
     # skip header and title
     f.readline()
-    f.readline()
+    spec_info = [f.readline().decode("utf-8")]
 
     data_type = f.readline().decode("utf-8").strip().upper()
     if data_type not in ["ASCII", "BINARY"]:
@@ -182,6 +182,7 @@ def read_buffer(f):
         point_data=info.point_data,
         cell_data=cell_data,
         field_data=info.field_data,
+        info=spec_info,
     )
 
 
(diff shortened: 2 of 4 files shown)


meshio/medit/_medit.py
```diff
@@ -97,11 +101,17 @@ def read_buffer(f):
...
                 raise ReadError("Unknown keyword '{}'.".format(items[0]))
 
-    return Mesh(points, cells, point_data=point_data, cell_data=cell_data)
+    return Mesh(
+        points, cells, point_data=point_data, cell_data=cell_data, info=spec_info,
+    )
 
 
-def write(filename, mesh, float_fmt=".15e"):
+def write(filename, mesh, float_fmt=".15e", info=None):
```
nschloe: No need for an extra info argument. Let's just take `mesh.info`.

meshio/medit/_medit.py
```diff
@@ -47,7 +48,10 @@ def read_buffer(f):
             break
 
         line = line.strip()
-        if len(line) == 0 or line[0] == "#":
+        if len(line) == 0:
+            continue
+        if line[0] == "#":
+            spec_info.append(line[1:])
```
nschloe: Let's just call this `info`, not `spec_info`.

meshio/medit/_medit.py
```diff
@@ -47,7 +48,10 @@ def read_buffer(f):
             break
 
         line = line.strip()
-        if len(line) == 0 or line[0] == "#":
+        if len(line) == 0:
+            continue
+        if line[0] == "#":
```
nschloe: _All_ comments are considered information? Well, I guess one can't say any better.

meshio/vtk/_vtk.py
```diff
@@ -581,7 +582,7 @@ def translate_cells(data, types, cell_data_raw):
     return cells, cell_data
 
 
-def write(filename, mesh, binary=True):
+def write(filename, mesh, binary=True, info=None):
```
nschloe: Likewise.

meshio/vtk/_vtk.py
```diff
@@ -616,8 +617,13 @@ def pad(array):
         logging.warning("VTK ASCII files are only meant for debugging.")
 
     with open_file(filename, "wb") as f:
+        if info is None:
+            info_line = "written by meshio v{}\n".format(__version__)
+        else:
+            info_line = "%s\n" % info[0]
```
nschloe: Assert that `len(jnfo) == 0` here.

nschloe: Also, don't use the `%`-style formatting.

meshio/xdmf/main.py
```diff
@@ -302,18 +297,27 @@ def read_xdmf3(self, root):  # noqa: C901
...
 class XdmfWriter:
     def __init__(
-        self, filename, mesh, data_format="HDF", compression="gzip", compression_opts=4
+        self,
+        filename,
+        mesh,
+        data_format="HDF",
+        compression="gzip",
+        compression_opts=4,
+        info=None,
```
nschloe: Likewise.

meshio/xdmf/main.py
```diff
@@ -333,6 +337,10 @@ def __init__(
 
         xdmf_file = ET.Element("Xdmf", Version="3.0")
 
+        if info:
```
nschloe: This needs a read method, too.

nschloe:
Tests are needed for this. Best add a `add_info()` method in `tests/helpers.py`, equip and mesh with same information and assert that input and output information is equal.

meshio/medit/_medit.py
vlukes: If I do this
 ```
 m = meshio.read('in.xdmf')
 m.write('out.vtk')
```
I want want the information from `in.xdmf` not be copied into `out.vtk` because it is mostly related to the format. In this case, `written by meshio v4.0.3` should be in the VTK file. The information should only be written if the user explicitly states it as an argument. Do you agree?

meshio/medit/_medit.py
vlukes: I started with XDMF `time_series.py` and there was already one `info` variable, so I used `spec_info`. And to be consistent, it is also in XDMF/main.py, MEDIT and VTK.
- `spec_info` in XDMF `time_series.py` and `info` everywhere else?
- `spec_info` everywhere?

meshio/medit/_medit.py
vlukes: Information is usually in the first line, but can be elsewhere.

meshio/vtk/_vtk.py
vlukes: Zero-length `info` is now tested. Also, `info[0]` longer than 255 characters is truncated.

String formatting fixed.

meshio/xdmf/main.py
vlukes: It is already there in `read_xdmf2()` and `read_xdmf3()`.

vlukes:
Yes, I can do some tests, but first let's decide whether to copy `mesh.info`.

nschloe:
Okay, I've made up my mind about this.

I want an extra attribute to `mesh` named `commentary`, being a (possibly empty) list of strings. When reading a file, the list is to be filled. When writing, all the commentary is to be put at the beginning of the file. If the format does not permit arbitrary comments (like VTK), then issue a warning and write what you can. If the list is empty, write the usual `# This file was created by meshio v...`.

It's probably best to start off from scratch with this, but I'll leave it up to you.

vlukes:
Okay, I will start from scratch.

PR #741, opened 2020-02-29
Improve robustness of Tecplot reader

keurfonluu:
**Reader**
- Changed: improve reading of zone entries,
- Fixed: handle entries separated by white spaces.

**Tests**
- Added: test weirdly formatted ``VARIABLES`` and ``ZONE`` records (but still valid).


nschloe:
Thanks!

PR #750, opened 2020-03-10
fix vtu writer: numeric format in text mode

vlukes:
When writing `VTU` files in the text mode, the numeric format of an array is chosen according its data type instead of using the fixed format. The fixed format makes troubles (Paraview crashes) for `PointData` and `CellData` arrays.


nschloe:
Really nice PR, thank you. Those are my favorites: They simplify _and_ fix things.

Issue #536, opened 2019-11-25
GMSH 4.1 / Elements with multiple physical tags / meshio ignores all but the first


hoelzlw:
I do not know, whether this is already mentioned somewhere, but I could not find it in the documentation.

I think, gmsh 4.0 is affected as well.

In gmsh one can assign an element to multiple physical groups, like in the example mesh file given below:
<details>

$MeshFormat
4.1 0 8
$EndMeshFormat
$PhysicalNames
7
1 3 "Mechanically Clamped"
1 4 "Mechanically Free"
1 5 "Electrically Clamped"
1 6 "Electrically Grounded"
1 7 "Electrically Free"
2 1 "Structural Mechanics"
2 2 "Piezoelectricity"
$EndPhysicalNames
$Entities
6 7 2 0
1 0 0 0 0 
2 2 0 0 0 
3 0 1 0 0 
4 2 1 0 0 
5 0 2 0 0 
6 2 2 0 0 
1 -9.999999994736442e-008 -1e-007 -1e-007 2.0000001 1e-007 1e-007 1 4 2 1 -2 
2 -9.999999994736442e-008 0.9999999000000001 -1e-007 2.0000001 1.0000001 1e-007 1 6 2 3 -4 
3 -9.999999994736442e-008 1.9999999 -1e-007 2.0000001 2.0000001 1e-007 2 4 5 2 5 -6 
4 1.9999999 -9.999999994736442e-008 -1e-007 2.0000001 1.0000001 1e-007 1 4 2 2 -4 
5 1.9999999 0.9999999000000001 -1e-007 2.0000001 2.0000001 1e-007 2 4 7 2 4 -6 
6 -1e-007 -9.999999994736442e-008 -1e-007 1e-007 1.0000001 1e-007 1 3 2 1 -3 
7 -1e-007 0.9999999000000001 -1e-007 1e-007 2.0000001 1e-007 2 3 7 2 3 -5 
1 -9.999999994736442e-008 -9.999999994736442e-008 -1e-007 2.0000001 1.0000001 1e-007 1 1 4 6 2 -4 -1 
2 -9.999999994736442e-008 0.9999999000000001 -1e-007 2.0000001 2.0000001 1e-007 2 1 2 4 7 3 -5 -2 
$EndEntities
$Nodes
15 13 1 13
0 1 0 1
1
0 0 0
0 2 0 1
2
2 0 0
0 3 0 1
3
0 1 0
0 4 0 1
4
2 1 0
[...]

PR #756, opened 2020-03-16
write cell_sets more carefully

gdmcbain:
Each value in cell_sets now has cell-indices carefully written to the correctly indexed item in the list corresponding to the cell-blocks.

### DIFF
--- a/meshio/gmsh/_gmsh41.py
+++ b/meshio/gmsh/_gmsh41.py
@@ -55,14 +55,13 @@ def read_buffer(f, is_ascii, data_size):
 
         if environ == "PhysicalNames":
             _read_physical_names(f, field_data)
-            cell_sets = {k: [] for k in field_data.keys()}
         elif environ == "Entities":
             physical_tags = _read_entities(f, is_ascii, data_size)
         elif environ == "Nodes":
             points, point_tags = _read_nodes(f, is_ascii, data_size)
         elif environ == "Elements":
-            cells, cell_tags = _read_elements(
-                f, point_tags, physical_tags, is_ascii, data_size, field_data, cell_sets
+            cells, cell_tags, cell_sets = _read_elements(
+                f, point_tags, physical_tags, is_ascii, data_size, field_data
             )
         elif environ == "Periodic":
             periodic = _read_periodic(f, is_ascii, data_size)
@@ -177,7 +176,7 @@ def _read_nodes(f, is_ascii, data_size):
 
 
 def _read_elements(
-    f, point_tags, physical_tags, is_ascii, data_size, field_data, cell_sets
+    f, point_tags, physical_tags, is_ascii, data_size, field_data
 ):
     fromfile = partial(numpy.fromfile, sep=" " if is_ascii else "")
     c_size_t = _size_type(data_size)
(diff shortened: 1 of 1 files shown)


nschloe:
I'm getting
```
ImportError: No module named 'numpy'
```
on another PR as well. This is only for Python 3.5. No idea what's wrong here.

gdmcbain:
Hmm, no, I have no idea either.  Sorry.

Here I tried:
```shell
conda deactivate
conda create -n meshio-py35 python=3.5
conda activate meshio-py35
conda install pip
pip install -e ~/src/meshio
```
which brought in numpy-1.18.1
```shell
pip install pytest-xdist
pytest -n auto ~/src/meshio/test
```
That failed 32 because of h5py.
```shell
pip install h5py
pytest -n auto ~/src/meshio/test
```
which passed 436 and skipped 62.

gdmcbain:
Running the same script as in #536 in this Python 3.5.6 conda env gives:
```
Electrically Grounded : [2, 3]
Mechanically Clamped : [8, 9]
Mechanically Free : [0, 1, 4, 5, 6, 7]
Electrically Free : [7, 9]
Piezoelectricity : [8, 9, 10, 11, 12, 13, 14, 15]
Structural Mechanics : [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
Electrically Clamped : [4, 5]
```
which is the same as the result obtained there in Python 3.7.3 except for the `dict` not being ordered (which is acceptable and expected).

MEMAndersen:
How would I go about testing this new package. Can I simply download gdmcbains branch and import it like:
`import meshio.meshio`
 or do I need to install it in a virtual environment? I am interested in testing it for a 3d solid mesh like this one:

https://pastebin.com/XJH3q12K

This one has PhysicalVolumes overlapping, which the other example does not.

nschloe:
Okay, I've disabled the 3.5 test for now.

gdmcbain:
> How would I go about testing this new package. Can I simply download gdmcbains branch and import it like:
```python
import meshio.meshio
```
> or do I need to install it in a virtual environment?

This is no longer necessary since this PR has been merged into master, but for future reference see [‘Git branching—remote branches’](https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches).  

Something like:
```shell
git remote add gdmcbain git@github.com:gdmcbain/meshio
git fetch gdmcbain/536-msh41-multiple-tags  # Except that I'm about to delete it.
git checkout --track gdmcbain/536-msh-multiple-tags
```

Virtual environments aren't related to git so are irrelevant here; I mentioned the conda env above as a way of switching between versions of Python.

gdmcbain:
> I am interested in testing it for a 3d solid mesh like this one:

    https://pastebin.com/XJH3q12K

> This one has PhysicalVolumes overlapping, which the other example does not.

Thank you for the test case.  I think it looks O. K. here, using master after the merger of this PR; I find that `'clay'` consists of the first 75 tetra of `'domain'` and `'soil'` of the remaining 77.

MEMAndersen:
I have also run some tests now and find that everything works. I am a huge fan of the new cell_sets methodology! Thanks a bunch

Issue #299, opened 2018-09-03
read MEDIT mesh nodes and elements in blocks rather than linewise


gdmcbain:
The idea of removing large Python loops from i/o goes back to #29.  Here in question is the MEDIT mesh format, as flagged in the earlier PR reading and writing its tags #297.  This should work quite well for this format since each node and element has exactly one tag so that the data in each block is 'rectangular'. 

This might well make `mesh_file_type="mesh"` competitively fast for `pygmsh.generate_mesh` (nschloe/pygmsh#212).

This is separate from reading and writing binary MEDIT mesh #295 since Gmsh doesn't write binary MEDIT mesh even when given the `-bin` option.

Issue #295, opened 2018-08-31
binary MEDIT mesh


gdmcbain:
Thus far, `meshio.medit_io` only reads and writes text MEDIT mesh whereas according to Frey ([2001](https://hal.inria.fr/inria-00069921), §7.2.1):
> This format is composed of a single (binary or text) data file.

This would be of much more interest in the context of nschloe/pygmsh#212 if Gmsh wrote binary MEDIT mesh but it appears to ignore its `-bin` flag for `-format mesh`.  I'm not sure how common the binary form is in the wild.

nschloe:
Fixed via #782.

Issue #239, opened 2018-05-11
medit_io: Missing functionality for MEDIT .mesh files.


leoschwarz:
So a few things are still missing from the MEDIT .mesh reader and writer in the medit_io module:

- [ ] Constraint fields: `Corners`, `RequiredVertices`, `Ridges`, `RequiredEdges`. Even if they don't get stored they should be skipped gracefully (maybe with a warning) instead of failing the parsing.
- [ ] Normals and tangents fields: `Normals`, `Tangents`, `NormalAtVertices`, `NormalAtTriangleVertices`, `NormalAtQuadrilateralVertices`, `TangentAtEdges`. Same concern as above.
- [ ] Read and store label data. Complicated by the fact that only one-dimensional data seems to be supported. Which meshio cell and point data does this get written to, should we just assume `"label"` on read and whatever there is on write as long there is only one key?

nschloe:
With the new `mesh` container class, it should be a lot easier to cram in the extra data.

leoschwarz:
Thanks for letting me know. I hope I'll get to submitting a few patches in a few days (can't guarantee an ETA though), for my purposes I'm now simply ignoring unknown fields and pinting a warning when something like that is encountered (it's in my fork on GitHub already, if nothing else I'll send that in a PR).

My main question was how to normalize the data? If the goal of meshio is to make different data formats as interchangeable as possible, then fields should probably have the same names for all formats. (Or is your suggestion to implement the normalization in the mesh container class, maybe even with a lazy approach?)

nschloe:
> and pinting a warning when something like that is encountered 

Not sure how you would do that. Check every possible `Mesh` attribute for `None`? The lists of attributes may grow in the future, so you'd always have to update the warning checker.

> If the goal of meshio is to make different data formats as interchangeable as possible, then fields should probably have the same names for all formats.

Most mesh formats support the same stuff, e.g., points and cells, and whenever that's the case, they should use the same attribute (`Mesh.points`, `Mesh.cells` in this case). We're now going into directions where certain data are only supported by a few, or just one format (e.g., your `NormalAtTriangleVertices`). I'd say whenever you need to add a new attribute to `Mesh`, check if it's not already there under a different name. We should be fine then.

gdmcbain:
>  Read and store label data. Complicated by the fact that only one-dimensional data seems to be supported. Which meshio cell and point data does this get written to, should we just assume "label" on read and whatever there is on write as long there is only one key?

[...]

PR #782, opened 2020-04-20
Introduce binary format for MEDIT files

ChristosT:
This PR introduces support for binary medit files resolving thus #295 .
It is built based on the documentation provided at https://github.com/LoicMarechal/libMeshb which serves as a reference implementation of the MEDIT/GMF format.

Following the suggestion in #299 it reads and writes data in blocks .
Also, it skips fields gracefully as it was proposed in #239 .

The code could be adapted in order to replace the current used for ascii files, but I will be looking into this in a separate PR after I get your feedback about this .


ChristosT:
the test meshes weren't under git lfs, this should fix this

nschloe:
This looks _very_ nice. @gdmcbain Would you like to have a look as well?

meshio/medit/_medit.py
```diff
@@ -171,4 +331,148 @@ def write(filename, mesh, float_fmt=".15e"):
...
+
+        version = 3
+        itype = "i4"
+        postype = "i8"
+        ftype = "f8"
+        # according to manual keywords are always written as i4 independently of
+        # the file version
+        keytype = "i4"
+
+        # if we store internaly 64bit integers upgrade file version
```
gdmcbain, line 345: spell 'internally'

meshio/medit/_medit.py
```diff
@@ -6,6 +6,8 @@
 <https://www.ljll.math.upmc.fr/frey/publications/RT-0253.pdf>
 <https://www.math.u-bordeaux.fr/~dobrzyns/logiciels/RT-422/node58.html>
 for something like a specification.
```
gdmcbain, line 8: Are the old references rendered obsolete by libMeshb?  Should they be deleted?

gdmcbain:
It does look very nice, doesn't it!  I had a quick glance on Tuesday and quickly concluded that @ChristosT knows a lot more about MEDIT than I do!  (I hadn't even heard of GMF.)  It's a most welcome contribution.

I don't have ready access to my own old MEDIT cases at the moment, but I've stepped through some basic conversions in pdb; as it's also passing the suite, including its own good new additions, I'd recommend merging.

meshio/medit/_medit.py
ChristosT: Done. thank you

meshio/medit/_medit.py
ChristosT: Yes the old ones are just a subset.  So I will be keeping only the latest

ChristosT:
Thank you for your kind words and the review !

PR #785, opened 2020-04-21
Fix issue #783

### LINKED ISSUE #783: meshio.read INP with ELSET
Recently I received an INP file [FEMMeshGmsh.inp](https://github.com/kinnala/scikit-fem/files/4485287/FEMMeshGmsh.zip) which resists `meshio.read`.   I believe the file was generated in FreeCAD for Calculix.  I also believe that that latter uses the same input file format as Abaqus.  I don't know too much about the format myself or about Abaqus, but I have installed Calculix 2.10 under Microsoft Windows 10 and 2.16 under Ubuntu and the file is processed happily by each of those.  Despite the name of the file, I don't think that it has anything to do with Gmsh.

The error is:
```
  File "<stdin>", line 1, in <module>
  File "c:\users\gmcbain\src\meshio\meshio\_helpers.py", line 68, in read
    return reader_map[file_format](filename)
  File "c:\users\gmcbain\src\meshio\meshio\abaqus\_abaqus.py", line 99, in read
    out = read_buffer(f)
  File "c:\users\gmcbain\src\meshio\meshio\abaqus\_abaqus.py", line 144, in read_buffer
    set_ids, line = _read_set(f, params_map)
  File "c:\users\gmcbain\src\meshio\meshio\abaqus\_abaqus.py", line 292, in _read_set
    set_ids += [int(k) for k in line.strip().strip(",").split(",")]
  File "c:\users\gmcbain\src\meshio\meshio\abaqus\_abaqus.py", line 292, in <listcomp>
    set_ids += [int(k) for k in line.strip().strip(",").split(",")]
ValueError: invalid literal for int() with base 10: 'Evolumes'
```

I believe the resistant line of FEMMeshGmsh.inp is line 540, the second of: 
```
** Volume elements
*Element, TYPE=C3D10, ELSET=Evolumes
```

This may duplicate an existing issue on INP; I haven't been following developments on this format too closely.

keurfonluu:
- Fixed: handle named element sets (fixes #783).

### DIFF
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -135,22 +135,31 @@ def read_buffer(f):
                 cell_sets_element_order += list(sets.keys())
         elif keyword == "NSET":
             params_map = get_param_map(line, required_keys=["NSET"])
-            set_ids, line = _read_set(f, params_map)
+            set_ids, _, line = _read_set(f, params_map)
             name = params_map["NSET"]
             point_sets[name] = numpy.array(
                 [point_ids[point_id] for point_id in set_ids], dtype="int32"
             )
         elif keyword == "ELSET":
             params_map = get_param_map(line, required_keys=["ELSET"])
-            set_ids, line = _read_set(f, params_map)
+            set_ids, set_names, line = _read_set(f, params_map)
             name = params_map["ELSET"]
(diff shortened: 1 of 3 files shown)


nschloe:
Thanks, Keurfon!

Issue #789, opened 2020-04-24
Pruning lower dimensional entities causes issues with cell_data


uvilla:
@nschloe ,
Thank you again for this great tool!

I found a small glitch in how `cell_data` is handled within `mesh.prune()`

Consider the following examples (which uses `pygalmesh`).

```python
import pygalmesh
import meshio
import numpy as np


if __name__ == '__main__':

    # Create a simple "brick" domain with 3 stripes
    data = np.zeros((100,100,100), dtype=np.uint8)
    data[:,20:40,20:80] = 1
    data[:,40:60,20:80] = 2
    data[:,60:80,20:80] = 3
    h    = [.01, .01, .01]
        
    mesh = pygalmesh.generate_from_array(data, h, facet_distance=.01, cell_size=.1, verbose=True)    
    

    meshio.xdmf.write("before_pruning.xdmf", mesh)
    
    mesh.prune()
    meshio.xdmf.write("after_pruning.xdmf", mesh)
```

The expected output of this code is a brick mesh with three layers corresponding to material ids 1,2,3. 

The figure below shows that `cell_data` is written correctly if the mesh is not pruned (left).
However, if we prune the mesh first, then the written `cell_data` is jitter (right).

![Image 4-24-20 at 2 37 PM](https://user-images.githubusercontent.com/17551415/80250594-c5528980-8639-11ea-9076-ec93fc625c35.jpg)

uvilla:
Related to the above question,

Can you explain what is the content of `cell_data`?

In particular, why does
```python
print(mesh.cell_data)
```

Shows that the `medit:ref` data is a list of two numpy arrays?

```bash
{'medit:ref': [array([3, 3, 2, ..., 1, 1, 1]), array([2, 3, 3, ..., 1, 1, 1])]}
```

I assume that one array is the cell labels. What is the meaning of the other one?

nschloe:
> Shows that the medit:ref data is a list of two numpy arrays?

Because `mesh.cells` is a list of two cellblocks.

nschloe:
What is going wrong exactly with the prune? One cannot tell from the screenshot alone. Perhaps the triangles are removed?

uvilla:
In `cgal` when creating a mesh from a domain with multiple labels each cell inherits a label corresponding to the subdomain it belonged to. Points are assigned a label which depends on the labels of the cells that are incident to the points.

Therefore in the figure of the left, you can see that each tetrahedron has the correct label (you can see the three layers).

[...]

PR #790, opened 2020-04-24
Pruning cell_data in mesh.prune()

uvilla:
Resolve issue #789

### DIFF
--- a/meshio/_mesh.py
+++ b/meshio/_mesh.py
@@ -74,13 +74,13 @@ class Mesh:
 
         new_cells = []
         new_cell_data = {}
-        for c in self.cells:
+        for block_counter, c in self.cells:
             if c.type not in prune_list:
                 new_cells.append(c)
                 for name, data in self.cell_data.items():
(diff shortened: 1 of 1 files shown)


meshio/_mesh.py
```diff
@@ -74,13 +74,13 @@ def prune(self):
 
         new_cells = []
         new_cell_data = {}
-        for c in self.cells:
+        for block_counter, c in enumerate(self.cells):
```
nschloe, line 77: Better `zip`  it, not `enumerate`.

uvilla: Do you mean

```python
for block_counter, c in zip(range(len(self.cells)), self.cells):
```

It seems to me too verbose

uvilla: And I can't either `zip(self.cells, self.cell_data)`, since `self.cell_data` is a Dictionary.

nschloe: You're right. Let's merge this.

PR #807, opened 2020-05-13
Remove cdata workaround

### LINKED ISSUE #806: Meshio seems to modify ElementTree, which breaks svgwrite (an external package)
Hi There,

I have a rather puzzling bug in my testharness (https://travis-ci.org/github/tmolteno/disko/jobs/686063990) which can be fixed by uninstalling meshio.  My code (https://github.com/tmolteno/disko)  uses dmsh for mesh generation which indirectly includes meshio.

It appears that, in meshio/_common.py,  the built-in ElementTree _serialize_xml method is substituted around line 181...

```
ET._original_serialize_xml = ET._serialize_xml


def _serialize_xml(write, elem, qnames, namespaces, short_empty_elements, **kwargs):
    if elem.tag == "![CDATA[":
        write("\n<{}{}]]>\n".format(elem.tag, elem.text))
        if elem.tail:
            write(ET._escape_cdata(elem.tail))
    else:
        return ET._original_serialize_xml(
            write, elem, qnames, namespaces, short_empty_elements, **kwargs   ## Here's the error
        )


ET._serialize_xml = ET._serialize["xml"] = _serialize_xml    # Here's the substitution
```

This causes an error when svgwrite tries to use elementTree to serialize XML:

```
TypeError: _serialize_xml_with_CDATA_support() takes 4 positional arguments but 5 were given

../../../virtualenv/python3.6.7/lib/python3.6/site-packages/meshio/_common.py:177: TypeError
```

I'm a little unsure of what is going on here, (a travis-ci log is available at https://travis-ci.org/github/tmolteno/disko/jobs/686063990).  It appears that any software that imports meshio (even indirectly and never calling its methods) will result in a modified elementTree that can trigger this error when it expected to call the original elementTree code for serializing XML. 

This happens in python 3.6, 3.7 and 3.8. Is there anything I can do to avoid this? I'm not even sure meshio is doing anything wrong here. Is it calling the original serialize_xml with the wrong number of arguments?

Kind Regards
Tim

nschloe:
Fixes #806.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E203, E266, E501, W503, C901
+ignore = E203, E266, E501, W503, C901, E741
 max-line-length = 88
 max-complexity = 18
 select = B,C,E,F,W,T4,B9
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ tag:
 	# Always create a github "release" right after tagging so it appears on zenodo
 	curl -H "Authorization: token `cat $(HOME)/.github-access-token`" -d '{"tag_name": "v$(VERSION)"}' https://api.github.com/repos/nschloe/meshio/releases
 
-upload: clean setup.py
+upload: clean
 	# Make sure we're on the master branch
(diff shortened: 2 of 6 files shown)


nschloe:
Fixes #806.

PR #811, opened 2020-05-19
Support for VTU polygon and polyhedron

keileg:
## Main contribution
Added support for reading and writing VTU polygon and polyhedron cells.

## Comments 

- The number of nodes is appended to the meshio cell type, similar to what is done for the vtk format
- Conversion to and from other formats than .vtu and internal meshio has not been tested
- Also added a new polyhedron mesh to the test.helpers
- Tests cover all cases I could think of, but please give feedback if more cases should be considered


nschloe:
LTGM, thanks!

PR #813, opened 2020-05-21
enable import of raw binary VTU

### LINKED ISSUE #483: Does meshio read binary vtu files?
Does meshio read binary vtu files? 
Trying to read the vtu files posted here: [https://github.com/topology-tool-kit/ttk-data](https://github.com/topology-tool-kit/ttk-data)
I get the same error for all:
```
XMLSyntaxError: Input is not proper UTF-8, indicate encoding !
Bytes: 0xC9 0x0C 0x00 0x00, line 28, column 6 (dragon.vtu, line 28)

During handling of the above exception, another exception occurred:
```
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc9 in position 1365: invalid continuation byte
```
I read the file as follows:

fdata = meshio.read("dragon.vtu")
For any other file at the above link the same errors are displayed, the only difference is the line number.

I doubt that all vtu files in that repo have the same errors. Is something I missing when reading the files?

c-abird:
This fixes #483 for uncompressed raw binary VTUs as produced e.g. by Firedrake.


nschloe:
We'll probably need a test for this. Could you edit `.gitattributes` to include vtu files, and then add one to `test/meshes/`?

nschloe:
Also, the comment about binary VTU in the readme can go.

nschloe:
LGTM, thanks! Frankly I never though we'd have this in meshio. Cool!

c-abird:
Seems that the XML parser in Python 3.5 throws a TypeError instead of a ParseError in case of raw data. Should I open another pull request to fix this?

nschloe:
If you know how to fix it, great. I wouldn't spend too much time on it though, Python 3.5 goes out of support this September.

PR #814, opened 2020-05-21
Big endian swap

waveform80:
The Medit tests (specifically the binary read tests) are currently failing on the S/390 (a big-endian platform) because the byte-swapping code introduced in 6c430e08e22940c87e0a313446a5e82384ed7842 (and modified in ecf8ec742228f5559bd7182e264845d5e6c540c4) constructs an invalid numpy typecode (the "S" prefix, presumably intended to imply byte-swapping, doesn't exist and was erroneously added as a suffix anyway for `keytype`).

As numpy (and struct) provide no byte-swap prefix, this patch uses a crude struct test to determine the native platform's endianness and applies the opposite to the constructed type-codes. The patch has been tested on an S/390 successfully, and adds a copy of the `hch_strct.4.meshb` file to the test suite to exercise the byte-swapping code on little-endian platforms.


nschloe:
Cool, thanks!

nschloe:
Do the other tests work on big endian? We've had problems with VTU before, see https://github.com/nschloe/pygalmesh/issues/48#issuecomment-541411579.

waveform80:
Yes, all other tests passed, although the version under test was the current one in Ubuntu (4.0.10) so I can't necessarily vouch for the status of master at this time.

nschloe:
Can you try reading the linked mesh file? https://github.com/nschloe/pygalmesh/blob/master/test/meshes/elephant.vtu
```python
import meshio
meshio.read("elephant.vtu")
```

mclemenceau:
Hey Nico,

I was able to read elephant.vtu on my s390x without issue!

However, I'm having an issue with test_vtk.py
When using numpy.sum on rbc_001.vtk I get very different results on amd64 versus s390x which makes the assert abs(s - ref_sum) < tol * ref_sum fail on s390x.

Let me know if you want me to formally report this or if you have an idea on what this could be,

Thanks!
Matthieu

nschloe:
@mclemenceau Thanks for checking! Does this mean the tests are failing on your machine?

mclemenceau:
Yes,

I'm trying to help python-meshio 4.0.12 to migrate to ubuntu groovy but test_vtk.py fails against numpy 1.18.4.

Matthieu

mclemenceau:
And to be clear it only fails on 390x architecture.

waveform80:
@nschloe I should clarify; @mclemenceau is my colleague who originally brought the medit issue to my attention. When I reported that (with the patch) it passed all tests on the current Ubuntu version it turns out that was only the case with the version of numpy in the release pocket (1.17.4). When the patched version was re-tested using the version of numpy in the proposed pocket (1.18.4) the issue with test_vtk.py surfaced.

It's likely that something's changed in numpy's sum (or add) implementation between the two versions, but we're not entirely sure whether the result is a bug in numpy (producing the wrong result on ppc) or meshio (requiring a finer tolerance than necessary with this particular test).

However, I would suggest that the vtk test failure is probably a separate issue (unrelated to the medit one here) and probably ought to be tracked as such.

nschloe:
I'd _love_ to fix this, but I don't have access to a big-endian machine. If you know any place where I can get access, let me know!

mclemenceau:
Thanks @waveform80, yes I should have introduced myself :P. Thanks!

@nschloe, I will open a specific issue. 
Now, I'm not sure how to get you access to a Big-endian machine at this time but I can run debugging by proxy if you want. I'd be happy to run test for you and send you back outputs.

Matthieu

mclemenceau:
Issue open https://github.com/nschloe/meshio/issues/824

nschloe:
> I'd be happy to run test for you and send you back outputs.

Thanks, but that won't be efficient. I'll keep looking.

waveform80:
> > I'd be happy to run test for you and send you back outputs.
> 
> Thanks, but that won't be efficient. I'll keep looking.

One possibility that springs to mind is that Travis recently added facilities for running tests on multiple architectures, including s390x: https://docs.travis-ci.com/user/multi-cpu-architectures

nschloe:
Do they offer a login?

waveform80:
> Do they offer a login?

No, strictly a batched interface by the looks of it; painful, but still possible to use for debugging purposes (this is something I have to do occasionally :). For interactive usage the MiniCloud at https://openpower.ic.unicamp.br/minicloud/ looks like it might be useful. There are a couple of others mentioned at https://developer.ibm.com/linuxonpower/cloud-resources/ but the sheer amount of information / legalese each demanded put me off experimenting!

(side note: I'm rather surprised IBM's own cloud doesn't just offer these as a simple paid option - I'd happily shell out a couple of bucks to test stuff on a power box every now and then; sadly the only "esoteric" arch on clouds at the moment seems to be arm64)

mclemenceau:
Another option would be to use a Debian porter box
https://wiki.debian.org/PorterBoxHowToUse
I'm not familiar with the process personally but it allows developer to
request access to those porter boxes
This should give real login access to any architecture necessary.

Matthieu

nschloe:
Yeah I applied for this but the process takes very long. Let's just wait then.

PR #817, opened 2020-05-22
Enable reading compressed raw VTU files

### LINKED ISSUE #793: Could not read VTU file with binary data when header is encoded separately
I was playing around with vtu meshes with different appended data formats, and it seems, that meshio assumes the header to be encoded with the following data (uncompressed, base64) in the appended section (maybe even in data-array sections). So I couldn't load the following mesh:
[example](http://www.earthmodels.org/software/vtk-and-paraview/vtk-file-formats/vox8_binary_appended.vtu)

Isn't that file written correctly? Paraview was able to read it.

c-abird:
Morever, this PR fixes an issue of GH-813 with UInt64 encoding and unaligned array data to mixed types.

Should also fix #793 (though I'm not sure since I'm still not able to read the example mesh, see comment in #793)


c-abird:
Thanks for the hint! I found the issue with Python 3.5 though (open cannot handle Path objects directly).

nschloe:
Very good.

PR #825, opened 2020-05-28
Support more polygonal facets for ply format

bwoodsend:
As the title suggests - support the other cell types besides `"triangle"` and `"quad"`. 

Could  you check [these lines](https://github.com/nschloe/meshio/compare/master...bwoodsend:master#diff-af0fc2872f0e1a5abc58931fc6b17b5dR39) and confirm if that is the correct behaviour. I've got it supporting all the polygonal cell types I could find and none of the polyhedron types.  

**Unknown cell types**
On writing, it skips cell types not listed with a warning.
On reading it silently invents a `polygon[count]` cell type if it doesn't know an appropriate type.

### DIFF
--- a/meshio/ply/_ply.py
+++ b/meshio/ply/_ply.py
@@ -36,16 +36,19 @@ ply_to_numpy_dtype = {
 numpy_to_ply_dtype = {numpy.dtype(v): k for k, v in ply_to_numpy_dtype.items()}
 
 
-cell_type_to_count = {
-    "line": 2,
-    "triangle": 3,
-    "quad": 4,
-    "polygon5": 5,
-    "polygon6": 6,
-    "quad8": 8,
-}
+_cell_type_to_count = {"vertex": 1, "line": 2, "triangle": 3, "quad": 4}
 
-_cell_type_from_count = {val: key for (key, val) in cell_type_to_count.items()}
+
+def cell_type_to_count(cell_type):
+    if cell_type in _cell_type_to_count:
+        return _cell_type_to_count[cell_type]
+
+    match = re.fullmatch(r"polygon(\d+)", cell_type)
+    if match:
+        return int(match.group(1))
+
+
+_cell_type_from_count = {val: key for (key, val) in _cell_type_to_count.items()}
 
 
 def cell_type_from_count(count):
@@ -406,7 +409,7 @@ def write(filename, mesh, binary=True):  # noqa: C901
 
         num_cells = 0
         for cell_type, c in mesh.cells:
-            if cell_type in cell_type_to_count.keys():
+            if cell_type_to_count(cell_type):
                 num_cells += c.data.shape[0]
         fh.write("element face {:d}\n".format(num_cells).encode("utf-8"))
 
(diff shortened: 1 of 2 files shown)


meshio/ply/_ply.py
```diff
@@ -35,6 +36,25 @@
...
+    "polygon5": 5,
+    "polygon6": 6,
+    "triangle6": 6,
+    "quad8": 8,
+}
+
+_cell_type_from_count = {val: key for (key, val) in cell_type_to_count.items()}
+
+
+def cell_type_from_count(count):
```
nschloe, line 52: This isn't unique, so we won't be able to convert some meshes forth and back without change.

bwoodsend: Yeah, is there any difference between a `polygon6` and a `triangle6`?

bwoodsend: Yeah, is there any difference between a `polygon6` and a `triangle6`?

nschloe: Yes. A `polygon6` is a polygon with 6 corners, a `triangle6` is a "quadratic" triangle (as used in finite elements, for example). It's geometrically the same, but some formats discriminate between the two, so meshio should, too. Do you know what the PLY specification says? A link in the code would be helpful.

bwoodsend: OK, sounds like we want `polygon6` and not `triangle6`. 

From the overview section of the [link at the top of that script](https://web.archive.org/web/20161221115231/http://www.cs.virginia.edu/~gfx/Courses/2001/Advanced.spring.01/plylib/Ply.txt) it looks like PLY isn't meant to hold that kind of information but I can't find anything explicit either way. Regardless it looks like `polygon6` is far more common so I think I'll just change it to that.

meshio/ply/_ply.py
```diff
@@ -35,6 +36,24 @@
 numpy_to_ply_dtype = {numpy.dtype(v): k for k, v in ply_to_numpy_dtype.items()}
 
 
+cell_type_to_count = {
+    "line": 2,
+    "triangle": 3,
+    "quad": 4,
+    "polygon5": 5,
+    "polygon6": 6,
+    "quad8": 8,
```
nschloe, line 45: `quad8` should also be removed.

nschloe: It seems that PLY supports and arbitrary number of nodes in a facet. We can try and support that. line, triangle, quad map to 2, 3, 4 (how about vertex for 1?), everything above to `polygonN` (where `N` is the respective number).

bwoodsend: Sure, I'll do that.

test/test_ply.py
```diff
@@ -15,6 +15,8 @@
         helpers.tri_quad_mesh,
         helpers.add_point_data(helpers.tri_mesh, 1, dtype=int),
         helpers.add_point_data(helpers.tri_mesh, 1, dtype=float),
+        helpers.line_mesh,
+        helpers.quad8_mesh,
```
nschloe, line 19: quad8 needs to go.

nschloe:
LGTM, thanks!

PR #828, opened 2020-05-30
Fix for vtu scalar DataArray with NumberOfComponents

michejansen:
Hi,

I'm facing a small issue with some vtu files created by a third-party software. Contrary to Paraview, this program adds the "NumberOfComponents" tag to all DataArray; even for scalar data "NumberOfComponents=1" is added. Meshio fails to read these files since it creates 2 dimensional numpy arrays for scalar data, while 1 dimensional arrays are expected further on the code. This leads to errors like the following:
```console
  File "/home/miche/workspace/meshio/meshio/vtu/_vtu.py", line 48, in _cells_from_data
    meshio_type = vtk_to_meshio_type[types[start]]
TypeError: unhashable type: 'numpy.ndarray'
```
In vtu files exported by Paraview, the "NumberOfComponents" tag is only added in case of vector or tensor data. Nevertheless, Paraview can read these third-party files without issue and it seems this shouldn't cause a failure in meshio either. This pull request fixes this small issue.

### DIFF
--- a/meshio/vtu/_vtu.py
+++ b/meshio/vtu/_vtu.py
@@ -491,7 +491,7 @@ class VtuReader:
         else:
             raise ReadError("Unknown data format '{}'.".format(fmt))
 
-        if "NumberOfComponents" in c.attrib:
+        if "NumberOfComponents" in c.attrib and int(c.attrib["NumberOfComponents"]) > 1:
             data = data.reshape(-1, int(c.attrib["NumberOfComponents"]))
         return data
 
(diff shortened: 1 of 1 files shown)


nschloe:
This looks like a fix at the wrong place. meshio should be able to handle arrays of shape (n, 1).

michejansen:
OK, thanks for the feedback. Indeed, it's only the input of the _cells_from_data function (connectivity, offsets, types) that is expected to be one-dimensional. Regular point and cell data work fine as 2d arrays. The fix is now limited to the _cells_from_data function call.

nschloe:
LGTM, thanks!

PR #830, opened 2020-06-04
tostring() is renamed to tobytes() for clarity.

tkoyama010:
Hi all. I renamed the method tostring to tobytes. Method tostring will be removed in Python 3.9.
https://docs.python.org/3/library/array.html#array.array.tobytes
Thank you for reading.


nschloe:
Nice, thanks.

Issue #831, opened 2020-06-05
Error converting VTK to VTU


keurfonluu:
**Platform**: Windows 10 (1909)
**Python**: 3.7.7
**meshio**: 4.0.14
**numpy**: 1.18.1

When reading a VTK file (written by `meshio`) and converting it to VTU, it raises a `KeyError: dtype('>i4')` as the data type of cells connectivity is `dtype('>i4')` and not `dtype('int32')`.

Code to reproduce the error:

```python
import meshio
import numpy

points = numpy.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]])
cells = [("triangle", numpy.array([[0, 1, 2]]))]
meshio.write_points_cells("test.vtk", points, cells)
mesh = meshio.read("test.vtk")
print(mesh.cells[0].data.dtype)  # dtype('>i4')
mesh.write("test.vtu")  # KeyError: dtype('>i4')
```

nschloe:
Ah, big-endian. There are some bugs for this architecture-type. I'd love to get my hands ona be machine for debugging.

nschloe:
Fixed by #832.

PR #832, opened 2020-06-05
Fix issue #831

keurfonluu:
- Fixed: convert cell data type to `int32` in VTK binary reader (fix issue #831).

### DIFF
--- a/meshio/vtk/_vtk.py
+++ b/meshio/vtk/_vtk.py
@@ -390,7 +390,7 @@ def _read_cells(f, is_ascii, num_items):
     if is_ascii:
         c = numpy.fromfile(f, count=num_items, sep=" ", dtype=int)
     else:
-        c = numpy.fromfile(f, count=num_items, dtype=">i4").astype(int)
+        c = numpy.fromfile(f, count=num_items, dtype=">i4")
         line = f.readline().decode("utf-8")
         if line != "\n":
             raise ReadError()
@@ -402,7 +402,7 @@ def _read_cell_types(f, is_ascii, num_items):
         ct = numpy.fromfile(f, count=int(num_items), sep=" ", dtype=int)
     else:
         # binary
-        ct = numpy.fromfile(f, count=int(num_items), dtype=">i4").astype(int)
+        ct = numpy.fromfile(f, count=int(num_items), dtype=">i4")
         line = f.readline().decode("utf-8")
         # Sometimes, there's no newline at the end
         if line.strip() != "":
--- a/meshio/vtk/_vtk.py
+++ b/meshio/vtk/_vtk.py
@@ -390,7 +390,7 @@ def _read_cells(f, is_ascii, num_items):
     if is_ascii:
         c = numpy.fromfile(f, count=num_items, sep=" ", dtype=int)
     else:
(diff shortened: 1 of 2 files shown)


nschloe:
This does a copy, we'd rather avoid it. Why the change?

keurfonluu:
`meshio`'s VTK format is forcing big endianness: https://github.com/nschloe/meshio/blob/ad906b4c2aa64635adfb3c076a0370ac8a4ff1ed/meshio/vtk/_vtk.py#L364).

The writer is storing data in big endian order:
https://github.com/nschloe/meshio/blob/ad906b4c2aa64635adfb3c076a0370ac8a4ff1ed/meshio/vtk/_vtk.py#L668

The reader is reading in big endian order:
https://github.com/nschloe/meshio/blob/ad906b4c2aa64635adfb3c076a0370ac8a4ff1ed/meshio/vtk/_vtk.py#L393

This causes cell connectivity data type to be `dtype('>i4')` which key does not exist in `numpy_to_vtu_type`

https://github.com/nschloe/meshio/blob/ad906b4c2aa64635adfb3c076a0370ac8a4ff1ed/meshio/vtu/_vtu.py#L230

thus raising the `KeyError: dtype('>i4')` observed in issue #831.

nschloe:
Seems like we'd rather want to adapt the dictionary.

nschloe:
I see
```python
# Use the native endianness. Not strictly necessary, but this simplifies things
# a bit.
```
We might want to change that to the endianness of the input data.

keurfonluu:
After further investigations, I found it weird that `meshio.Mesh.points` with data type `dtype('>f8')` was successfully converted. It appears that the byte order of the attributes of a `meshio.Mesh` were all swapped except for `cells` in VTU writer.

nschloe:
Not sure if we should do that at all. How about using the byteorder of the data, not system byte order? (as mentioned above)

keurfonluu:
I tried using the data byteorder before coming up with the latter workaround. Didn't manage to produce a valid VTU file.

What I did:
1. Define byteorder from data if not native, otherwise from system,

```python
byte_order = mesh.points.dtype.byteorder
byte_order = (
    "LittleEndian" if sys.byteorder == "little" else "BigEndian"
    if byte_order == "="
    else "LittleEndian" if byte_order == "<" else "BigEndian"
)
```

2. Remove codes related to byte swapping,

3. Appended new keys to `numpy_to_vtu_type`:

```python
numpy_to_vtu_type[numpy.dtype(">i4")] = "Int32"
numpy_to_vtu_type[numpy.dtype(">f8")] = "Float64"
```

With these few modifications, ParaView fails to read the points (and probably the rest).

nschloe:
`sys.byteorder` should now appear, only the data byte order.

keurfonluu:
`dtype.byteorder` returns `"="` (native) for all arrays created by `numpy` unless explicitly stated in `dtype`.

keurfonluu:
Another problem raised by determining byteorder from data is that since byteorder is a global attribute for VTU format, we have to assume that all the data in a mesh has the same byteorder if we want to avoid any conversion, which might not always be the case.

nschloe:
Alright, let's do it like you suggest.

PR #858, opened 2020-07-01
fix for point_offsets when cell blocks are not the same length

chrishavlin:
The calculation of `point_offsets` in the `vtu` reader assumed that and cell blocks have the same number of elements.  This PR adjusts the calculation to allow variable cell block size.

For my `.vtu` file,  I have 8 cell blocks, and the 8th block has more elements: 

```
print([pts.shape[0] for pts in points])
[17472, 17472, 17472, 17472, 17472, 17472, 17472, 17536]
``` 

The original method results in the following `point_offsets` : 

``` 
[     0  17472  34944  52416  69888  87360 104832 122368]
```
The final element should be `104832+17472=122304` to reference the start of the 8th cell block. This results in an error when trying to load and evaluate the coordinates of the vertices in my cell blocks: 

```
import meshio
sP=meshio.read('solution-00005.0000.vtu')
Coords = sP.points[sP.cells[-1].data.ravel()]
```
throws `IndexError: index 139840 is out of bounds for axis 0 with size 139840` as the connectivity array ends up with values that exceed the length of the points array. The data I'm using is available here: https://osf.io/56axf/ any of the processor 0 `.vtu` files will throw the error. 
 
The PR adjusts the calculation to account for the variable cell block size. 

Results of my local `pytest` are attached. 
[testoutput.txt](https://github.com/nschloe/meshio/files/4860033/testoutput.txt)


meshio/vtu/_vtu.py
```diff
@@ -363,9 +363,8 @@ def __init__(self, filename):  # noqa: C901
         if len(cell_data_raw) != len(cells):
             raise ReadError()
 
-        point_offsets = (
-            numpy.cumsum([pts.shape[0] for pts in points]) - points[0].shape[0]
-        )
+        point_offsets = [0]
+        point_offsets.extend(numpy.cumsum([pts.shape[0] for pts in points])[:-1])
```
nschloe, line 367: Let's make this a one-liner:
```python
 point_offsets = [0] + numpy.cumsum...
```

chrishavlin:
Oh, sorry, indentation got screwed up... one more commit coming.

chrishavlin:
sorry again, forgot  whitespace around the `+`, I'll squash all these commits after making sure the tests pass.

nschloe:
Thanks for the nice fix!

PR #860, opened 2020-07-03
Gmsh: Fix v2.2 reader producing bad cell data arrays

dalcinl:
This bug shows up if all cells of the same type are not listed consecutively in the file. This reader generates a new cell block each time there is a cell type change in the file. Therefore, the cell data array lengths should agree with the number of cells on each cell block.

This bugs do not normally trigger with files generated from the Gmsh app. However, other mesh generators like Pointwise, when exporting to v2.2 format, produce files where cells of a given type are not all packed together in the file. The file is still valid and follows the documented format, but then meshio's reader produces cell data arrays with values for all cells of a given type rather than for the cells in the corresponding cell block.


nschloe:
Thanks!

mparsani:
Thanks!

PR #838, opened 2020-06-09
fix ordering hex20 and fixed support wedge15

ffilotto:

PR #863, opened 2020-07-11
Fix node ordering of the vtkWedge element

michejansen:
Wedge6 elements are currently converted as-is between gmsh and vtk/vtu formats. However, the gmsh Prism element and the vtkWedge have in fact a different node ordering; compare http://gmsh.info/doc/texinfo/gmsh.html#Node-ordering and https://lorensen.github.io/VTKExamples/site/VTKFileFormats/: the normal of the (0,1,2) triangle in the gmsh element points inwards, while this normal points outwards in the vtkWedge.
As a result, converting these elements from gmsh format to vtk or vtu format leads to elements with a negative volume in Paraview. This can be checked by running the MWE in attachment ([wedge6.txt](https://github.com/nschloe/meshio/files/4907066/wedge6.txt)) to create a wedge element in gmsh, followed by:
`gmsh wedge6.txt`
`meshio-convert wedge6.msh wedge6.vtk`
Inspecting the element's volume in Paraview (using the Cell Size filter) shows that the element has a negative volume.
While meshio mostly follows the vtk node ordering, I think it would be better to use the gmsh ordering in this case, since this one seems to be more conventional: the gmsh convention is used in most other formats (Ansys, Nastran, ...). In fact, the vtkWedge also uses a different ordering than the vtkHigherOrderWedge; see e.g.:
https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6821
https://vtk.org/doc/nightly/html/classvtkHigherOrderWedge.html

This PR therefore proposes a fix which converts the vtkWedge ordering while reading/writing vtk and vtu files.

(Sorry for creating this PR twice)

### DIFF
--- a/meshio/_common.py
+++ b/meshio/_common.py
@@ -232,6 +232,12 @@ meshio_to_vtk_type = {v: k for k, v in vtk_to_meshio_type.items()}
 
 
 def _vtk_to_meshio_order(vtk_type, numnodes, dtype=int):
+    # meshio uses the same node ordering as VTK for most cell types. However, for the
+    # linear wedge, the ordering of the gmsh Prism [1] is adopted since this is found in
+    # most codes (Abaqus, Ansys, Nastran,...). In the vtkWedge [2], the normal of the
+    # (0,1,2) triangle points outwards, while in gmsh this normal points inwards.
+    # [1] http://gmsh.info/doc/texinfo/gmsh.html#Node-ordering
+    # [2] https://vtk.org/doc/nightly/html/classvtkWedge.html
     if vtk_type == 13:
         return numpy.array([0, 2, 1, 3, 5, 4], dtype=dtype)
     else:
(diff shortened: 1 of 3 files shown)


michejansen:
The isort check failed on files that were not modified in this PR. It seems that the version of isort has recently changed from 4.3.21 to 5.0.3 in the automated checks?

nschloe:
I think this is a reasonable change, but it needs more documentation. Please put a sentence and a link or two at the conversion functions.

michejansen:
Thanks for the feedback. Some documentation has been added. Seeing that there are some other issues related to the node ordering (https://github.com/nschloe/meshio/pull/838), maybe it would be useful if the ordering used by meshio for the different cell types was documented somewhere?

nschloe:
> maybe it would be useful if the ordering used by meshio for the different cell types was documented somewhere?

That'd be _really_ good. Perhaps an md file with some ASCII art?

nschloe:
First of all, let's merge this. Thanks for the nice contribution.

michejansen:
> > maybe it would be useful if the ordering used by meshio for the different cell types was documented somewhere?
> 
> That'd be _really_ good. Perhaps an md file with some ASCII art?

Yes, some ASCII art would be great. I'm not an ASCII artist, but I could look into it.

nschloe:
> Yes, some ASCII art would be great. I'm not an ASCII artist, but I could look into it.

That'd be A+.

PR #867, opened 2020-07-19
Add .su2 support

### LINKED ISSUE #801: SU2 mesh format support
SU2 is a popular open source CFD code with its own mesh format ([https://su2code.github.io/docs_v7/Mesh-File/](https://su2code.github.io/docs_v7/Mesh-File/)). GMSH itself can already output .su2 mesh files but it seems like neither pygmsh nor meshio can do so at the moment. Or am I missing something?

ChristosT:
This PR introduces initial support for the su2 file format used by the SU2 CFD solver and decribed here https://su2code.github.io/docs_v7/Mesh-File/.

The su2 file format is an ascii based format and has some similarities with the vtk legacy ascii format.

The new code parses only coordinates , element connectvity and boundary tags and skips all other simulation info that can be found in the meshes.

SU2 files use mainly strings to tags elements . I converted them to ints with the appropriate warnings, let me know if you have a better idea how to support them while not breaking other meshio-writers.

Also it does not support multizone descriptions (see https://su2code.github.io/docs_v7/Multizone/) this could be  something for the future.

I tried it on several meshes from the [Test Cases](https://github.com/su2code/TestCases) repository of SU2 and it works well.
If however you find any issues let me know !

This also resolves #801


nschloe:
Very good. Ping me when the tests are passing or you need any help.

ChristosT:
@nschloe this should be good to go !

nschloe:
LGTM, thanks!

PR #870, opened 2020-07-23
Add a graphical overview of the different cell types

michejansen:
As proposed in #863, I prepared an overview of the node ordering for the different cell types. In the end, I decided to use tikz instead of ascii art, because that was a lot simpler for me. This means the result is an image in png, svg, pdf ... format. Hopefully, this can be used as well? Otherwise, no problem, I can find some other use case for these images anyway. Running pdflatex on the file [cell_types.tex](https://github.com/michejansen/meshio/blob/doc_cell_types/doc/cell_types.tex) generates a single svg image containing most first- and second-order elements (pdf version: [cell_types.pdf](https://github.com/nschloe/meshio/files/4967946/cell_types.pdf)):
`pdflatex -shell-escape cell_types.tex`
It is also possible to generate a separate image per cell type by uncommenting the tikzexternalize command.

The actual numbering of the nodes is mainly adopted from vtk and occasionally gmsh. The tex file contains the source in comments in case there are several options. Some other remarks:
* the orderings of the pyramid13, hexadron20 (#838), hexadron27 and wedge15 (#838) elements differ from gmsh, so this should still be added in _gmsh_to_meshio_order. 
* pyramid13 uses the vtk ordering, while pyramid14 uses gmsh ordering, because this element does not seem to exist in vtk. This might seem a bit weird.

### DIFF
--- a/doc/cell_types.tex
+++ b/doc/cell_types.tex
@@ -18,7 +18,7 @@
 
 % cnode(tag,x,y,z,label,label_pos)
 \def\cnode(#1,#2,#3,#4,#5,#6){
-\node (#1) at (#2,#3,#4) [vertex,label=#6:$#5$] {};
+\node (#1) at (#2,#3,#4) [vertex,label=#6:$\mathsf{#5}$] {};
 }
 
 \pagestyle{empty}
--- a/doc/cell_types.tex
+++ b/doc/cell_types.tex
@@ -1,0 +1,414 @@
+\documentclass[convert=pdf2svg]{standalone}
+% \documentclass{article}
+
+\usepackage[T1]{fontenc}
+\usepackage{lmodern}
+\renewcommand{\familydefault}{\sfdefault}
+\usepackage{tikz}
+\usepackage{tikz-3dplot}
+\usetikzlibrary{external}
+\tikzset{external/force remake}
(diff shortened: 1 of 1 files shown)


nschloe:
This looks really nice. My first thought was that nobody looks at this (because it's not in code), but we'll find a place to put it. Perhaps the wiki?

Anyways, could you use sans font throughout? (It's easier to read for things like this.)

michejansen:
> This looks really nice. My first thought was that nobody looks at this (because it's not in code), but we'll find a place to put it. Perhaps the wiki?

True, but it can still help to settle the orderings adopted by meshio and to make it easier to detect which conversions should still be added in the different readers, right? In that case, making it available in the wiki with maybe a link in [_common.py](https://github.com/nschloe/meshio/blob/master/meshio/_common.py) and/or [README.md](https://github.com/nschloe/meshio/blob/master/README.md) would already be quite useful, I think.

nschloe:
It's now in the wiki: https://github.com/nschloe/meshio/wiki/Node-ordering-in-cells.

michejansen:
OK, thanks. Should I add the elements mentioned above that are still missing in _gmsh_to_meshio_order in another PR?

PR #874, opened 2020-07-26
Add hexahedron27 and pyramid13 in _gmsh_to_meshio_order

michejansen:
Hi, I went ahead and added the node ordering for the hexahedron27 and pyramid13 elements in the gmsh reader. I tested the conversion between gmsh and paraview for a mesh containing hexahedron27 elements and it works fine now. I didn't test the pyramid13 element but it should be fine. I think those were the last missing second-order elements.


nschloe:
Cool, thanks!

Issue #865, opened 2020-07-13
Error writing GMSH format: "Can only deal with one cell type for now"


dajuno:
Hello,

meshio 4.0.16 fails to write GMSH files.

Consider the following MWE (pygmsh demo):

```python
import pygmsh
import numpy as np

geom = pygmsh.built_in.Geometry()

# Draw a cross.
poly = geom.add_polygon([
    [ 0.0,  0.5, 0.0],
    [-0.1,  0.1, 0.0],
    [-0.5,  0.0, 0.0],
    [-0.1, -0.1, 0.0],
    [ 0.0, -0.5, 0.0],
    [ 0.1, -0.1, 0.0],
    [ 0.5,  0.0, 0.0],
    [ 0.1,  0.1, 0.0]
    ],
    lcar=0.05
)

axis = [0, 0, 1]

geom.extrude(
    poly,
    translation_axis=axis,
    rotation_axis=axis,
    point_on_axis=[0, 0, 0],
    angle=2.0 / 6.0 * np.pi
)

mesh = pygmsh.generate_mesh(geom)

import meshio

# VTK works
meshio.write("test.vtk", mesh)

# gmsh fails 
meshio.write("test.msh", mesh)
```
Fails with:

```
Traceback (most recent call last):
  File "scripts/pygmsh_demo.py", line 38, in <module>
    meshio.write("test.msh", mesh)
  File "/home/nolte/.local/lib/python3.6/site-packages/meshio/_helpers.py", line 144, in write                                               
    return writer(filename, mesh, **kwargs)
  File "/home/nolte/.local/lib/python3.6/site-packages/meshio/gmsh/main.py", line 116, in <lambda>                                           
    "gmsh": lambda f, m, **kwargs: write(f, m, "4.1", **kwargs),
  File "/home/nolte/.local/lib/python3.6/site-packages/meshio/gmsh/main.py", line 107, in write                                              
    writer.write(filename, mesh, binary=binary, float_fmt=float_fmt)
  File "/home/nolte/.local/lib/python3.6/site-packages/meshio/gmsh/_gmsh41.py", line 277, in write                                           
    write4_1(filename, mesh, float_fmt=float_fmt, binary=binary)
  File "/home/nolte/.local/lib/python3.6/site-packages/meshio/gmsh/_gmsh41.py", line 317, in write4_1                                        
    _write_nodes(fh, mesh.points, mesh.cells, float_fmt, binary)
[...]

PR #880, opened 2020-08-08
Msh4 to msh2

michejansen:
Hi, I tried to use meshio to convert a gmsh41 file to a gmsh22 file and noticed that rereading this converted file in gmsh does not give a mesh with the same tags. This PR tries to fix this issue by mimicking the conversion behavior of gmsh itself. The following things are  modified:
1. When converting from gmsh4 to gmsh22 with meshio, the geometrical tag is missing in the resulting file. Change: the readers for gmsh40 and gmsh41 now also store the geometrical tag as cell data with the "gmsh:geometrical" key similar to the gmsh22 reader (edc7764).
2. If the `Entities` block is missing (which is allowed in the 4.1 format), the mesh file converted by gmsh will only contain the geometrical tag and the physical tags are set to zero to maintain the correct order (see number-of-tags entry [here](http://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-2-_0028Legacy_0029)). However, when using meshio to write a gmsh22 file of a mesh with "gmsh:geometrical" but without "gmsh:physical" tags, the geometrical tags are written directly after the element number. As a result, the geometrical tags are considered as physicals tags in gmsh. 2fbb8df fixes this by also writing a zero when either the physical or geometrical tag is missing.  
3. When `MeshFormat` is 4 (without decimals), the format corresponds to 4.0 instead of 4.1 (501eebd).

### DIFF
--- a/meshio/gmsh/_gmsh22.py
+++ b/meshio/gmsh/_gmsh22.py
@@ -352,10 +352,10 @@ def _write_elements(fh, cells, tag_data, binary):
     for k, (cell_type, node_idcs) in enumerate(cells):
         tags = []
         for name in ["gmsh:physical", "gmsh:geometrical", "cell_tags"]:
-            try:
+            if name in tag_data:
                 tags.append(tag_data[name][k])
-            except KeyError:
-                pass
+            elif name in ["gmsh:physical", "gmsh:geometrical"]:
+                tags.append(numpy.zeros(len(node_idcs), dtype=int))
         fcd = numpy.concatenate([tags]).astype(c_int).T
 
         if len(fcd) == 0:
--- a/meshio/gmsh/_gmsh22.py
+++ b/meshio/gmsh/_gmsh22.py
@@ -277,6 +277,25 @@ def write(filename, mesh, float_fmt=".16e", binary=True):
             [mesh.points[:, 0], mesh.points[:, 1], numpy.zeros(mesh.points.shape[0])]
         )
 
+    # Split the cell data: gmsh:physical and gmsh:geometrical are tags, the rest is
+    # actual cell data.
+    tag_data = {}
+    other_data = {}
+    for key, d in mesh.cell_data.items():
+        if key in ["gmsh:physical", "gmsh:geometrical", "cell_tags"]:
+            tag_data[key] = d
+        else:
+            other_data[key] = d
+
+    # Always include the physical and geometrical tags. See also the quoted excerpt from
+    # the gmsh documentation in the _read_cells_ascii function above.
+    for tag in ["gmsh:physical", "gmsh:geometrical"]:
+        if tag not in tag_data:
+            logging.warning(
+                "Appending zeroes to replace the missing {} tag data.".format(tag[5:])
+            )
+            tag_data[tag] = [numpy.zeros(len(x.data), dtype=c_int) for x in mesh.cells]
+
     if binary:
         for k, (key, value) in enumerate(mesh.cells):
             if value.dtype != c_int:
@@ -301,16 +320,6 @@ def write(filename, mesh, float_fmt=".16e", binary=True):
 
         if mesh.field_data:
             _write_physical_names(fh, mesh.field_data)
-
-        # Split the cell data: gmsh:physical and gmsh:geometrical are tags, the rest is
-        # actual cell data.
-        tag_data = {}
-        other_data = {}
-        for key, d in mesh.cell_data.items():
-            if key in ["gmsh:physical", "gmsh:geometrical", "cell_tags"]:
-                tag_data[key] = d
-            else:
-                other_data[key] = d
 
         _write_nodes(fh, mesh.points, float_fmt, binary)
         _write_elements(fh, cells, tag_data, binary)
@@ -354,8 +363,6 @@ def _write_elements(fh, cells, tag_data, binary):
         for name in ["gmsh:physical", "gmsh:geometrical", "cell_tags"]:
             if name in tag_data:
                 tags.append(tag_data[name][k])
-            elif name in ["gmsh:physical", "gmsh:geometrical"]:
-                tags.append(numpy.zeros(len(node_idcs), dtype=int))
         fcd = numpy.concatenate([tags]).astype(c_int).T
 
         if len(fcd) == 0:
--- a/meshio/gmsh/_gmsh22.py
+++ b/meshio/gmsh/_gmsh22.py
@@ -292,7 +292,7 @@ def write(filename, mesh, float_fmt=".16e", binary=True):
     for tag in ["gmsh:physical", "gmsh:geometrical"]:
         if tag not in tag_data:
             logging.warning(
-                "Appending zeroes to replace the missing {} tag data.".format(tag[5:])
+                "Appending zeros to replace the missing {} tag data.".format(tag[5:])
             )
             tag_data[tag] = [numpy.zeros(len(x.data), dtype=c_int) for x in mesh.cells]
 
--- a/meshio/gmsh/main.py
+++ b/meshio/gmsh/main.py
@@ -7,7 +7,7 @@ from . import _gmsh22, _gmsh40, _gmsh41
 
 # Some mesh files out there have the version specified as version "2" when it really is
 # "2.2". Same with "4" vs "4.1".
-_readers = {"2": _gmsh22, "2.2": _gmsh22, "4.0": _gmsh40, "4": _gmsh41, "4.1": _gmsh41}
+_readers = {"2": _gmsh22, "2.2": _gmsh22, "4.0": _gmsh40, "4": _gmsh40, "4.1": _gmsh41}
 _writers = {"2.2": _gmsh22, "4.0": _gmsh40, "4.1": _gmsh41}
 
 
--- a/meshio/gmsh/main.py
+++ b/meshio/gmsh/main.py
@@ -7,7 +7,7 @@ from . import _gmsh22, _gmsh40, _gmsh41
 
 # Some mesh files out there have the version specified as version "2" when it really is
 # "2.2". Same with "4" vs "4.1".
-_readers = {"2": _gmsh22, "2.2": _gmsh22, "4.0": _gmsh40, "4": _gmsh40, "4.1": _gmsh41}
+_readers = {"2": _gmsh22, "2.2": _gmsh22, "4.0": _gmsh40, "4": _gmsh41, "4.1": _gmsh41}
 _writers = {"2.2": _gmsh22, "4.0": _gmsh40, "4.1": _gmsh41}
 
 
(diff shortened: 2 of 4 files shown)


meshio/gmsh/_gmsh22.py
```diff
@@ -352,10 +352,10 @@ def _write_elements(fh, cells, tag_data, binary):
     for k, (cell_type, node_idcs) in enumerate(cells):
         tags = []
         for name in ["gmsh:physical", "gmsh:geometrical", "cell_tags"]:
-            try:
+            if name in tag_data:
                 tags.append(tag_data[name][k])
-            except KeyError:
-                pass
+            elif name in ["gmsh:physical", "gmsh:geometrical"]:
```
nschloe, line 357: So if physical or geometrical tags are not present, write zeros? I'm really not sure about this. If the user wants those tags written out, she can manually add them, right?

meshio/gmsh/main.py
```diff
@@ -7,7 +7,7 @@
 
 # Some mesh files out there have the version specified as version "2" when it really is
 # "2.2". Same with "4" vs "4.1".
-_readers = {"2": _gmsh22, "2.2": _gmsh22, "4.0": _gmsh40, "4": _gmsh41, "4.1": _gmsh41}
+_readers = {"2": _gmsh22, "2.2": _gmsh22, "4.0": _gmsh40, "4": _gmsh40, "4.1": _gmsh41}
```
nschloe, line 10: Unfortunately, gmsh didn't change between 4.0 and 4.1 in that "4" can mean both. So, this change needs to be reverted.

nschloe:
In general, I don't understand the changes of this PR well. And I dislike the gmsh format enough to not much care about 2.2. Typically I tell people to upgrade, but I'm willing to spend some time on this PR.

> Change: the readers for gmsh40 and gmsh41 now also store the geometrical tag as cell data with the "gmsh:geometrical" key similar to the gmsh22 reader (edc7764).

That doesn't sound right. What if you read a gmsh41 file that doesn't have a geo tag? You still get a dummy one? It seems that the gmsh writer needs an update here. Or is a missing geo tag not allowed in gmsh4[0,1]? Then we should throw an error.

> When MeshFormat is 4 (without decimals), the format corresponds to 4.0 instead of 4.1 (501eebd).

This needs to be reverted. 4.0 plays zero role and can be considered a mistake by the gmsh devs. Unfortunately, the gmsh devs allow version "4" which can mean both 4.0 and 4.1, depending on your gmsh version. Bad luck for us.

meshio/gmsh/_gmsh22.py
michejansen: That's basically what gmsh does. The physical and geometrical tags are always written in msh22 format. See also the number-of-tags entry [here](http://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format-version-2-_0028Legacy_0029). A zero means no tag and is used to ensure the correct order if the physical and/or geometrical tag is missing.

meshio/gmsh/main.py
michejansen: In the source code of gmsh, the mesh version is read as a double and the reader is filled with comparisons like `if(version >= 4.1)` so  '4' should imply gmsh40 format. In addition, if you write a mesh with a recent gmsh (I'm using 4.5), the `MeshFormat` block will always contain '4.1'. Only if you specify `gmsh -format msh40` on the command line, will `MeshFormat` contain '4', so that seems to be quite consistent? I included this change because I wanted to test this PR on a mesh file generated with `gmsh -format msh40`. meshio failed to read this initially, and only by switching to the gmsh40 did it work. The error was:
```
  File "meshio/meshio/gmsh/_gmsh41.py", line 115, in _read_entities
    (num_physicals,) = fromfile(f, c_size_t, 1)
ValueError: not enough values to unpack (expected 1, got 0)
```

michejansen:
> That doesn't sound right. What if you read a gmsh41 file that doesn't have a geo tag? You still get a dummy one? It seems that the gmsh writer needs an update here. Or is a missing geo tag not allowed in gmsh4[0,1]? Then we should throw an error.

In the msh40 and msh41 format, the geo tag (= entityTag in the $Elements block [here](http://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format)) is a (fixed) part of the header for each element block, so yes, it is always there. I do not really see the problem with always adding the "gmsh::geometrical" tag when reading a gmsh4 file.

meshio/gmsh/_gmsh22.py
nschloe: When you write "here", it's not clear what you mean. Please be more specific.

michejansen: I mean this section of the gmsh documentation (following that link):

> number-of-tags gives the number of integer tags that follow for the n-th element. By default, the first tag is the tag of the physical entity to which the element belongs; the second is the tag of the elementary model entity to which the element belongs; the third is the number of mesh partitions to which the element belongs, followed by the partition ids (negative partition ids indicate ghost cells). A zero tag is equivalent to no tag. Gmsh and most codes using the MSH 2 format require at least the first two tags (physical and elementary tags).

meshio/gmsh/main.py
nschloe: >  so '4' should imply gmsh40 format.

See [here](https://github.com/nschloe/meshio/blob/master/meshio/gmsh/main.py#L9). We've have problems with "4" meaning "4.0" in the past. I remember we've changed that before, and it's always going back and forth, which is why I'm reluctant to change here. Honestly if gmsh writes "4" and mean "4.0", than that's a bug in gmsh.

Since this creates some confusion, I'm inclined to disallow format "4" altogether and always error out.

>  gmsh -format msh40

Wait, _this_ leads to the 4.1 reader? That looks like a bug.

nschloe:
> I do not really see the problem with always adding the "gmsh::geometrical" tag when reading a gmsh4 file.

It's generally a problem if the writer silently adds data. If the data is required and the user didn't provide it, it usually means there's a user error. In this case it's best to raise and exception and force the user to provide the missing data. If you silently add the data, then in `mesh2 = read(write(mesh1))`, `mesh2` will contain more than `mesh1`, which is weird.

nschloe:
@gdmcbain Geordie, do you remember the confusion around gmsh 4.0 and 4.1?

meshio/gmsh/_gmsh22.py
nschloe: Okay, good. If one of the two "default" tags is missing, I think we should issue warning. I don't care much if we add the data (zeros) and warn about it or don't add it and show a different warning. This should happen at the very beginning of the write step.

michejansen:
> > I do not really see the problem with always adding the "gmsh::geometrical" tag when reading a gmsh4 file.
> 
> It's generally a problem if the writer silently adds data. If the data is required and the user didn't provide it, it usually means there's a user error. In this case it's best to raise and exception and force the user to provide the missing data. If you silently add the data, then in `mesh2 = read(write(mesh1))`, `mesh2` will contain more than `mesh1`, which is weird.

I agree that's a bit weird, but couldn't we use the same solution as you just proposed for the gmsh22 writer: issue a warning when the gmsh40 writer adds a nonexistant entityTag. It is at that point that additional data is created. At the moment, the gmsh40 writer always writes '1' as entity tag (see [here](https://github.com/nschloe/meshio/blob/6dc52f1ae09aa1c65936679e1b68a1939ba83a85/meshio/gmsh/_gmsh40.py#L366)). I suppose this could be replaced by the "gmsh:geometrical" data if it exists.

gdmcbain:
> @gdmcbain Geordie, do you remember the confusion around gmsh 4.0 and 4.1?

Hi Nico. I sure do. It was bad; nasty problems like certain long-term supported Linux distributions providing a clearly buggy version of Gmsh 4 which Gmsh itself had quickly moved on from. Really miserable stuff.. See your comment of 2019-03-23 in #365.

I had thought that after a lot of to and fro meshio had struck a pretty reasonable compromise, except for.the huge hole in that we can't write MSH 4.1 but fall back on MSH 2.2, which also means that pygmsh, e.g., writes MSH 2. Besides other programs still relying on MSH 2, that's the main reason we do still need to support MSH 2.2. See my comment of 2020-07-14 in #865.

Sorry not to be more hands-on in this issue, I'm away from.the terminal. Thank you @michejansen for the thoughtful pull request and patient advocacy of it.  

Both sides of the argument on missing geometrical tags are persuasive. Being away from the terminal, I can't run tests myself : @michejansen, under which circumstances do you encounter MSH 4.1 files in which geometrical or physical tags are missing? In my use cases, the geometrical tags are never of interest, whereas the physical tags very much are, representing subdomains or patches of boundary in mixed boundary value problems.

michejansen:
> Both sides of the argument on missing geometrical tags are persuasive. Being away from the terminal, I can't run tests myself : @michejansen, under which circumstances do you encounter MSH 4.1 files in which geometrical or physical tags are missing? In my use cases, the geometrical tags are never of interest, whereas the physical tags very much are, representing subdomains or patches of boundary in mixed boundary value problems.

@gdmcbain thanks for the input! I'm using a (commercial) software that requires MSH 2.2 files containing both physical and geometrical tags. I think the geometrical tags are mainly used for mesh adaptation in this software, which is usually performed on geometrical instead of physical entities (see also [this comment](https://github.com/nschloe/meshio/pull/880#discussion_r467587507)).
Personally, I did not yet encounter a MSH 4.1 file with missing geometrical or physical tags. But the gmsh documentation mentions the following difference between 4.0 and 4.1:

> The $Entities section is now optional.
 
Since the Entities block contains the physical tags, it is possible to remove the physical tags by manually removing this block from the file and gmsh will read it without problem.

nschloe:
@michejansen Can you implement the suggested change for 2.2 and revert back to "4" meaning "4.1"?

michejansen:
> @michejansen Can you implement the suggested change for 2.2 and revert back to "4" meaning "4.1"?

OK, done. I did move around some code in 6b815ab to be able to put the warning and the actual appending of zeroes in the same place with the other warnings without modifying the input mesh object.

meshio/gmsh/_gmsh22.py
```diff
@@ -277,6 +277,25 @@ def write(filename, mesh, float_fmt=".16e", binary=True):
...
+            tag_data[key] = d
+        else:
+            other_data[key] = d
+
+    # Always include the physical and geometrical tags. See also the quoted excerpt from
+    # the gmsh documentation in the _read_cells_ascii function above.
+    for tag in ["gmsh:physical", "gmsh:geometrical"]:
+        if tag not in tag_data:
+            logging.warning(
+                "Appending zeroes to replace the missing {} tag data.".format(tag[5:])
```
nschloe, line 295: zeros not zeroes

michejansen: Right :/ Fixed.

nschloe:
LGTM now, thanks for the good work!

Issue #855, opened 2020-06-24
KeyError: 'vtkidtype'


nilswagner:
```
import meshio
m = meshio.Mesh.read('test_mod.vtk', "vtk")
m.write('test.dato')
```

nilswagner:
[test_mod.vtk.gz](https://github.com/nschloe/meshio/files/4824925/test_mod.vtk.gz)

nilswagner:
The vtk file was generated by pyansys 0.41.3

nschloe:
Fixed now.

ylgrst:
Hello,
I am having the same error when reading a vtk file created using pyvista version 0.42.2 (see file attached)
The error occurs when using command line: ```meshio convert test.vtk test.mesh``` as well as through python functions:

```
import meshio
m = meshio.read('test.vtk')
m.write('test.mesh')
```

I have also tried saving as ascii or binary, to no avail...

[test.vtk.gz](https://github.com/nschloe/meshio/files/12708517/test.vtk.gz)

marchirschvogel:
I encounter the same problem with a file I've received, and with the example file provided here. The error happens in
`m = meshio.read('test.vtk')`
Would be great if this gets resolved.

PR #881, opened 2020-08-10
skip COLOR_SCALARS gracefully in vtk files

### LINKED ISSUE #854: meshio._exceptions.ReadError: Unknown section 'COLOR_SCALARS'
```
import meshio
m = meshio.Mesh.read('test.vtk', "vtk")
m.write('test.dato')
```
[test.vtk.gz](https://github.com/nschloe/meshio/files/4823444/test.vtk.gz)

ChristosT:
This resolves #854 but the file attached to that issue cannot be read completly yet since it depends at least on #855 too.

### DIFF
--- a/meshio/vtk/_vtk.py
+++ b/meshio/vtk/_vtk.py
@@ -113,6 +113,7 @@ vtk_sections = [
     "POINT_DATA",
     "CELL_DATA",
     "LOOKUP_TABLE",
+    "COLOR_SCALARS",
 ]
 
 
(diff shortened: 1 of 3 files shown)


nschloe:
> cannot be read completly yet since it depends at least on #855 too.

Why do the tests pass then?

ChristosT:
> > cannot be read completly yet since it depends at least on #855 too.
> 
> Why do the tests pass then?

I used a simpler file that just adds COLOR_SCALARS.

I will be looking to #855 soon so, as an alternative we can leave this open until I have both.

nschloe:
> I used a simpler file that just adds COLOR_SCALARS.


I was just wondering if we shouldn't assert the existence and shape etc. of COLOR_SCALARS in the output.

ChristosT:
> I was just wondering if we shouldn't assert the existence and shape etc. of COLOR_SCALARS in the output.

Sorry I didn't get that. With this commit the COLOR_SCALARS are skipped just like the LOOKUP_TABLE data. Are you suggesting that we should assert the existence and the size of the entry in the test ?

ChristosT:
Together with #885 it can parse the test file in #854 successfully

nschloe:
> With this commit the COLOR_SCALARS are skipped just like the LOOKUP_TABLE data.

Fair enough. This is already an improvement.

PR #885, opened 2020-08-14
Vtk5.1

### LINKED ISSUE #877: VTK file format 5.0
I created a vtk file using paraview 5.8 and it wirtes the file using format 5.0. This file contains keywords like offset and connectivity which are not recognize by meshio.

vlukes:
In `vtk DataFile Version 5.1` the cells are stored in OFFSETS/CONNECTIVITY data blocks.

My _Paraview_ v5.8 produces VTK files with `vtktypeint64` types after  OFFSETS/CONNECTIVITY keywords, so I added `vtktypeint32` and `vtktypeint64` to `vtk_to_numpy_dtype_name` dictionary.

When extracting points or cells in _Paraview_, the original ids are stored in fields `vtkOriginalPointIds`/`vtkOriginalCellIds` and their type is `vtkIdType`. This is usually `int` (`int32`) type, but it can also be `long` (`int64`). See file [1], lines 308-336. I don't know an easy way to determine it in the code.

This PR should fix #877.

[1]  https://vtk.org/doc/nightly/html/vtkType_8h_source.html

### DIFF
--- a/meshio/vtk/_vtk.py
+++ b/meshio/vtk/_vtk.py
@@ -76,6 +76,7 @@ vtk_to_numpy_dtype_name = {
     "double": "float64",
     "vtktypeint32": "int32",  # vtk DataFile Version 5.1
     "vtktypeint64": "int64",  # vtk DataFile Version 5.1
+    "vtkidtype": "int32",  # may be either 32-bit or 64-bit (VTK_USE_64BIT_IDS)
 }
 
 numpy_to_vtk_dtype = {
--- a/meshio/vtk/_vtk.py
+++ b/meshio/vtk/_vtk.py
@@ -223,7 +223,11 @@ def _read_section(f, info):
             line = f.readline().decode("utf-8")
         except UnicodeDecodeError:
             line = ""
-        if "OFFSETS" in line:  # vtk DataFile Version 5.1
+        if "OFFSETS" in line:
+            # vtk DataFile Version 5.1 - appearing in Paraview 5.8.1 outputs
+            # No specification found for this file format.
+            # See the question on ParaView Discourse Forum:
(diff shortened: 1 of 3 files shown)


nschloe:
For some reason, the vtk file wasn't picked up by git lfs. Any idea why?

ChristosT:
Good Job ! I just tried it  and it also resolves https://github.com/nschloe/meshio/issues/855

vlukes:
I force-pushed the comits. I think that should be fine by now.

nschloe:
@vlukes I still see test/meshes/vtk/06_unstructured.vtk as non gitlfs.

vlukes:
Sorry, my fault in local LFS settings. Now, you can see `test/meshes/vtk/06_unstructured.vtk` as Git LFS file.

nschloe:
> In vtk DataFile Version 5.1 the cells are stored in OFFSETS/CONNECTIVITY data blocks.

Can you provide a link to the specification?

vlukes:
Unfortunately, I can't find the exact specification for version `vtk DataFile Version 5.1`. This PR is based only on the results produced by Paraview 5.8. I asked on the paraview forum, but still no answer.

nschloe:
> Unfortunately, I can't find the exact specification for version vtk DataFile Version 5.1. This PR is based only on the results produced by Paraview 5.8. I asked on the paraview forum, but still no answer.

Okay, then this needs to be documented, including a link to the question you asked.

I was wondering about the specification because I don't see the format version appear anywhere except in a comment. This is very untypical, even bad. I don't like the fact that we have to guess our way through the file with tell and seek.

vlukes:
I added a comment about the missing specification and a link to my question [1].

[1] https://discourse.paraview.org/t/specification-of-vtk-datafile-version-5-1/5127

nschloe:
Good, let's take this. Thanks a lot!

PR #900, opened 2020-09-01
Tetgen: read and write point/cell data

vlukes:
Reading:
* Nodes can have attributes and  boundary markers (see [2]), these are loaded into `point_data` dictionary using `tetgen:attr1`, `tetgen:attr2`, ... keys for the attributes and `tetgen:ref`, `tetgen:ref2`, ... keys for the markers. The first boundary marker (if exists) is assumed to be a _point group id_ and is labeled as `tetgen:ref`.
* Cells can only have attributes. A common use is to identify a subregion or material of cells (cell groups), see [1].  Cell attributes
are loaded as `tetgen:ref`, `tetgen:ref2`, ... cell data fields.

Writing:
* If there is a `:ref` field in `point_data`, it is used as the boundary marker, if not, the "first" field in point data is employed. The other `point_data` fields are stored as node attributes.
* If there is a `:ref` field in `cell_data`, it is used as the first region marker. 
* The names of point/cell data fields are stored in comment lines.

[1] http://wias-berlin.de/software/tetgen/fformats.ele.html
[2] http://wias-berlin.de/software/tetgen/fformats.node.html

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,3 +13,5 @@
 *.meshb filter=lfs diff=lfs merge=lfs -text
 *.tec filter=lfs diff=lfs merge=lfs -text
 *.su2 filter=lfs diff=lfs merge=lfs -text
+*.ele filter=lfs diff=lfs merge=lfs -text
+*.node filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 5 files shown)


nschloe:
Cool, thanks.

Issue #911, opened 2020-09-11
xdmf.TimeSeriesWriter speed regression


gdmcbain:
`xdmf.TimeSeriesWriter` slowed down significantly between 3.3.1 (2019-12-19) and now (4.2.0).

I have a test-case based on https://github.com/hplgit/fenics-tutorial/blob/master/pub/python/vol1/ft08_navier_stokes_cylinder.py (which doesn't use meshio) which writes 5000 time-steps of velocity on a triangular mesh with about 2000 points.  My [test-case](https://github.com/gdmcbain/fenics-tuto-in-skfem/tree/master/08_navier_stokes_cylinder) does use `meshio.xdmf.TimeSeriesWriter` and on 2019-10-17 took 12 minutes but today 21 minutes.

The test-case is a bit involved, but to reproduce the problem, try:
```python
import numpy as np

import meshio


def mwe(m=512, n=32, l=2):

    print("meshio", meshio.__version__)

    mesh = meshio.Mesh(np.zeros((2048, 2)), [])
    point_data = {"u": np.zeros_like(mesh.points)}

    with meshio.xdmf.TimeSeriesWriter("mwe.xdmf") as writer:
        writer.write_points_cells(mesh.points, mesh.cells)

        for k in range(m):
            writer.write_data(k, point_data=point_data)


if __name__ == "__main__":
    from sys import argv

    mwe(*map(int, argv[1:]))
```
Then:
```shell
pip install meshio==3.3.0
time python mwe.py
pip install -U meshio
time python mwe.py
```
The time increases from 3 to 6 s.  Then try
```shell
time python mwe.py 1024
```
and it's 23 s; this is O (*m*&sup2;).

gdmcbain:
Initially I suspected #623 &lsquo;lxml &rarr; built-in xml&rsquo; (2020-01-06), but I'm hoping to be able to fix this without bringing back the dependency.

gdmcbain:
meshio 3.3.0 is also quadratic: *m* = 1024, 2048, & 4096 takes 10, 32, & 110&nbsp;s.  I'm thinking that this task should be O (*m*).

gdmcbain:
meshio 4.2.0 takes 332&nbsp;s for _m_ = 4096; increase of 332/23=14 which is close enough to (4096/1024)&sup2; = 16.

gdmcbain:
I wonder whether it would be reasonable to not

[...]

PR #912, opened 2020-09-11
write_xml in TimeSeriesWriter exit

gdmcbain:
This uses the `TimeSeriesWriter.__exit__` block of the context manager for writing the XML metadata to the `.xdmf` file; the actual data is still written each `write_data` time-step via `.point_data` and `.numpy_to_xml_string`.

### DIFF
--- a/meshio/xdmf/time_series.py
+++ b/meshio/xdmf/time_series.py
@@ -285,7 +285,6 @@ class TimeSeriesWriter:
         self.cells(cells, grid)
         self.has_mesh = True
 
-
     def write_data(self, t, point_data=None, cell_data=None):
         cell_data = {} if cell_data is None else cell_data
         # <Grid>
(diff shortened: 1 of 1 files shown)


gdmcbain:
With this, the MWE from #911 runs in 0.9, 2.1, & 3.9 s with _m_ = 1024, 2048, & 4096 which is O (*m*) and 85 times faster than master on the last.

gdmcbain:
This failure is codecov/project.  I don't think it's valid.  Two lines of code&mdash;calls to write XML when writing points & cells and when writing data&mdash;have been replaced by one, just writing the XML on exit.

gdmcbain:
This is also 110/3.9 = 28 times faster than the old pre-#623 lxml implementation on the _m_ = 4096 MWE run.

nschloe:
Cool, thanks! (codecov complains because the coverage went down -- that's not a problem here, quite the contrary.)

PR #914, opened 2020-09-15
stop at EOF and warn if $EndElements omitted

### LINKED ISSUE #913: "meshio" can't read a version-2 "Gmsh" msh file
Hi,

originally I wanted to read "Gmsh" version-4 mesh files with tags / regions using "meshio". I first tried to read a version-2 file in order to convert it into a version-4 file with tags / regions but "meshio-info" and the "meshio.read" function both do not terminate / return on reading the version-2 file (attached). "Gmsh" is able to load the file so I think the syntax is valid.

[test.zip](https://github.com/nschloe/meshio/files/5209698/test.zip)

A second problem which I thought was already fixed in a previous version of "meshio" is that "meshio" can't read off-files with RGB color information ("Can only read triangular faces"). I'm not using off-files any more, I just wanted to mention that...

And finally: Are there any plans to support "*.dat" files from CATIA? ;-)

Thanks & kind regards,

Rainer

gdmcbain:
Fixes #913. 

Should a missing `$EndElements` raise a warning or an exception?

This just reads until it finds `$EndElements` or reaches the end of the file.  It doesn't attempt to handle files which begin a new block (e.g. `$NodeData`) without having closed the `$Elements` block; it will just read right through all that and ignore it.

### DIFF
--- a/meshio/gmsh/_gmsh22.py
+++ b/meshio/gmsh/_gmsh22.py
@@ -137,7 +137,7 @@ def _read_cells(f, cells, point_tags, is_ascii):
         cells[ic] = (ct, remap[cd])
 
     for line in f:
-        if line.decode("utf-8") == "$EndElements":
+        if line.decode("utf-8") == "$EndElements\n":
             break
     else:
         logging.warning("$Elements not closed by $EndElements.")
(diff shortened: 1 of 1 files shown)


nschloe:
@gdmcbain Some errors here?

nschloe:
TIL about `for`-`else`. Thanks!

Issue #660, opened 2020-02-04
polyMesh


gdmcbain:
Wishlist stub for a new format to support: [OpenFOAM](https://cfd.direct/openfoam)'s [polyMesh](https://cfd.direct/openfoam/user-guide/v7-mesh-description/#x24-1680005.1).

> By default OpenFOAM defines a mesh of arbitrary polyhedral cells in 3-D, bounded by arbitrary polygonal faces, i.e. the cells can have an unlimited number of faces where, for each face, there is no limit on the number of edges nor any restriction on its alignment. A mesh with this general structure is known in OpenFOAM as a polyMesh. 

Having arbitrary polyhedral cells bounded by arbitrary polygonal faces is not a great fit for meshio, but in practice many meshes consist purely of hexahedra bounded by quadrilaterals (e.g. those built with [blockMesh](https://cfd.direct/openfoam/user-guide/v7-blockmesh/#x26-1850005.3)) so even supporting just that subset might be worthwhile.

Given the importance of OpenFOAM in contemporary computational fluid mechanics, it's a mesh format that often has to be dealt with; e.g. for coupling OpenFOAM with finite element solvers.  As mentioned in #555, I've done a bit of this (unpublished) in [pyparsing](https://github.com/pyparsing/pyparsing) (which is expressive but slow) but it might be worth porting to meshio and sharing.

OpenFOAM does provide a [`foamToVTK`](https://cfd.direct/openfoam/user-guide/v7-standard-utilities/#dx14-116011) utility which is often used to view results in ParaView &c.; it's output might be a helpful guide in developing a meshio polyMesh module.

tianyikillua:
Totally upvote for this. As I understand it, the OpenFOAM mesh is not just composed of one file but a folder of files: nodes, cells, faces...

gdmcbain:
Yes.  In the top directory of an OpenFOAM ‘case’ not involving dynamic meshing, the polyMesh is stored in the directory `constant`; e.g.
```shell
tree constant/polyMesh
```
gives
```
constant/polyMesh/
├── boundary
├── cellZones
├── faces
├── neighbour
├── owner
└── points
```
[...]

PR #916, opened 2020-09-16
Fix of the vtu polyhedron filter

keileg:
## Background
The current vtu filter for polyhedron cells is not compatible with the specification of the vtu format. The error was introduced (by me) in a5f59bb.

## Changes
* The vtu polyhedron filter requires, for each cell, the nodes of each of the cell's faces. This is implemented as a list of numpy arrays. 
* Reading vtu files with general and varying number of faces per cell and nodes per face, is complex compared to more standardized cases. The extensions are (mainly) confined to specific blocks of code, which should not need to be touched for changes to other grid types etc. At a few points, this localization led to slightly more complex code, but I think it is worth it.

## Main problem with the presented solution
The list of cell-faces has, from what I can see, no natural placement in the Mesh data structure. I tried to put it as cell_data, but this breaks the convention that cell_data is numpy arrays. This lead to all kind of trouble - for a glaring example see the changes to helpers.py, function write_read(). 

My preferred solution would be to introduce a new dictionary on the top level of the Mesh class (name: cell_faces?), and move the information there. This should have no impact on the rest of the code base (except that the ugly hack in the test function can be moved to a subfunction), but it will significantly reduce the entropy of the affected parts of the vtu filter. On a wider level, a dedicated field may be useful for other grid formats as well. However, I thought it best to first present a solution with less impacts on the fundamentals - I assume you try to be restrictive with changes to the Mesh class. 

Should you be willing to introduce such a new top-level field, I can take care of implementation (should be easy), and include it in this PR. I do not have oversight of all other grid formats supported by meshio, but from what I can tell, no changes outside Mesh and vtu are needed.

## Other comments
* The naming convention for polyhedron cell blocks is now explicitly set as polyhedron{number_of_nodes_per_cell} (it could have been faces per cell).
* The filter can handle polyhedra with different number of nodes, but not combinations of polyhedra with other types of cells. It should not be too difficult to implement, but the preferred solution in this case is to represent all cells as polyhedra.

### DIFF
--- a/meshio/vtu/_vtu.py
+++ b/meshio/vtu/_vtu.py
@@ -152,17 +152,12 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                     start_cn[items + 1],
                     _vtk_to_meshio_order(types[start], sz, dtype=offsets.dtype) - sz,
                 )
-                if meshio_type == "polyhedron":
-                    # Polyhedra CellBlocks have special data
-                    cells.append(
-                        CellBlock(
-                            meshio_type + str(sz), polyhedron_faces[f"polyhedron{sz}"]
-                        )
-                    )
-                else:
-                    cells.append(
-                        CellBlock(meshio_type + str(sz), connectivity[indices])
-                    )
+                data = (
+                    polyhedron_faces[f"polyhedron{sz}"]
+                    if meshio_type == "polyhedron"
+                    else connectivity[indices]
+                )
+                cells.append(CellBlock(meshio_type + str(sz), data))
 
                 # Store cell data for this set of cells
                 for name, d in cell_data_raw.items():
@@ -640,6 +635,7 @@ def write(filename, mesh, binary=True, compression="zlib", header_type=None):
     for c in mesh.cells:
         if c.type[:10] == "polyhedron":
             is_polyhedron_grid = True
+            break
 
     if not binary:
         logging.warning("VTU ASCII files are only meant for debugging.")
--- a/meshio/vtu/_vtu.py
+++ b/meshio/vtu/_vtu.py
@@ -65,11 +65,6 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                 #   ...
                 # num_faces_cell_1,
                 #   ...
-                #
-                # The faceoffsets describes the end of the face description for each
-                # cell, but this is 1-offset, so it is effectively the start of the
-                # next cell.
-                #
                 # See https://vtk.org/Wiki/VTK/Polyhedron_Support for more.
 
                 # There is a tacit assumption here that the cell-data keys 'faces'
@@ -85,6 +80,8 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                     # in _organize_cells()
                     faces = cell_data_raw.pop("faces")
                     faceoffsets = cell_data_raw.pop("faceoffsets")
+                    # The faceoffsets describes the end of the face description for each
+                    # cell.
                     # Switch faceoffsets to give start points, not end points
                     faceoffsets = numpy.append([0], faceoffsets[:-1])
                     # Take note that we have found the faces
--- a/meshio/vtu/_vtu.py
+++ b/meshio/vtu/_vtu.py
@@ -152,7 +152,17 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                     start_cn[items + 1],
                     _vtk_to_meshio_order(types[start], sz, dtype=offsets.dtype) - sz,
                 )
-                cells.append(CellBlock(meshio_type + str(sz), connectivity[indices]))
+                if meshio_type == "polyhedron":
+                    # Polyhedra CellBlocks have special data
+                    cells.append(
+                        CellBlock(
+                            meshio_type + str(sz), polyhedron_faces[f"polyhedron{sz}"]
+                        )
+                    )
+                else:
+                    cells.append(
+                        CellBlock(meshio_type + str(sz), connectivity[indices])
+                    )
 
                 # Store cell data for this set of cells
                 for name, d in cell_data_raw.items():
@@ -160,7 +170,7 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                         cell_data[name] = []
                     cell_data[name].append(d[start + items])
         else:
-            # Same number of nodes per cell
+            # Non-polyhedron cell. Same number of nodes per cell makes everything easier.
             n = num_nodes_per_cell[meshio_type]
             indices = numpy.add.outer(
                 offsets[start:end],
@@ -172,7 +182,7 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                     cell_data[name] = []
                 cell_data[name].append(d[start:end])
 
-    return cells, cell_data, polyhedron_faces
+    return cells, cell_data
 
 
 def _organize_cells(point_offsets, cells, cell_data_raw):
@@ -180,7 +190,6 @@ def _organize_cells(point_offsets, cells, cell_data_raw):
         raise ReadError()
 
     out_cells = []
-    polyhedron_faces = {}
     for offset, cls, cdr in zip(point_offsets, cells, cell_data_raw):
 
         # Special treatment of information on polyhedron cells: Add
@@ -190,25 +199,28 @@ def _organize_cells(point_offsets, cells, cell_data_raw):
         if "faces" in cls.keys() and "faceoffsets" in cls.keys():
             cdr["faces"] = cls["faces"]
             cdr["faceoffsets"] = cls["faceoffsets"]
-
-        cls, cell_data, polyhedron_faces = _cells_from_data(
+        cls, cell_data = _cells_from_data(
             cls["connectivity"].ravel(),
             cls["offsets"].ravel(),
             cls["types"].ravel(),
             cdr,
         )
+
         for c in cls:
-            out_cells.append(CellBlock(c.type, c.data + offset))
+            if c.type[:10] == "polyhedron":
+                # Adding offsets is cumbersome for polyhedra due to its convoluted
+                # data types. Moreover, it is assumed that polyhedra are not combined
+                # with other cell types (if so, these should also be represented as
+                # polyhedra)
+                if offset != 0:
+                    raise ValueError(
+                        "Polyhedron cells should not be combined with other cell types"
+                    )
+                out_cells.append(CellBlock(c.type, c.data))
+            else:
+                out_cells.append(CellBlock(c.type, c.data + offset))
 
-        # NOTE: if polyhedron_faces are used, it is assumed that all cells
-        # are represented as polyhedra. Thus the outer for-loop will run a single
-        # iteration, and polyhedral_faces will be available to return
-        if len(polyhedron_faces) > 0 and point_offsets.size > 1:
-            raise NotImplementedError(
-                "Polyhedron meshes should not be combined with other cell types"
-            )
-
-    return out_cells, cell_data, polyhedron_faces
+    return out_cells, cell_data
 
 
 def get_grid(root):
@@ -478,7 +490,7 @@ class VtuReader:
         else:
             self.point_data = None
 
-        self.cells, self.cell_data, self.polyhedron_faces = _organize_cells(
+        self.cells, self.cell_data = _organize_cells(
             point_offsets, cells, cell_data_raw
         )
         self.field_data = field_data
@@ -606,7 +618,6 @@ def read(filename):
         point_data=reader.point_data,
         cell_data=reader.cell_data,
         field_data=reader.field_data,
-        polyhedron_faces=reader.polyhedron_faces,
     )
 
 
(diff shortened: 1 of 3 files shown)


gdmcbain:
> My preferred solution would be to introduce a new dictionary on the top level of the Mesh class (name: cell_faces?), and move the information there. This should have no impact on the rest of the code base (except that the ugly hack in the test function can be moved to a subfunction), but it will significantly reduce the entropy of the affected parts of the vtu filter. On a wider level, a dedicated field may be useful for other grid formats as well.

Yes, e.g. polyMesh #660.

nschloe:
> My preferred solution would be to introduce a new dictionary on the top level of the Mesh class (name: cell_faces?)

This sounds good. `cell_faces` or `polyhedron_faces` are a reasonable names. Best go ahead and add it to the mesh class.

keileg:
Great. I will take care of it within a few days.

Wrt name, I like polyhedron_faces better - it makes it clearer that this is necessary for certain cell types only. So I will use this unless I hear otherwise.

keileg:
polyhedron_faces are now added to the Mesh, and the vtu reader uses it.

meshio/vtu/_vtu.py
```diff
@@ -53,6 +54,93 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
         if meshio_type == "polygon" or meshio_type == "polyhedron":
```
nschloe, line 54: Can we change this to `meshio_type in ["polygon", "polyhedron"]`? (I know you didn't touch this, it just caught my attention.)

meshio/vtu/_vtu.py
```diff
@@ -53,6 +54,93 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
...
+                # been checked, we will take our chances.
+                try:
+                    # pop removes the fields that were added to the raw cell data
+                    # in _organize_cells()
+                    faces = cell_data_raw.pop("faces")
+                    faceoffsets = cell_data_raw.pop("faceoffsets")
+                    # Switch faceoffsets to give start points, not end points
+                    faceoffsets = numpy.append(0, faceoffsets[:-1])
+                    # Take note that we have found the faces
+                    found_faces_before = True
```
nschloe, line 88: This variable needs to be initialized somwhere.

meshio/vtu/_vtu.py
```diff
@@ -53,6 +54,93 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
...
+                # here.
+                # Whether such a file is permissible under the .vtu standard has not
+                # been checked, we will take our chances.
+                try:
+                    # pop removes the fields that were added to the raw cell data
+                    # in _organize_cells()
+                    faces = cell_data_raw.pop("faces")
+                    faceoffsets = cell_data_raw.pop("faceoffsets")
+                    # Switch faceoffsets to give start points, not end points
+                    faceoffsets = numpy.append(0, faceoffsets[:-1])
```
nschloe, line 86: These two lines should go in the `else` block of `try`-`except`.

nschloe, line 86: The `numpy.append(0, ...` is a bit weird since `0` is not an array. I like `numpy.append([0], faceoffsets[::-1])` better.

meshio/vtu/_vtu.py
```diff
@@ -53,6 +54,93 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
...
+                    # Take note that we have found the faces
+                    found_faces_before = True
+                except KeyError:
+                    # We will end up here in the second iteration of the loop of types.
```
nschloe, line 90: Thanks for the explanation, I wouldn't have understood it otherwise. I'd vote for refactoring this a bit. Have a  variable `is_first_iteration` or `is_second_iteration` and `if` our way around the code. Seems more explicit to me. I'm not 100% on this, what do you think?

meshio/vtu/_vtu.py
```diff
@@ -689,6 +830,16 @@ def text_writer(f):
...
                 key_ = k[:10]
+                # Get face-cell relation on the vtu format. See comments in helper
+                # function for more information of how to specify this.
+                faces_loc, faceoffsets_loc = _polyhedron_face_cells(
+                    mesh.polyhedron_faces[k]
+                )
+                # Adjust offsets to global numbering
+                if faceoffsets.size > 0:
+                    faceoffsets_loc += faceoffsets[-1]
+                faces = numpy.append(faces, faces_loc)
```
nschloe, line 841: Better collect all faces in a list, and at the end call a `numpy.concatenate`.

meshio/vtu/_vtu.py
keileg: Not sure I agree:  the fields are popped on the first iteration, and then available on other iterations. 

Another matter is that the usage of try-except is artificial. It would arguably be better to pop the information before the loop over b (L52). This would add a bit of entropy for non-polyhedron meshes. Let me know what you prefer.

meshio/vtu/_vtu.py
keileg: Your suggestion is far better than the current code. Or we can move the popping out of the main for-loop, as suggested above.

meshio/vtu/_vtu.py
keileg: thanks

meshio/vtu/_vtu.py
keileg: Will do.

meshio/vtu/_vtu.py
nschloe: > Not sure I agree: the fields are popped on the first iteration, and then available on other iterations.

Right.

> Another matter is that the usage of try-except is artificial. It would arguably be better to pop the information before the loop over b (L52). This would add a bit of entropy for non-polyhedron meshes. Let me know what you prefer.

Just do it before the loop, let's see what that looks like.

keileg: agree.

keileg: I ended up implementing a version of your suggestion below, with first iteration (although I called the variable found_polyhedron_faces_before). If nothing else, that is explicit on when the variable will be used.

meshio/vtu/_vtu.py
keileg: Implemented.

meshio/vtu/_vtu.py
keileg: I ended up with a version of your suggestion.

test/helpers.py
```diff
@@ -242,20 +242,48 @@
             [1.0, 0.0, 1.0],
             [1.0, 1.0, 1.0],
             [0.0, 1.0, 1.0],
-            [0.0, 0.0, 2.0],
-            [1.0, 0.0, 2.0],
-            [1.0, 1.0, 2.0],
-            [0.0, 1.0, 2.0],
         ]
     ),
     [  # Split the lower cube into tets and pyramids. The upper cube is hexahedron
```
nschloe, line 247: This comment is no longer correct, is it?

nschloe:
> The list of cell-faces has, from what I can see, no natural placement in the Mesh data structure

Why do we keep a separate list `polyhedron_faces` instead of treating the faces as polygon `cells`?

meshio/vtu/_vtu.py
```diff
@@ -43,16 +43,96 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
...
+            if meshio_type == "polyhedron":
+                # The data format for face-cells is:
+                # num_faces_cell_0,
+                #   num_nodes_face_0, node_ind_0, node_ind_1, ..
+                #   num_nodes_face_1, node_ind_0, node_ind_1, ..
+                #   ...
+                # num_faces_cell_1,
+                #   ...
+                #
+                # The faceoffsets describes the end of the face description for each
```
nschloe, line 69: Is faceoffsets a variable? I don't see it anywhere here.

nschloe:
> This sounds good. cell_faces or polyhedron_faces are a reasonable names. Best go ahead and add it to the mesh class.

Quoting myself here. Question: Why don't we include polyhedra into the cells list? We do that for polygons, right?

keileg:
Not sure I follow you. You want to add the polyhedron faces (which by themselves are polygons) as separate cells? I guess that can be done, but you would then need to identify which of these new cells go together to form a 3d cell? Or am I missing something?

nschloe:
> Not sure I follow you.

It's a question. If I see things correctly, a VTU file contains polyhedra as a list of polygonal faces. Is this correct. Why do we need a separate member in `Mesh` for it? Can't we just store the polyhedra as they come in a `CellBlock` like the other cells?

keileg:
So, each cell must have each of its faces specified by its nodes (with an ordering of the nodes). Since the polyhedron can have any number of faces, which again have any number of nodes, you need more than a specific ordering of the nodes in a cell-node relation. As an example, in a pyramid the information needed to define the square at the base is different than the other 4 sides (and it would change again if the base was split into two triangles). 

The polygon is different since it is 2d, thus the faces can be assumed to be defined by subsequent nodes in the cell-node relation.

I hope that was an answer to the question.

nschloe:
Alright, yes. Still: Why don't we put this exact infromation in a `CellBlock`?

keileg:
If that does not break with what you would assume about the content in a CellBlock, I agree it is much better. I kind of assumed that the CellBlock content should be a standard numpy array. 

Will this require modification of export filters where polyhedrons are invalid?

nschloe:
> If that does not break with what you would assume about the content in a CellBlock, I agree it is much better. I kind of assumed that the CellBlock content should be a standard numpy array.

I'm not too sure about it myself but I think it should be alright. Try it out.

> Will this require modification of export filters where polyhedrons are invalid?

What do you mean when you say "filter"?

meshio/vtu/_vtu.py
keileg: It is part of the vtu data format. The comment was more useful a few lines down, so I'll move it.

test/helpers.py
keileg: Corrected.

keileg:
Pushed a version with storage in CellBlocks. It works. The price for not having a separate variable for the face-nodes is special treatment of mesh.cells whenever there is the possibility that a polyhedron is present in the mesh. For instance in test.helpers.read_write():

https://github.com/keileg/meshio/blob/55501aceddf2bae7ab5ae7aef826166342bf778f/test/helpers.py#L349-L358

Similarly, if-elses are needed at a few places in _vtu.py, and will likely be needed as other mesh formats are extended to cover polyhedra.

I'm not sure myself which version I would prefer, but this decision clearly requires wider perspectives than mine.

meshio/vtu/_vtu.py
```diff
@@ -75,14 +152,25 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
                     start_cn[items + 1],
                     _vtk_to_meshio_order(types[start], sz, dtype=offsets.dtype) - sz,
                 )
-                cells.append(CellBlock(meshio_type + str(sz), connectivity[indices]))
+                if meshio_type == "polyhedron":
```
nschloe, line 155: This could be a bit shortened with
```
data = polyhedron_faces[f"polyhedron{sz}"] if meshio_type == "polyhedron" else connectivity[indices]
cells.append(CellBlock(meshio_type + str(sz), data)
```

meshio/vtu/_vtu.py
```diff
@@ -526,6 +634,13 @@ def write(filename, mesh, binary=True, compression="zlib", header_type=None):
     # very memory hungry. See <https://stackoverflow.com/q/59272477/353337>.
     from .._cxml import etree as ET
 
+    # Check if the mesh contains polyhedral cells, this will require special treatment
+    # in certain places.
+    is_polyhedron_grid = False
+    for c in mesh.cells:
+        if c.type[:10] == "polyhedron":
+            is_polyhedron_grid = True
```
nschloe, line 642: `break`?

meshio/vtu/_vtu.py
```diff
@@ -43,16 +43,93 @@ def _cells_from_data(connectivity, offsets, types, cell_data_raw):
...
+                # num_faces_cell_0,
+                #   num_nodes_face_0, node_ind_0, node_ind_1, ..
+                #   num_nodes_face_1, node_ind_0, node_ind_1, ..
+                #   ...
+                # num_faces_cell_1,
+                #   ...
+                # See https://vtk.org/Wiki/VTK/Polyhedron_Support for more.
+
+                # There is a tacit assumption here that the cell-data keys 'faces'
+                # and 'faceoffsets' are reserved for this purpose.
```
nschloe, line 71: Wait, polyhedra are not part of the standard?

nschloe:
Can you append a file that contains polyhedral data as described in this PR?

meshio/vtu/_vtu.py
keileg: Sure.

keileg:
A splitting of the unit cube is attached (same as that generated by helpers.polyhedron_mesh). The file extension is changed to .txt to be compatible with what GitHub supports.


[polyhedron_example.txt](https://github.com/nschloe/meshio/files/5325918/polyhedron_example.txt)

meshio/vtu/_vtu.py
keileg: It is a part of the vtu format and has been for (at least) a few years. However, as in this PR, the 'anything goes'-property of polyhedra calls for a specialized data forma

nschloe: I cannot find the word "polyhedron" in [the spec](https://vtk.org/wp-content/uploads/2015/04/file-formats.pdf).

nschloe: The link you posted is about the VTK software package, not the file formats, isn't it?

keileg: Yes, I believe so

nschloe: Okay. And did I understand correctly that `faces` and `faceoffsets` are cell data (connected to which cell?) that contain polyhedral info?

keileg: Yes. The data for cells are merged, so that `faces` contains datafor all polyhedral cells, `faceoffsets` gives the end of the data for each cell.

nschloe: I don't get it. Can we have a video call perhaps? I think we can better talk about the two PRs that way. Can you get back to me at nico.schloemer@gmail.com?

PR #918, opened 2020-09-22
factor out _read_end_block in all Gmsh MSH

gdmcbain:
In #914, new code was merged to fast-forward to `$EndElements` in MSH 2.2, but it's a general feature of all Gmsh MSH formats, so here it's factored into a module function `gmsh.common._read_end_block` and used everywhere.

There was some inconsistency in whether failure to find the required ending raised a `ReadError`; here I've settled on warning but if its thought that being strict is more helpful, I don't mind that being enforced instead.

I didn't look up nschloe/pygalmesh#34 and the `UnicodeDecodeError` but just applied that everywhere too.


nschloe:
Let's call the method _fast_forward_to_end_block or something like this.

nschloe:
Thank you!

Issue #919, opened 2020-09-22
Support for *include statements in the abaqus input format


Krande:
Hi,

I noticed some of my abaqus input files weren't converted by meshio. If I am not mistaken, currently there is no support for "*include" statements (which is used to import content from other input files) in the abaqus reader?

Therefore I created a suggested edit to the abaqus reader here https://github.com/Krande/meshio/commit/d24cd85993d8f52b362e94926e3ca97bebe06321 which also includes two example abaqus input files containing the *include statement. 

Let me know if this functionality already exists (but I am not seeing it\using it correctly), or if more people would be interested in the same functionality I can create a pull request!

Best Regards
Kristoffer

nschloe:
Always PR, then we'll have something to talk about.

PR #920, opened 2020-09-22
Add support for "*include" statements in abaqus input files

Krande:
As mentioned in #919, here is a pull request for supporting *include statements in abaqus input files

### DIFF
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -1,12 +1,9 @@
 """
 I/O for Abaqus inp files.
 """
-import os
-import re
-import io
+from pathlib import Path
 
 import numpy
-
 
 from ..__about__ import __version__
 from .._exceptions import ReadError
@@ -97,33 +94,8 @@ abaqus_to_meshio_type = {
 meshio_to_abaqus_type = {v: k for k, v in abaqus_to_meshio_type.items()}
 
 
-def bulk_w_includes(inp_path, bulk_str):
-    re_in = re.compile(
-        r"\*Include,\s*input=(.*?)$", re.IGNORECASE | re.MULTILINE | re.DOTALL
-    )
-    bulk_repl = dict()
-
-    for m in re_in.finditer(bulk_str):
-        search_key = m.group(0)
-        with open(os.path.join(os.path.dirname(inp_path), m.group(1)), "r") as d:
-            bulk_repl[search_key] = d.read()
-
-    for key, val in bulk_repl.items():
-        bulk_str = bulk_str.replace(key, val)
-    return bulk_str
-
-
 def read(filename):
     """Reads a Abaqus inp file."""
-
-    out = None
-    with open_file(filename, "r") as f:
-        bulk_str = f.read()
-        if "*include" in bulk_str.lower():
-            out = read_buffer(io.StringIO(bulk_w_includes(filename, bulk_str)))
-
-    if out is not None:
-        return out
 
     with open_file(filename, "r") as f:
         out = read_buffer(f)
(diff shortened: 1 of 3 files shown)


nschloe:
Run `make format`.

meshio/abaqus/_abaqus.py
```diff
@@ -92,8 +92,39 @@
 meshio_to_abaqus_type = {v: k for k, v in abaqus_to_meshio_type.items()}
 
 
+def bulk_w_includes(inp_path, bulk_str):
+    import os
```
nschloe, line 96: All imports, system imports at least, do on the top of the file.

meshio/abaqus/_abaqus.py
```diff
@@ -92,8 +92,39 @@
...
+    import re
+
+    re_in = re.compile(
+        r"\*Include,\s*input=(.*?)$", re.IGNORECASE | re.MULTILINE | re.DOTALL
+    )
+    bulk_repl = dict()
+
+    for m in re_in.finditer(bulk_str):
+        search_key = m.group(0)
+        with open(os.path.join(os.path.dirname(inp_path), m.group(1)), "r") as d:
```
nschloe, line 106: Better use pathlib.Path.

meshio/abaqus/_abaqus.py
```diff
@@ -92,8 +92,39 @@
...
+        bulk_str = bulk_str.replace(key, val)
+    return bulk_str
+
+
 def read(filename):
     """Reads a Abaqus inp file."""
+
+    out = None
+    with open_file(filename, "r") as f:
+        bulk_str = f.read()
```
nschloe, line 119: This is a bit much. You read the entire file, search if for `*include`, and if found, you do something else, right?

meshio/abaqus/_abaqus.py
Krande: Sure! Pathlib is new to me. Would something like this be okay? 

```python
from pathlib import Path
str(Path(inp_path, m.group(1)).absolute())
```

nschloe: No need for `str`. Also, use `/` for joining the paths.

meshio/abaqus/_abaqus.py
Krande: That's correct. I am definitively open for suggestions! Maybe, "bulk_w_include" should be kicked off only IF you encounter a line containing *include flag when using the existing readbuffer? And instead of reading the entire file in memory you do a new read_buffer on the referenced file and appending it contents with the existing read_buffer stuff? I think that would remove any need to read the entire file in memory.

nschloe: Sounds better. The question is where in an abaqus file you can `include` things. Could there be an abaqus file that lists some data explicitly, and other data is included from other files?

Krande: I think *include can be used everywhere (also in included files). So perhaps some use of recursive function would be helpful? I can take a stab at it and see what I come up with?

nschloe: Yeah, it should be recursive somehow then. Sounds like a bigger change, do you want to work on it and submit another PR?

Krande: I hacked away and made something that worked with sample files I tested it on locally, but I see that it fails on the internal checks. I might have been a bit premature when pushing that last commit. I will give it a bit more time, and try to come up with a better solution. If you want, feel free to close this PR and I'll just create a new one once I have something a bit more robust. Ideally I do not want to touch the original implementation.

nschloe: Alright, let's close this then. Open a new PR when ready.

PR #921, opened 2020-09-24
Another attempt at adding support for "*include" statements in abaqus

Krande:
Here is a redo of pull request #920. Now I have left the original implementation as is for input files not containing the *include flag. 

The current implementation does however still read the files in memory and assembles the included files into 1 input file before it parses it back into the read_buffer using StringIO. So its an improvement over my last suggestion (as now it does not break anything), but not as efficient as it probably could be.

### DIFF
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -1,6 +1,10 @@
 """
 I/O for Abaqus inp files.
 """
+import io
+import os
+import re
+
 import numpy
 
 from ..__about__ import __version__
@@ -92,6 +96,23 @@ abaqus_to_meshio_type = {
 meshio_to_abaqus_type = {v: k for k, v in abaqus_to_meshio_type.items()}
 
 
+def read_w_includes(inp_path):
+    re_in = re.compile(r'\*Include,\s*input=(.*?)$', re.IGNORECASE | re.MULTILINE | re.DOTALL)
+    bulk_repl = dict()
+    with open_file(inp_path, "r") as f:
+        bulk_str = f.read()
+
+    for m in re_in.finditer(bulk_str):
+        search_key = m.group(0)
+        with open(os.path.join(os.path.dirname(inp_path), m.group(1)), 'r') as d:
+            bulk_repl[search_key] = d.read()
+
+    for key, val in bulk_repl.items():
+        bulk_str = bulk_str.replace(key, val)
+
+    return read_buffer(io.StringIO(bulk_str))
+
+
 def read(filename):
     """Reads a Abaqus inp file."""
     with open_file(filename, "r") as f:
@@ -165,6 +186,8 @@ def read_buffer(f):
                         cell_sets[name].append(cell_sets_element[set_name])
                     else:
                         raise ReadError(f"Unknown cell set '{set_name}'")
+        elif keyword == 'INCLUDE':
+            return read_w_includes(f.name)
         else:
             # There are just too many Abaqus keywords to explicitly skip them.
             line = f.readline()
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -2,8 +2,8 @@
 I/O for Abaqus inp files.
 """
 import io
-import re
 import pathlib
+import re
 
 import numpy
 
@@ -97,19 +97,21 @@ meshio_to_abaqus_type = {v: k for k, v in abaqus_to_meshio_type.items()}
 
 
 def read_w_includes(inp_path):
-    re_in = re.compile(r'\*Include,\s*input=(.*?)$', re.IGNORECASE | re.MULTILINE | re.DOTALL)
+    re_in = re.compile(
+        r"\*Include,\s*input=(.*?)$", re.IGNORECASE | re.MULTILINE | re.DOTALL
+    )
     bulk_repl = dict()
     with open_file(inp_path, "r") as f:
         bulk_str = f.read()
 
     for m in re_in.finditer(bulk_str):
         search_key = m.group(0)
-        incl_ref = m.group(1).replace('\\', '/')
+        incl_ref = m.group(1).replace("\\", "/")
         # The include ref can be absolute or relative to .inp file. Will check for both
         if pathlib.Path(incl_ref).exists() is False:
             cd = pathlib.Path(inp_path).parents[0]
             incl_ref = cd / incl_ref
-        with open(pathlib.Path(incl_ref).absolute(), 'r') as d:
+        with open(pathlib.Path(incl_ref).absolute(), "r") as d:
             bulk_repl[search_key] = d.read()
 
     for key, val in bulk_repl.items():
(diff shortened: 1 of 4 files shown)


nschloe:
The `bulk_str = f.read()` is not good.

Krande:
I agree with you. Okay how about this alternative. I create a new generator object for the included file and store the original generator in a temp variable. On end of file I check if there is stored a parent generator. Maybe to ensure that it can handle multi-layer *include (file in file in file - a bit inception-esque) the temp variable should instead be a list, and retrived last entry and pop.

Note! I see the current test is failing. I will fix that

meshio/abaqus/_abaqus.py
```diff
@@ -165,6 +177,17 @@ def read_buffer(f):
                         cell_sets[name].append(cell_sets_element[set_name])
                     else:
                         raise ReadError(f"Unknown cell set '{set_name}'")
+        elif keyword == "INCLUDE":
+            # Get absolute path to referred input file
+            incl_ref = line.split("=")[-1].rstrip().replace("\\", "/")
```
nschloe, line 182: No idea what this all means. You can simply convert the string to a Path and do your business there, will be much easier.

meshio/abaqus/_abaqus.py
```diff
@@ -111,10 +113,20 @@ def read_buffer(f):
...
     point_data = {}
 
+    parent_file_gen = None
+
     line = f.readline()
     while True:
         if not line:  # EOF
-            break
+            if parent_file_gen is not None:
+                f.close()
```
nschloe, line 122: This part is a bit mysterious. Why do we close the file? Then after that, `f.readline()`? Something is fishy...

Krande: I'll admit it does get a bit fishy. Inside the keyword=INCLUDE condition I store the current file generator inside the "parent_file_gen" variable and replace the variable "f" with a generator of the external file (as shown below).
 https://github.com/nschloe/meshio/blob/5e8f401868691e9215b18350ca9e27e947867a6c/meshio/abaqus/_abaqus.py#L180-L190

What the change in the EOF eval does is that it checks if there is stored anything in the temp "parent_file_gen" variable. If it is, then it will close the current generator (which is the external file), and assigns back the original generator to the variable "f" and moves beyond the "include" line with a new readline() statement.

https://github.com/nschloe/meshio/blob/5e8f401868691e9215b18350ca9e27e947867a6c/meshio/abaqus/_abaqus.py#L120-L129

I agree it is not pretty replacing variables like this, but for now I cant see any real alternatives that won't need changing other parts of the read_buffer potentially breaking dependencies from other readers (which was the case during my last attempt). Anyways, let me know if you find this acceptable, or you have some suggestions for any alternative implementations that you might want me to explore. Otherwise I guess you can close this pull request for now.

meshio/abaqus/_abaqus.py
Krande: Ah, yes that did get messy. How about ```pathlib.Path(line.split("=")[-1])```?

nschloe: Better, but I still don't get what's up with the equality sign.

meshio/abaqus/_abaqus.py
nschloe: Nope, too messy. Why can't we do something like the following? (Actual question.)
```python
_read(filename):
   # ...
   elif keyword == "INCLUDE": 
        include_filename = # read line and get filename
        out = _read(include_filename)
        # merge `out` into the mesh
```

meshio/abaqus/_abaqus.py
Krande: Okay, well the "include" statement in abaqus is `*INCLUDE,INPUT=wInclude_bulk.inp`. I just split the line on the equal sign to create a list and use the last entry to get the path to the external input file. There are probably a bunch of alternative ways of doing this, but this was just the first that popped into my head :)

nschloe: No it's fine, I just didn't know the syntax. You can best add a code comment here with this example.

Krande: Yup, will do.

meshio/abaqus/_abaqus.py
Krande: A good question. 

That was actually my ambition in my previous pull request. But unfortunately I made it (perhaps) unecessary complex by trying to account for points being defined within the external file AND the current file (which in hindsight would probably only apply when the model is using a "parts and assembly" model definition and\or multiple parts, and I am unsure if this is supported by meshio?). 

See my previous attempt (https://github.com/Krande/meshio/blob/35c012bdecf56d5e045bf2b5ef775ac3d1e5473b/meshio/abaqus/_abaqus.py#L180-L197) (excerpts shown below) where I ended up having to add an initial decleration of "points" variable as an empty numpy array and concatenate any points found in the current and external file. In the process I also think I messed up the "point_ids_in" also somehow (but on closer inspection, that variable might not be used anywhere else?). 

```python
def read_buffer(f):
    # Initialize the optional data fields
    points = numpy.empty([0, 3], dtype=numpy.float64)
    ...
    while True:
        ...
        elif keyword == "INCLUDE":
            incl_ref = line.lower().split("input=")[-1].rstrip().replace("\\", "/")
            if Path(incl_ref).exists() is False:
                cd = Path(f.name).parents[0]
                incl_ref = Path(cd, incl_ref).absolute()
            with open(incl_ref, "r") as d:
                submsh = read_buffer(d)


            points = numpy.concatenate((points, submsh.points))
            cells += submsh.cells
            point_data.update(submsh.point_data)
            cell_data.update(submsh.cell_data)
            field_data.update(submsh.field_data)
            point_sets.update(submsh.point_sets)
            cell_sets.update(submsh.cell_sets)


            line = f.readline()
        else:
```

I guess I could make it work by assuming that the points are either declared in the current file or the external file by simply doing something like `points = submsh.points`. But I might have to study the internal workings of the Mesh class first as to not break any potential inter-dependencies of list indices between the various lists and dicts.

Let me know what you think and if you have some suggestions while I contemplate and study the Mesh class :)

nschloe: This is the way it should work, except that you don't `concat` in each step, but just
```python
points = []
for [...]:
    points.append(more_points)
points = numpy.concatenate(points)
```
Or even
```
meshes = []
for [...]:
    meshes.append(new_mesh)
mesh = merge(meshes)
```
The problem with the previous PR was not this approach, but the fact you read an entire file at once. This is not necessary.

Krande: Thanks a lot for taking the time helping me out! I really appreciate it! Okay, so I've given it another go and I've created a merge function and have started combining the meshes found in the external files with the "main" mesh containers in the read_buffer. I still have to figure out the cell_sets and CellBlock relationship and how to safely merge cell_sets (that was a bit more complex than I anticipated). 

Hopefully I will find some time in the next couple of days to fix that. In the mean time I thought it would be a good idea to push what I have so far. Let me know If you have any comments!

nschloe:
Looks good now, thanks for the contribution!

PR #922, opened 2020-09-24
Export gmsh entities

keileg:
## Summary
Expansions of the gmsh filter (only format _41):
* Gmsh entity tags are read and written. This necessitated storage of entity tags for nodes and cells.
* With this extension, it is also possible to write .msh files with more than one cell block, provided the necessary entity tags are present in the Mesh. The existing functionality (deal with a single cell block, no need to care about tags) is preserved as an alternative.
* Boundary entities (which 0d entities form the boundary of a 1d entity etc.) can now be read and written.

## Testing
I have not written any tests for the added functionality: This would require tailored setups, which breaks with the standard approach to meshio testing (simple setups, throw a lot of grids on it). I will of course write tests if requested, but I would welcome some guidance on how to set up such tests. 

For clarity, the testing during development focused on a .msh files with somewhat general combinations of entities in different dimensions, different physical objects etc. I can of course share this for use as a reference file in testing, if you are interested.

### DIFF
--- a/meshio/gmsh/_gmsh41.py
+++ b/meshio/gmsh/_gmsh41.py
@@ -351,9 +351,10 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
         # dim_tag information to two arrays (n x 1). The data in gmsh:dim_tag is then
         # temporarily removed from point_data, and restored to Mesh when writing
         # is done.
-        dim_tags = mesh.point_data.pop("gmsh:dim_tags", None)
-        if dim_tags is not None:
+
+        if "gmsh:dim_tags" in mesh.point_data:
             # Temporarily store the data as 1d arrays
+            dim_tags = mesh.point_data.pop("gmsh:dim_tags", None)
             mesh.point_data["node_entity_dimension"] = dim_tags[:, 0]
             mesh.point_data["point_entity"] = dim_tags[:, 1]
 
@@ -368,7 +369,7 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
             _write_data(fh, "ElementData", name, dat, binary)
 
         # Restore dim_tag information if available
-        if dim_tags is not None:
+        if "point_entity" in mesh.point_data:
             mesh.point_data["gmsh:dim_tags"] = dim_tags
             # Also remove the temporary 1d arrays
             mesh.point_data.pop("node_entity_dimension")
@@ -425,7 +426,7 @@ def _write_entities(fh, cells, cell_data, cell_sets, point_data, binary):
     # The entities section must therefore be built on the node-entities, if
     # these are available. If this is not the case, we leave this section blank.
     # TODO: Should this give a warning?
-    if "gmsh:dim_tags" not in point_data.keys():
+    if "gmsh:dim_tags" not in point_data:
         return
 
     fh.write(b"$Entities\n")
@@ -454,20 +455,16 @@ def _write_entities(fh, cells, cell_data, cell_sets, point_data, binary):
         ]
 
     # We will only deal with bounding entities if this information is available
-    has_bounding_elements = "gmsh:bounding_entities" in cell_sets.keys()
+    has_bounding_elements = "gmsh:bounding_entities" in cell_sets
 
     # The node entities form a superset of cell entities. Write entity information
     # based on nodes, supplement with cell information when there is a matcihng
     # cell block.
-    for j in range(node_dim_tags.shape[0]):
-        dim, tag = node_dim_tags[j]
+    for dim, tag in node_dim_tags:
 
         # Find the matching cell block, if it exists
         matching_cell_block = numpy.where(
-            numpy.logical_and(
-                cell_dim_tags[:, 0] == node_dim_tags[j, 0],
-                cell_dim_tags[:, 1] == node_dim_tags[j, 1],
-            )
+            numpy.logical_and(cell_dim_tags[:, 0] == dim, cell_dim_tags[:, 1] == tag)
         )[0]
 
         # The information to be written varies according to entity dimension,
@@ -598,7 +595,7 @@ def _write_nodes(fh, points, cells, point_data, float_fmt, binary):
     # If node (entity) tag and dimension is available, we make a list of unique
     # combinations thereof, and a map from the full node set to the unique
     # set.
-    if "gmsh:dim_tags" in point_data.keys():
+    if "gmsh:dim_tags" in point_data:
         # reverse_index_map maps from all nodes to their respective representation
         # in (the uniquified) node_dim_tags. This approach works for general
         # orderings of the nodes
@@ -727,7 +724,7 @@ def _write_elements(fh, cells, cell_data, binary):
             # entityDim(int) entityTag(int) elementType(int) numElementsBlock(size_t)
             dim = _geometric_dimension[cell_type]
             # The entity tag should be equal within a CellBlock
-            if "gmsh:geometrical" in cell_data.keys():
+            if "gmsh:geometrical" in cell_data:
                 entity_tag = cell_data["gmsh:geometrical"][ci][0]
             else:
                 entity_tag = 0
--- a/meshio/gmsh/_gmsh41.py
+++ b/meshio/gmsh/_gmsh41.py
@@ -441,6 +441,9 @@ def _write_entities(fh, cells, cell_data, cell_sets, point_data, binary):
 
     # Write number of entities per dimension
     num_occ = numpy.bincount(node_dim_tags[:, 0], minlength=4)
+    if num_occ.size > 4:
+        raise ValueError("Encountered entity with dimension > 3")
+
     if binary:
         num_occ.astype(c_size_t).tofile(fh)
     else:
@@ -469,6 +472,10 @@ def _write_entities(fh, cells, cell_data, cell_sets, point_data, binary):
         matching_cell_block = numpy.where(
             numpy.logical_and(cell_dim_tags[:, 0] == dim, cell_dim_tags[:, 1] == tag)
         )[0]
+        if matching_cell_block.size > 1:
+            # It is not 100% clear if this is not permissible, but the current
+            # implementation for sure does not allow it.
+            raise ValueError("Encountered non-unique CellBlock dim_tag")
 
         # The information to be written varies according to entity dimension,
         # whether entity has a physical tag, and between ascii and binary.
--- a/meshio/gmsh/_gmsh41.py
+++ b/meshio/gmsh/_gmsh41.py
@@ -101,7 +101,7 @@ def read_buffer(f, is_ascii, data_size):
     cell_data.update(cell_tags)
 
     # Add node entity information to the point data
-    point_data.update(point_entities)
+    point_data.update({"gmsh:dim_tags": point_entities})
 
     return Mesh(
         points,
@@ -192,14 +192,10 @@ def _read_nodes(f, is_ascii, data_size):
 
     _fast_forward_to_end_block(f, "Nodes")
 
-    # Entity information for nodes is stored as point data, with clearly named
-    # items.
-    point_data = {
-        "gmsh:node_entity": entity_of_nodes,
-        "gmsh:node_entity_dimension": entity_dimension,
-    }
+    # Entity information for nodes is stored as point data, as an nx2 array
+    dim_tags = numpy.vstack((entity_dimension, entity_of_nodes)).T
 
-    return points, tags, point_data
+    return points, tags, dim_tags
 
 
 def _read_elements(
@@ -219,16 +215,15 @@ def _read_elements(
 
     for k in range(num_entity_blocks):
         # entityDim(int) entityTag(int) elementType(int) numElements(size_t)
-        dim_entity, tag_entity, type_ele = fromfile(f, c_int, 3)
+        dim, tag, type_ele = fromfile(f, c_int, 3)
         (num_ele,) = fromfile(f, c_size_t, 1)
         for physical_name, cell_set in cell_sets.items():
             cell_set[k] = numpy.arange(
                 num_ele
                 if (
                     physical_tags
-                    and field_data[physical_name][1] == dim_entity
-                    and field_data[physical_name][0]
-                    in physical_tags[dim_entity][tag_entity]
+                    and field_data[physical_name][1] == dim
+                    and field_data[physical_name][0] in physical_tags[dim][tag]
                 )
                 else 0,
                 dtype=type(num_ele),
@@ -240,13 +235,13 @@ def _read_elements(
         )
 
         # Find physical tag, if defined; else it is None.
-        pt = None if not physical_tags else physical_tags[dim_entity][tag_entity]
+        pt = None if not physical_tags else physical_tags[dim][tag]
         # Bounding entities (of lower dimension) if defined. Else it is None.
-        if dim_entity > 0 and bounding_entities:  # Points have no boundaries
-            be = bounding_entities[dim_entity][tag_entity]
+        if dim > 0 and bounding_entities:  # Points have no boundaries
+            be = bounding_entities[dim][tag]
         else:
             be = None
-        data.append((pt, be, tag_entity, tpe, d))
+        data.append((pt, be, tag, tpe, d))
 
     _fast_forward_to_end_block(f, "Elements")
 
@@ -351,6 +346,20 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
             fh, cells, mesh.cell_data, mesh.cell_sets, mesh.point_data, binary
         )
         _write_nodes(fh, mesh.points, mesh.cells, mesh.point_data, float_fmt, binary)
+
+        # IMPLEMENTATION NOTE: If dimtag information is available (key gmsh:dim_tags)
+        # this is stored in meshio as an n x 2 array. Gmsh accepts only arrays of size
+        # 1, 3 or 9, and _write_data() will raise an error if point_data with the
+        # wrong size is encountered. As a workaround, _write_nodes() converts the
+        # dim_tag information to two arrays (n x 1). The data in gmsh:dim_tag is then
+        # temporarily removed from point_data, and restored to Mesh when writing
+        # is done.
+        dim_tags = mesh.point_data.pop("gmsh:dim_tags", None)
+        if dim_tags is not None:
+            # Temporarily store the data as 1d arrays
+            mesh.point_data["node_entity_dimension"] = dim_tags[:, 0]
+            mesh.point_data["point_entity"] = dim_tags[:, 1]
+
         _write_elements(fh, cells, mesh.cell_data, binary)
         if mesh.gmsh_periodic is not None:
             _write_periodic(fh, mesh.gmsh_periodic, float_fmt, binary)
@@ -360,6 +369,13 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
         cell_data_raw = raw_from_cell_data(mesh.cell_data)
         for name, dat in cell_data_raw.items():
             _write_data(fh, "ElementData", name, dat, binary)
+
+        # Restore dim_tag information if available
+        if dim_tags is not None:
+            mesh.point_data["gmsh:dim_tags"] = dim_tags
+            # Also remove the temporary 1d arrays
+            mesh.point_data.pop("node_entity_dimension")
+            mesh.point_data.pop("point_entity")
 
 
 def _write_entities(fh, cells, cell_data, cell_sets, point_data, binary):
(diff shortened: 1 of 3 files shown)


nschloe:
Ping me once the tests are passing.

keileg:
@nschloe Now the tests running. Additional changes:

* In the test that read a reference .msh file in 4.1 format, I had to introduce a separate reference number of cells for the cell_sets; since the bounding entities are of the cell blocks are stored in this field.
* The same test is extended to do a read-write consistency check. This is possible since the .msh writer now handle meshes with multiple cell blocks. This alleviates much of my previous concerns regarding the missing testing - let me know if you want more tests.

keileg:
What do you think of expanding `test.helpers.write_read()` with an additional, optional, argument `additional_tests`? This would in itself be a function which takes the input and output mesh; the idea is that the function would do tests tailored for the particular mesh (type). This would be a way to test the consistency of non-standard mesh-information, without polluting `write_read()` with various conditionals on mesh type, present fields etc. The additional_tests function would have to be defined in the specific test function, but that should localize the entropy fairly well.

For example, the gmsh:entity information introduced in this PR could be tested by such a function (in particular, I'm thinking of the test on the reference mesh), as could the polyhedral_faces to be introduced in #916. 

Let me know what you think.

nschloe:
> let me know if you want more tests.

Is all the entity reading/writing covered by this?

nschloe:
> What do you think of expanding test.helpers.write_read() with an additional, optional, argument additional_tests? 

The user interface and the test suite should be separate, so I'd be pretty reserved towards this idea. Or do you mean consistency checks in case the user catches a bad mesh from somewhere?

keileg:
Sorry for the poor formulations. I think of enhanced tests only, and certainly no mixing of tests and meshio proper. To give the background, testing of writing of gmsh entities probably should be better; it was while thinking of how to do this I came up with the suggestion outlined last week.

To clarify, what I suggest is to make it easier to use the `test.helpers.write_read` to check mesh information that is particular to certain formats and perhaps mesh types. Specifically, write_read can be modified to

```python
def write_read(writer, reader, input_mesh, atol, extension=".dat", additional_test_function=None):
    # Additional test function has types `Callable[meshio.Mesh, meshio.Mesh] -> None`. 
    # Towards the end of write_read, I would insert
    if additional_test_function is not None: 
        additional_test_function(mesh, input_mesh)
```

Usage of this function would be, illustrated by checking `cell_sets['gmsh:bounding_entities'] introduced in this PR:

```python
@ parameters .. 
def test_reference_file_readonly():
    def helper_test_cell_sets(mesh1, mesh2):
        # Define helper function, in this case it compares cell sets
        for name1, val2 in mesh1.cell_sets.items():
            val2 = mesh2.cell_sets[name1]
            # val1 and val2 are now lists of numpy arrays, one list item per CellBlock
            for arr1, arr2 in zip(val1, val2):
                  asert numpy.allclose(arr1, arr2)
    # ... mesh construction etc.
    write_read(writer,  meshio.gmsh.read, mesh, 1.0e-15, helper_test_cell_sets)
```
Now, in this case, the same effect could have been achieved by adding cell_sets to the data that is to be added in `tests.helpers.write_read()`. However, this presumes that all data in Mesh.cell_sets has the standardized form of `Dict[str, List[numpy.ndarray]]`; it is not clear to me whether this is actually a requirement on data in cell_sets. Moreover, when new information is introduced (e.g. Mesh.polyhedral_faces cf #916), one can currently choose between 
* extend `tests.helpers.write_read` further, with the resulting entropy, 
* write specialized tests that essentially copy write_read(), and then add testing for the specialized fields
* forego checks that the specialized fields are consistently treated by Mesh.write() followed by Mesh.read().

With the above suggestion, the additional testing would be achieved by writing a new tailored helper function, and passing this to write_read. Some of my confusion may be caused by my lacking understanding of tacit assumptions on the format of the different fields in Mesh - for instance, if cell_sets always is a list of numpy arrays, my above example becomes redundant. Nevertheless, to me it appears that the added oportunities could become useful in certain cases.

If you do not like the above idea, I will do so more modification to the testing so that writing of entities is better covered before considering the PR finished.

Sorry about the long answer!

nschloe:
> Nevertheless, to me it appears that the added oportunities could become useful in certain cases.

Let's not over-engineer this. _If_ something comes up in the future, let's to it then. For now, I think it's easiest to keep `test.helpers.write_read` uniform for all meshes. The `helper_test_cell_sets` certainly looks useful and should be included there to test cell sets for all mesh types.

Having one test function for all mesh types also makes sure that the output is the same, no matter if I read/write from a gmsh or a vtk file.

keileg:
Fair enough.

I added the test of cell_sets, and now believe that the added functionality of to gmsh has sufficient testing. However, the enhanced testing uncovered missing cell_sets in other mesh formats (I would say about 10 formats). Thus this turned into a much bigger task than enhancing gmsh, and I do not have the capacity nor the background to patch them all. 

Do you have guidance on how to proceed?

nschloe:
Not all formats have cell sets. Just add a conditional before the test.

keileg:
That explains it all. Thanks.

The test coverage for the new gmsh functionality should be sufficient now.

nschloe:
Good. I'll look at it later this afternoon.

meshio/gmsh/_gmsh41.py
```diff
@@ -62,12 +63,20 @@ def read_buffer(f, is_ascii, data_size):
         if environ == "PhysicalNames":
             _read_physical_names(f, field_data)
         elif environ == "Entities":
-            physical_tags = _read_entities(f, is_ascii, data_size)
+            # Read physical tags and information on bounding entities.
+            # The information is passed to the processing of elements.
+            physical_tags, bounding_entities = _read_entities(f, is_ascii, data_size)
         elif environ == "Nodes":
-            points, point_tags = _read_nodes(f, is_ascii, data_size)
+            points, point_tags, point_entities = _read_nodes(f, is_ascii, data_size)
```
nschloe, line 70: Question: What are `point_entities` in gmsh lingo?

meshio/gmsh/_gmsh41.py
```diff
@@ -103,26 +115,40 @@ def read_buffer(f, is_ascii, data_size):
...
+    # entities of dimension > 0) the bounding entities (so points that form
+    # the boundary of a line etc).
+    # Note that the bounding box of the entities is disregarded. Adding this
+    # is not difficult, but for the moment, the entropy of adding more data
+    # does not seem warranted.
+
     fromfile = partial(numpy.fromfile, sep=" " if is_ascii else "")
     c_size_t = _size_type(data_size)
     physical_tags = tuple({} for _ in range(4))  # dims 0, 1, 2, 3
+    bounding_entities = tuple({} for _ in range(4))
```
nschloe, line 128: Can be have `({}, {}, {}, {})` here and in the line above?

meshio/gmsh/_gmsh41.py
```diff
@@ -155,13 +184,27 @@ def _read_nodes(f, is_ascii, data_size):
         # Store the point densely and in the order in which they appear in the file.
         # x(double) y(double) z(double) (* numNodes)
         points[ixx] = fromfile(f, c_double, num_nodes * 3).reshape((num_nodes, 3))
+
+        # Entity tag and entity dimension of the nodes. Stored as point-data.
+        entity_of_nodes[ixx] = entity_tag
+        entity_dimension[ixx] = dim
```
nschloe, line 190: Isn't this always 0?

meshio/gmsh/_gmsh41.py
keileg: From what I understand, Gmsh keeps track of which entities (geometric objects in the .geo file) points and cells have been assigned to:
* in the $Nodes section, the nodes are grouped into these entities. Since the entity numbering starts from 1 in each dimension, the dimension of the entity of the node must be stored. 
* the $Elements section is similar, but it contains only entities for which cells were generated (e.g. only those decleared Physical in gmsh lingo). Thus the entity information for cells is a subset of the full entity set.

The $Entities section is (I believe) generated from the entities of nodes, with physical tags and boundary representation from the cells. I have no idea how all this is handled internally in gmsh, but when I tried to get away with preserving less information, gmsh could not process the file.

meshio/gmsh/_gmsh41.py
keileg: Sure. Will push.

meshio/gmsh/_gmsh41.py
keileg: See comment above regarding entities in general.

nschloe: Gmsh calls this `dim_tag`, let's follow suit with a `dim_tag` array, not separate `entity_of_nodes` and `entity_dimension`.

meshio/gmsh/_gmsh41.py
```diff
@@ -332,24 +389,157 @@ def _write_entities(fh, cells, binary):
...
+    # Both nodes and cells have enities, but the cell entities are a subset of
+    # the nodes. The reason is (if the inner workings of Gmsh has been correctly
+    # understood) that node entities are assigned to all
+    # objects necessary to specify the geometry whereas only cells of Physical
+    # objcets (gmsh jargon) are present among the cell entities.
+    # The entities section must therefore be built on the node-entities, if
+    # these are available. If this is not the case, we leave this section blank.
+    # TODO: Should this give a warning?
+    if not (
+        "gmsh:node_entity_dimension" in point_data.keys()
```
nschloe, line 402: Yeah, `dim_tags` again.

meshio/gmsh/_gmsh41.py
```diff
@@ -155,13 +184,27 @@ def _read_nodes(f, is_ascii, data_size):
...
+        entity_of_nodes[ixx] = entity_tag
+        entity_dimension[ixx] = dim
         idx += num_nodes
 
     _fast_forward_to_end_block(f, "Nodes")
-    return points, tags
 
+    # Entity information for nodes is stored as point data, with clearly named
+    # items.
+    point_data = {
```
nschloe, line 197: No need for a dict here, just make it `dim_tags`.

meshio/gmsh/_gmsh41.py
keileg: Agreed, this would be cleaner. However, storage as point data is then not permissible, since 
gmsh only accepts 1, 3 or 9 components per data fields, cf `meshio.gmsh.common._write_data()` L244. Should I store it in Mesh.point_sets instead? 

Sorry for the trouble; it is a bit hard to understand what is the intended usage for the different kinds of variables, although I appreciate the flexibility that comes with not specifying this clearly.

nschloe: Well, how do you write `entity_of_nodes` and `entity_dimension` now?

keileg: They are two different arrays (because of the different keys) with 1 component each. Thus no problem

meshio/gmsh/_gmsh41.py
```diff
@@ -366,34 +556,86 @@ def _write_nodes(fh, points, cells, float_fmt, binary):
...
-        numpy.array([n], dtype=c_size_t).tofile(fh)
-        numpy.arange(1, 1 + n, dtype=c_size_t).tofile(fh)
-        points.tofile(fh)
-        fh.write(b"\n")
     else:
         fh.write(f"{num_blocks} {n} {min_tag} {max_tag}\n".encode("utf-8"))
-        fh.write(f"{dim_entity} {entity_tag} {is_parametric} {n}\n".encode("utf-8"))
-        numpy.arange(1, 1 + n, dtype=c_size_t).tofile(fh, "\n", "%d")
+
+    for entity in range(num_blocks):
```
nschloe, line 608: iterables are always `k`, `i`, `j`.

keileg:
I implemented the changes we discussed to this PR. It looks cleaner now.

meshio/gmsh/_gmsh41.py
```diff
@@ -155,13 +184,23 @@ def _read_nodes(f, is_ascii, data_size):
...
+        # Entity tag and entity dimension of the nodes. Stored as point-data.
+        entity_of_nodes[ixx] = entity_tag
+        entity_dimension[ixx] = dim
         idx += num_nodes
 
     _fast_forward_to_end_block(f, "Nodes")
-    return points, tags
 
+    # Entity information for nodes is stored as point data, as an nx2 array
+    dim_tags = numpy.vstack((entity_dimension, entity_of_nodes)).T
```
nschloe, line 196: Even easier: `numpy.array([entity_dimension, entity_dimension])`. This way, the data remains contiguous in memory, too. And even better: Instead of `entity_of_nodes` and `entity_dimension`, instantiate an n-by-2 array `dim_tags` right at the start.

keileg: Agree. pushed.

meshio/gmsh/_gmsh41.py
```diff
@@ -292,9 +339,25 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+
+        # IMPLEMENTATION NOTE: If dimtag information is available (key gmsh:dim_tags)
+        # this is stored in meshio as an n x 2 array. Gmsh accepts only arrays of size
+        # 1, 3 or 9, and _write_data() will raise an error if point_data with the
+        # wrong size is encountered. As a workaround, _write_nodes() converts the
+        # dim_tag information to two arrays (n x 1). The data in gmsh:dim_tag is then
+        # temporarily removed from point_data, and restored to Mesh when writing
+        # is done.
+        dim_tags = mesh.point_data.pop("gmsh:dim_tags", None)
+        if dim_tags is not None:
```
nschloe, line 355: Better: `if "gmsh:dim_tags" in mesh.point_data: [...]`.

meshio/gmsh/_gmsh41.py
```diff
@@ -304,52 +367,214 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+
+    # Both nodes and cells have enities, but the cell entities are a subset of
+    # the nodes. The reason is (if the inner workings of Gmsh has been correctly
+    # understood) that node entities are assigned to all
+    # objects necessary to specify the geometry whereas only cells of Physical
+    # objcets (gmsh jargon) are present among the cell entities.
+    # The entities section must therefore be built on the node-entities, if
+    # these are available. If this is not the case, we leave this section blank.
+    # TODO: Should this give a warning?
+    if "gmsh:dim_tags" not in point_data.keys():
```
nschloe, line 428: `.keys()` is not necessary

meshio/gmsh/_gmsh41.py
```diff
@@ -304,52 +367,214 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+    # similar not array.
+    cell_dim_tags = numpy.empty((len(cells), 2), dtype=int)
+    for ci in range(len(cells)):
+        cell_dim_tags[ci] = [
+            _geometric_dimension[cells[ci][0]],
+            cell_data["gmsh:geometrical"][ci][0],
+        ]
+
+    # We will only deal with bounding entities if this information is available
+    has_bounding_elements = "gmsh:bounding_entities" in cell_sets.keys()
```
nschloe, line 457: `keys()` not necessary

meshio/gmsh/_gmsh41.py
```diff
@@ -304,52 +367,214 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+
+    # The node entities form a superset of cell entities. Write entity information
+    # based on nodes, supplement with cell information when there is a matcihng
+    # cell block.
+    for j in range(node_dim_tags.shape[0]):
```
nschloe, line 462: Better: `for dim, tag in node_dim_tags:`

meshio/gmsh/_gmsh41.py
```diff
@@ -304,52 +367,214 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+        dim, tag = node_dim_tags[j]
+
+        # Find the matching cell block, if it exists
+        matching_cell_block = numpy.where(
+            numpy.logical_and(
+                cell_dim_tags[:, 0] == node_dim_tags[j, 0],
```
nschloe, line 468: dim[0]

meshio/gmsh/_gmsh41.py
```diff
@@ -304,52 +367,214 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+                cell_dim_tags[:, 1] == node_dim_tags[j, 1],
```
nschloe, line 469: tag[0]

meshio/gmsh/_gmsh41.py
nschloe: There are more instances of this further down

meshio/gmsh/_gmsh41.py
```diff
@@ -307,52 +371,210 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+
+    fh.write(b"$Entities\n")
+
+    # Array of entity tag (first row) and dimension (second row) per node.
+    # We need to combine the two, since entity tags are reset for each dimension.
+    # Uniquify, so that each row in node_dim_tags represent a unique entity
+    node_dim_tags = numpy.unique(point_data["gmsh:dim_tags"], axis=0)
+
+    # Write number of entities per dimension
+    num_occ = numpy.bincount(node_dim_tags[:, 0], minlength=4)
```
nschloe, line 443: Can we have more than 4 entries here? Better `assert len(num_occ) == 4`.

meshio/gmsh/_gmsh41.py
```diff
@@ -307,52 +371,210 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+        fh.write(
+            f"{num_occ[0]} {num_occ[1]} {num_occ[2]} {num_occ[3]}\n".encode("utf-8")
+        )
+
+    # Array of dimension and entity tag per cell. Will be compared with the
+    # similar not array.
+    cell_dim_tags = numpy.empty((len(cells), 2), dtype=int)
+    for ci in range(len(cells)):
+        cell_dim_tags[ci] = [
+            _geometric_dimension[cells[ci][0]],
```
nschloe, line 456: `_geometric_dimension` is now called `_topological_dimension`. This needs to be adapted.

meshio/gmsh/_gmsh41.py
```diff
@@ -307,52 +371,210 @@ def write4_1(filename, mesh, float_fmt=".16e", binary=True):
...
+    # We will only deal with bounding entities if this information is available
+    has_bounding_elements = "gmsh:bounding_entities" in cell_sets
+
+    # The node entities form a superset of cell entities. Write entity information
+    # based on nodes, supplement with cell information when there is a matcihng
+    # cell block.
+    for dim, tag in node_dim_tags:
+
+        # Find the matching cell block, if it exists
+        matching_cell_block = numpy.where(
```
nschloe, line 469: We can probably live without `where` here. Just store the boolean array and replace `matching_cell_block.size > 0` by `numpy.any(matching_cell_block)`.

keileg: If matching_cell_block is boolean, we still need the index for the match (see calls `cell_data["gmsh:physical"][matching_cell_blocks[0]]` below).

nschloe: Okay, yes.

keileg:
I see there is some issue with the vtu polyhedral_cells, in one of the test functions. I will push the update in this PR.

I have no idea why that issue did not reveal itself long before - it should have been a consistent problem - but anyhow

dalcinl:
I have objections to some side-effects of this PR. The output meshes have `$NodeData` and `$ElementData` sections that should not be there, as it is redundant information.

@keileg Do you really need these sections in the output file? IMHO, all these `gmsh:xxx` point/cell data attributes should be considered extra metadata that could be used by power users to load-edit-save meshes.

@nschloe It really looks like that it is a bad idea to stash `gmsh:xxx` as regular cell/point data. As the code evolve, people keep handling/messing around with them, the rules are not clear, and behavior change over and over again.

It there any objection to the following change?

```diff
@@ -355,28 +356,18 @@
         # temporarily removed from point_data, and restored to Mesh when writing
         # is done.
 
-        if "gmsh:dim_tags" in mesh.point_data:
-            # Temporarily store the data as 1d arrays
-            dim_tags = mesh.point_data.pop("gmsh:dim_tags", None)
-            mesh.point_data["node_entity_dimension"] = dim_tags[:, 0]
-            mesh.point_data["point_entity"] = dim_tags[:, 1]
-
         _write_elements(fh, cells, mesh.cell_data, binary)
         if mesh.gmsh_periodic is not None:
             _write_periodic(fh, mesh.gmsh_periodic, float_fmt, binary)
+
         for name, dat in mesh.point_data.items():
-            _write_data(fh, "NodeData", name, dat, binary)
+            if name not in ["gmsh:dim_tags"]:
+                _write_data(fh, "NodeData", name, dat, binary)
 
         cell_data_raw = raw_from_cell_data(mesh.cell_data)
         for name, dat in cell_data_raw.items():
-            _write_data(fh, "ElementData", name, dat, binary)
-
-        # Restore dim_tag information if available
-        if "point_entity" in mesh.point_data:
-            mesh.point_data["gmsh:dim_tags"] = dim_tags
-            # Also remove the temporary 1d arrays
-            mesh.point_data.pop("node_entity_dimension")
-            mesh.point_data.pop("point_entity")
+            if name not in ["gmsh:physical", "gmsh:geometrical"]:
+                _write_data(fh, "ElementData", name, dat, binary)
```

BTW, the v2.2 writer will need a similar hack to skip saving `$NodeData`.

nschloe:
@dalcinl Good call. The concerns are addressed in https://github.com/nschloe/meshio/pull/961. (Next time you can just PR, it's easier for me to press a button than to copy-paste a patch.)

dalcinl:
@nschloe Can you please decline #961 ? I'm working on a proper PR. When I dump a patch, my intention is to show you in code what I'm talking about to get early feedback.

keileg:
@dalcinl the extra information is needed to build meaningful input for gmsh (that is possibly true only for cases where lower-dimensional entities have been assigned physical tags). And yes, I have a case where such functionality is needed. However, I fully support a cleaner way of handling this information internally to `meshio`; the `gmsh:xx` fields are indeed messy.

dalcinl:
@keileg But if you need that extra information, then it should go to the `$Entities` section, that is the original source of it!

dalcinl:
Or maybe meshio should store an `gmsh_entities` attribute to have the additional metadata. But dumping alien stuff in the files is not the proper solution, IMHO. Right now I'm working with 50GB meshes, within a read-preprocess-write-compute cycle. Dumping even more stuff to disk is not nice.

PR #927, opened 2020-09-26
Read incomplete cell data for MED

### LINKED ISSUE #876: Issue with converting some .med files with results
[Meshes.zip](https://github.com/nschloe/meshio/files/4981269/Meshes.zip)
I have encountered some issues using meshio in converting med files (generated by CodeAster) to xdmf. Attached are 4 .med files:

UnitCube.med - the mesh used
Meshanical.med - results for mechanical simulation
Thermal.med - results for thermal simulation
TransientThermal.med - results for time-dependent thermal simulation

UnitCube.med and Thermal.med both convert as expected. 

TransientThermal.med converts, however the results for each timestep are stored as individual results.  This means that they can't be viewed as an animation in paraview, however this can be corrected using the TimeSeriesWriter attribute.

Mechanical.med I receive the error:
```
AssertionError: Incompatible cell data. 3 cell blocks, but 'reslin__SIEF_ELGA' has 1 blocks
```
I believe this is to do with the fact that the results generated are explicitly for volume elements, so it is only to these that the results are applied, however meshio also looks for results for the remaining elements (faces and edges).

Note that the med files created by CodeAster are often output as .rmed files however the extension for these has been changed to .med to use with meshio

Thanks
Rhydian

tianyikillua:
When cell data is not available for a cell type, `None` is used in the list.

We read cell blocks first, to get an ordered cell types.

Then we fill in the cell data according to the corresponding position.

Should fix #876.


nschloe:
Thanks!

PR #932, opened 2020-09-28
Update _common.py

samkaksam:
fixed raw_from_cell_data to be able to process cell data with only one dimension

### DIFF
--- a/meshio/_common.py
+++ b/meshio/_common.py
@@ -156,7 +156,7 @@ def cell_data_from_raw(cells, cell_data_raw):
 
 
 def raw_from_cell_data(cell_data):
-    return {name: numpy.concatenate(value) for name, value in cell_data.items()}
+    return {name: value if len(value.shape) == 1 else numpy.concatenate(value) for name, value in cell_data.items()}
 
 
 def write_xml(filename, root):


nschloe:
I need to see what the issue is exactly. Best append a mesh that cannot be read/written with meshio master.

samkaksam:
here a minimal example
```
import meshio as mo
import numpy as np

points = np.array(
    [
        [0.0, 0.0, 0.0],
        [1.0, 0.0, 0.0],
        [0.0, 1.0, 0.0],
        [1.0, 0.0, 0.0],
        [2.0, 0.0, 0.0],
        [1.0, 1.0, 0.0],
        [2.0, 0.0, 0.0],
        [3.0, 0.0, 0.0],
        [2.0, 1.0, 0.0],
    ]
)
cells = [
        ("triangle", np.array([[0, 1, 2]])),
        ("triangle", np.array([[3, 4, 5]])),
        ("triangle", np.array([[6, 7, 8]]))
        ]
cell_data = {'Element label' : np.array([1,2,3])}
mo.write_points_cells(
    "foo.vtu",
    points,
    cells,
    cell_data=cell_data,
)
```

nschloe:
`cell_data` is wrong, it should be `np.array([[1], [2], [3]])`.

PR #933, opened 2020-09-29
nastran writer: use fixed length format

vlukes:
Meshio Nastran writer produces files with comma separated items, called "free field format" in [1]. Unfortunately this is not accepted by my Marc Mentat 2010 (yes, a bit outdated version). So I have prepared this PR, which:

 * implements writing data in "fixed" format
 * adds new `free_format` attribute to `writer()` function to switch between "free" and "fixed" formats. The default value is `free_format=True`.
 * fixes the "free" format code - according to [1], page 28,  data cannot contain embedded blanks

Original output:
```
GRID, 1331,, -4.0000000000000002e-01, 4.0000000000000002e-01, -4.0000000000000002e-01
CHEXA, 1,, 1, 3, 360, 359, 32, 113
+, 603, 522
```
New output - free format:
```
GRID,1331,,-4.0000000000000002e-01,4.0000000000000002e-01,-4.0000000000000002e-01
CHEXA,1,,1,3,360,359,32,113
+,603,522
```
Still, I'm not sure if the line continuation, i.e. `+,`,  in the example above (free format) is correct. I can not find any comments regarding to this in [1].

[1] https://docs.plm.automation.siemens.com/data_services/resources/nxnastran/10/help/en_US/tdocExt/pdf/User.pdf

### DIFF
--- a/meshio/nastran/_nastran.py
+++ b/meshio/nastran/_nastran.py
@@ -192,7 +192,68 @@ def read_buffer(f):
     return mesh
 
 
-def write(filename, mesh):
+def print_scientific(value, length=16):
+    """
+    Prints a value in 16-character scientific notation.
+    This is a sub-method and shouldnt typically be called
+    """
+    aux = length - 2
+    # sfmt = "{" + f":{length}s" + "}"
+    sfmt = "{" + ":s" + "}"
+    pv_fmt = "{" + f":{length}.{aux}e" + "}"
+
+    if value == 0.0:
+        return sfmt.format("0.")
+
+    python_value = pv_fmt.format(value)  # -1.e-2
+    svalue, sexponent = python_value.strip().split("e")
+    exponent = int(sexponent)  # removes 0s
+
+    sign = "-" if abs(value) < 1. else "+"
+
+    # the exponent will be added later...
+    sexp2 = str(exponent).strip("-+")
+    value2 = float(svalue)
+
+    # the plus 1 is for the sign
+    len_sexp = len(sexp2) + 1
+    leftover = length - len_sexp
+    leftover = leftover - 3 if value < 0 else leftover - 2
+    fmt = "{" + f":1.{leftover:d}f" + "}"
+
+    svalue3 = fmt.format(value2)
+    svalue4 = svalue3.strip("0")
+    field = "{:s}".format(svalue4 + sign + sexp2)
+    return field
+
+
+def write(filename, mesh, point_format="fixed-large", cell_format="fixed-small"):
+    if point_format == "free":
+        grid_fmt = "GRID,{:d},{:s},{:s},{:s},{:s}\n"
+        float_fmt = print_scientific
+    elif point_format == "fixed-small":
+        grid_fmt = "GRID    {:<8d}{:<8s}{:>8s}{:>8s}{:>8s}\n"
+        float_fmt = lambda x: f"{x:f}".rstrip("0")[:8]
+    elif point_format == "fixed-large":
+        grid_fmt = "GRID*   {:<16d}{:<16s}{:>16s}{:>16s}\n*       {:>16s}\n"
+        float_fmt = print_scientific
+    else:
+        raise RuntimeError(f'unknown "{format}" format')
+
+    if cell_format == "free":
+        int_fmt, cell_info_fmt = "{:d}", "{:s},{:d},{:s},"
+        sjoin = ","
+    elif cell_format == "fixed-small":
+        int_fmt, cell_info_fmt = "{:<8d}", "{:<8s}{:<8d}{:<8s}"
+        sjoin, cchar = "", "+"
+        nipl1, nipl2 = 6, 14
+    elif cell_format == "fixed-large":
+        int_fmt, cell_info_fmt = "{:<16d}", "{:<8s}{:<16d}{:<16s}"
+        sjoin, cchar = "", "*"
+        nipl1, nipl2 = 2, 6
+    else:
+        raise RuntimeError(f'unknown "{format}" format')
+
     if mesh.points.shape[1] == 2:
         logging.warning(
             "Nastran requires 3D points, but 2D points given. "
@@ -208,45 +269,47 @@ def write(filename, mesh):
 
         # Points
         point_refs = mesh.point_data.get("nastran:ref", None)
-        if point_refs is not None:
-            for point_id, x in enumerate(points):
-                f.write(
-                    "GRID, {:d}, {:d}, {:.16e}, {:.16e}, {:.16e}\n".format(
-                        point_id + 1, point_refs[point_id], x[0], x[1], x[2]
-                    )
-                )
-        else:
-            for point_id, x in enumerate(points):
-                f.write(
-                    "GRID, {:d},, {:.16e}, {:.16e}, {:.16e}\n".format(
-                        point_id + 1, x[0], x[1], x[2]
-                    )
-                )
+        for point_id, x in enumerate(points):
+            fx = [float_fmt(k) for k in x]
+            pref = str(point_refs[point_id]) if point_refs is not None else ""
+            f.write(grid_fmt.format(point_id + 1, pref, fx[0], fx[1], fx[2]))
 
         # CellBlock
         cell_id = 0
         cell_refs = mesh.cell_data.get("nastran:ref", None)
         for ict, (cell_type, cells) in enumerate(mesh.cells):
             nastran_type = meshio_to_nastran_type[cell_type].replace("_", "")
+            if cell_format.endswith("-large"):
+                nastran_type += "*"
             if cell_refs is not None:
                 cell_refs_t = cell_refs[ict]
             else:
                 cell_ref = ""
             for ic, cell in enumerate(cells):
                 if cell_refs is not None:
-                    cell_ref = " " + str(int(cell_refs_t[ic]))
+                    cell_ref = str(int(cell_refs_t[ic]))
                 cell_id += 1
-                cell_info = f"{nastran_type}, {cell_id:d},{cell_ref}, "
+                cell_info = cell_info_fmt.format(nastran_type, cell_id, cell_ref)
                 cell1 = cell + 1
                 cell1 = _convert_to_nastran_ordering(cell1, nastran_type)
-                conn = ", ".join(str(nid) for nid in cell1[:6])
-                f.write(cell_info + conn + "\n")
-                if len(cell1) > 6:
-                    conn = ", ".join(str(nid) for nid in cell1[6:14])
-                    f.write("+, " + conn + "\n")
-                    if len(cell1) > 14:
-                        conn = ", ".join(str(nid) for nid in cell1[14:])
-                        f.write("+, " + conn + "\n")
+                conn = sjoin.join(int_fmt.format(nid) for nid in cell1[:nipl1])
+                if len(cell1) > nipl1:
+                    if cell_format == "free":
+                        cflag1 = cflag3 = ""
+                        cflag2 = cflag4 = "+,"
+                    else:
+                        cflag1 = cflag2 = f"{cchar}1{cell_id:<6x}"
+                        cflag3 = cflag4 = f"{cchar}2{cell_id:<6x}"
+                    f.write(cell_info + conn + cflag1 + "\n")
+                    conn = sjoin.join(int_fmt.format(nid) for nid in cell1[nipl1:nipl2])
+                    if len(cell1) > nipl2:
+                        f.write(cflag2 + conn + cflag3 + "\n")
+                        conn = sjoin.join(int_fmt.format(nid) for nid in cell1[nipl2:])
+                        f.write(cflag4 + conn + "\n")
+                    else:
+                        f.write(cflag2 + conn + "\n")
+                else:
+                    f.write(cell_info + conn + "\n")
 
         f.write("ENDDATA\n")
 
--- a/meshio/nastran/_nastran.py
+++ b/meshio/nastran/_nastran.py
@@ -227,6 +227,10 @@ def print_scientific(value, length=16):
     return field
 
 
+def _float_rstrip(x, n=8):
+    return f"{x:f}".rstrip("0")[:n]
+
+
 # There are two basic categories of input data formats in NX Nastran:
 #
 #     "Free" format data, in which the data fields are simply separated by commas. This type of data is known as free field data.
@@ -243,7 +247,7 @@ def write(filename, mesh, point_format="fixed-large", cell_format="fixed-small")
         float_fmt = print_scientific
     elif point_format == "fixed-small":
         grid_fmt = "GRID    {:<8d}{:<8s}{:>8s}{:>8s}{:>8s}\n"
-        float_fmt = lambda x: f"{x:f}".rstrip("0")[:8]
+        float_fmt = _float_rstrip
     elif point_format == "fixed-large":
         grid_fmt = "GRID*   {:<16d}{:<16s}{:>16s}{:>16s}\n*       {:>16s}\n"
         float_fmt = print_scientific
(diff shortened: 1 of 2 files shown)


nschloe:
Interesting. The documentation you link says
> There are two basic categories of input data formats in NX Nastran:
> * “Free” format data, in which the data fields are simply separated by commas. This type of data is known as free field data.
> * “Fixed” format data, in which your data must be aligned in columns of specific width. There are two subcategories of fixed format data that differ based on the size of the fixed column width: 
>   - Small field format, in which a single line of data is divided into 10 fields that can contain eight characters each.
>   - Large field format, in which a single line of input is expanded into two lines The first and last fields on each line are eight columns wide, while the intermediate fields are sixteen columns wide. The large field format is useful when you need greater numerical accuracy.

I would like to see a link to the document and this text as a comment in the code. I would also like the argument to be `format="free"`, with a choice between `"free", "small", "large"`. Not all three have to be implemented now.

vlukes:
The format switching argument is now `format`, which can be `"free"` or `"fixed"` ("small" or "large" are usually mixed).
The suggested comments and the link to the specification are added.

Another thing to discuss: My experience is that "fixed" format for bulk data is more common (GMSH, MSC.Mentat, ...) than the "free" format and I would like the default format to be "fixed".  What is your opinion?
"Fixed" format does not preserve the accuracy required in the tests, because mesh coordinates are stored as `%.8e` floats,  mantissa + exponent must fit into 16 characters.

meshio/nastran/_nastran.py
```diff
@@ -192,7 +192,24 @@ def add_cell(cell, cell_type, cell_ref, keyword_prev):
...
+#         Small field format, in which a single line of data is divided into 10 fields that can contain eight characters each.
+#         Large field format, in which a single line of input is expanded into two lines The first and last fields on each line are eight columns wide, while the intermediate fields are sixteen columns wide. The large field format is useful when you need greater numerical accuracy.
+#
+# See: http://help.autodesk.com/view/NSTRN/2019/ENU/?guid=GUID-42B54ACB-FBE3-47CA-B8FE-475E7AD91A00:
+
+
+def write(filename, mesh, format="free"):
+    if format == "free":
+        is_free_format = True
+    elif format == "fixed":
```
nschloe, line 208: I would like to see `fixed-small` and `fixed-large` here.

nschloe:
> "Fixed" format does not preserve the accuracy required in the tests, because mesh coordinates are stored as %.8e floats, mantissa + exponent must fit into 16 characters.

I'm against defaulting to a format variant that does not keep the data. If you want fixed to be the default, I could perhaps settle with `fixed-large`.

meshio/nastran/_nastran.py
vlukes: "Large" and "small" are not different formats but two possible ways of storing data in "fixed" field format. Tetrahedrons are stored using "small" record while hexahedral cells require "large" variant, because (key + cell_id + mat_id + 8nodes) > 10 fields.

vlukes: Mesh coordinates can be exported in the "short" variant, but the accuracy is limited to 8 characters. So it is better to use the "large" format where 16 characters per number are available (implemented in this PR).

nschloe: I don't understand the format. The spec says that, for using large, you'll have to use two lines. Correct? And the first and last entry in each line can be eight characters. Why?? It seems to defeat the purpose of greater accuracy if only _some_ data is more accurate. What is `mat_id`?

vlukes: By `mat_id`I mean a material id or a cell group.

Yes, the "fixed" format can be a bit confusing. Each line of a file has 80 characters divided into ten blocks of 8 characters.

- short format: item (cell, coordinates) per line, max 10 fields, 8chars per field 

  example: quad cell - key, cell_id, cell group, ND1, ND2, ND3, ND4
  ```
  CQUAD4  18      1       20      21      30      29
  ```

- long format: item = at least two lines, the last field (8 chars) in line is a "continuation tag", which is equal to the first field in the following line

  example: hexahedron
  line1: key, cell id, cell group (empty), ND1, ND2, ND3, ND4, ND4, ND5, ND6, unique continuation tag
  line2: continuation tag, ND7, ND8
  ```
  CHEXA   1000            1331    449     241     341     159     123     +13e8   
  +13e8   122     141
  ```

Coordinates - long format means 2 * 8chars per coordinate, two lines:
line1: key (`GRID*` = long grid), id, node group (empty), X, Y
line2: `*` = continuation, Z
```
GRID*   45                                  6.000000E-02    3.000000E-02
*           0.000000E+000
```               

Short format:
key, id, node group (empty), X=0.003456, Y=0.001234, Z=0.005678
```
GRID    45      0.0034560.0012340.005678
```

nschloe: Things I don't understand:
 * Where are the 16 digits in the long format for cells?
 * How is the last entry in long nodes `0.000000E+000`. According to the spec, we only have 8 digits.

nschloe: Ah, I get the second question: Each line is divided into a grid if 8+16+16+16+8 columns. The first and the last are not the acutal entries, but the columns.

vlukes: - As far as I know, there are no 16 digits for cells.
- Right, 8+16+16+16+16+8 is the line division for long nodes. The keywords and continuation tags are always limited to 8 characters.

nschloe: > As far as I know, there are no 16 digits for cells. 

But the documentation says so, doesn't it? If I understand correctly, this
```
CHEXA   1000            1331    449     241     341     159     123     +13e8   
+13e8   122     141
```
shouldn't exist. (Sorry if I'm getting into much detail here, I just want to make sure we don't implement our own standard.)

vlukes: > But the documentation says so, doesn't it? 

There is no specification for 16 digit cell entries in [1] (I did find any, but, of cource, I have not read all the pages).

> If I understand correctly, this

> ```
> CHEXA   1000            1331    449     241     341     159     123     +13e8   
> +13e8   122     141
> ```
> 
> shouldn't exist.

This is according to the documentation, see specifications around page 30 and examples at pages 32, 167, 747, 865.
The continuation tag can be an arbitrary alphanumerical string, see page 31. I use a combination of extension line number and hexadecimal value of cell id (1. extra line + 3e8 --> +13e8).

> (Sorry if I'm getting into much detail here, I just want to make sure we don't implement our own standard.)

That is fine, I understand to your approach.

SteveDoyle2:
I'm the author of pyNastran.  I don't think this addresses the issue with:

`GRID,1331,,-4.0000000000000002e-01,4.0000000000000002e-01,-4.0000000000000002e-01
`
A row can only use 72 characters.  The data beyond that is a comment.  For large filed format, you have to write it like:

```
GRID*              1331                0-4.0000000000e-1-4.0000000000e-1
*      -4.0000000000e-1
```
So, the first field (the name) has 8 characters, then it's 16 characters for each field until the end of the row.  "Row 2" must start with a * and is 8 characters.  Then it has field width of 16 characters.

vlukes:
Thank you @SteveDoyle2 for your comments. I don't use free format, so I left it unchanged in this PR, except for removing blanks. For free format, I tried to reduce the grid rows to 72 characters, but my MSC.Mentat is still not able to read it correctly.
@SteveDoyle2, do you have a proper free format example file?

SteveDoyle2:
That's odd.  I don't know the specifics for MSC.Mentat (that's SOL 600?).

I don't use free format that often (mostly because it throws lots of warnings).  I much prefer large field.  I don't have a robust writer for free field.  I can make one work with assumptions, but large field is more robust.

I assume you're not writing coordinate systems so you can use:

```
# cp='' also works
cp = 0
msg = ('GRID*   %16i%16s%16s%16s\n'
      '*       %16s\n' % (
         self.nid,
         cp,
         print_scientific_16(xyz[0]),
         print_scientific_16(xyz[1]),
         print_scientific_16(xyz[2])))

```
```
def print_scientific_16(value: float) -> str:
    """
    Prints a value in 16-character scientific notation.
    This is a sub-method and shouldnt typically be called

    """
    if value == 0.0:
        return '%16s' % '0.'

    python_value = '%16.14e' % value  # -1.e-2
    svalue, sexponent = python_value.strip().split('e')
    exponent = int(sexponent)  # removes 0s

    if abs(value) < 1.:
        sign = '-'
    else:
        sign = '+'

    # the exponent will be added later...
    sexp2 = str(exponent).strip('-+')
    value2 = float(svalue)

    # the plus 1 is for the sign
    len_sexp = len(sexp2) + 1
    leftover = 16 - len_sexp

    if value < 0:
        fmt = "%%1.%sf" % (leftover - 3)
    else:
        fmt = "%%1.%sf" % (leftover - 2)

    svalue3 = fmt % value2
    svalue4 = svalue3.strip('0')
    field = "%16s" % (svalue4 + sign + sexp2)
    return field

```

nschloe:
@SteveDoyle2 Would you say that the fixed format is more prevalent than the free format?

SteveDoyle2:
@nschloe Fixed format is definitely more common than free format.  Any model that's created by Patran, FEMAP, or Simcenter 3d (the common GUIs) is in fixed format.  

Small field format is probably even more common, but there are precision issues with it.  It's slower to write though.

The only caveat with large field is that you can't "break" a line, so if you only have 4 fields (the first line), you'd stick a * on the next line:

```
SPOINT* 42
*
```

Same goes for say 13 fields.  The GRID card doesn't have that issue though.

nschloe:
And does the entire file have to be in the same format or can the format change between nodes and cells, or even from one node/cell to another?

SteveDoyle2:
You can mix small, large, and free field format.  In general, if you're going to use small field format (common for elements), you'd do something like:

```
CHEXA       1703      28      16     259    1537     258     540    2198
            4685    2195
```
It just has 8-character fields and the first field has to be left justified.  All cards (large/small) are 72-characters wide.  The next field is a continuation flag (8-characters).  Anything beyond 80 characters is a comment.  However, if you're going to write a comment, it's preferred to just write it before the card.

```
$ a comment
CHEXA       1703      28      16     259    1537     258     540    2198
            4685    2195
```

You can even mix small and large field format on the same card, but don't do that :)

nschloe:
Okay, @vlukes, then let's add _two_ flags to `write()`, `point_format` and `cell_format` and let's default the first to `"fixed-large"` and the second to `"fixed-small"`.

vlukes:
There is the question of what float format to use for grids.

- `"fixed-large"`: this is clear, value 1.23456789e-3 is represented in scientific notation (employing the code of @SteveDoyle2 SteveDoyle2) as "-1.23456789-3"
- `"fixed-small"`: limitation to 8 characters, scientific notation "-1.235-3" or "-0.00123"?
- `"free"` format: line limited to 72 chars --> (72 - len('GRID') - 5*len(',') - (9 + 3)) / 3 = 17 characters per float when taking 9 characters for point id and 3 for point group id. I suggest using the same scientific notation as for `"fixed-large"`.

nschloe:
What do you mean by "grid"?

vlukes:
By "grid" I mean an entry for coordinates.

nschloe:
Okay yes, that's what I mean by "points". Let's do `fixed-large` for them.

nschloe:
@vlukes Ping me once this is ready.

SteveDoyle2:
@vlukes 
Here's a attached mixed small/free-formatted dat/bdf file.  There's plenty of examples that come with MSC/NX Nastran.  Marc may have a different reader and I don't know the specifics of that.  I know for a fact that MSC Patran and MSC Nastran use different readers.

[aabslg1.txt](https://github.com/nschloe/meshio/files/5342362/aabslg1.txt)

As I mentioned though, you may need to add property IDs for Marc to be able to read it.  Have you verified Marc works on small/large field formatted decks?

vlukes:
@SteveDoyle2 Thank you for the file. I was able to import it in MSC.Mentat. Mentat works with all the variants (free/small/large).

meshio/nastran/_nastran.py
```diff
@@ -192,7 +192,82 @@ def add_cell(cell, cell_type, cell_ref, keyword_prev):
     return mesh
 
 
-def write(filename, mesh):
+def print_scientific(value, length=16):
```
nschloe, line 195: This method doesn't print, it returns. Any why "shouldnt it be called"? Also, why not using `{:.16e}`?

vlukes: This is a method proposed by @SteveDoyle2, see comments above. The method can make full use of all 16 characters. It checks the sign, leaves out "e" for exponent, ...

vlukes: Yes, the method name can be changed, eg. to `_scientific_format()`. Or `_float_nastran()`, because it is a "inverse" method to `_nastran_float()` employed by the reader.

nschloe: `float_to_nastran_string` sounds good. Please rename the inverse, too. The doc string needs a couple of illustrating examples which floats are converted to what.

vlukes: Okay, renamed. Examples of float to string conversion are in the docstring of  `_float_to_nastran_string()`.

nschloe:
I'm still pretty skeptical about the "NASTRAN scientific notation". Is this documented somewhere? It just seems to weird to invent another float format. If it's only about the extra `e`, I'd be fine with
```
f"{x:.15e}".replace("e", "")
```
though. That produces trailing zeros, e.g.,
```
1.234567890000000+03
```
for `1234.56789`, but if the notation is fixed, it doesn't matter anyway. It probably even looks better if everthing is en bloc. What do you think?

SteveDoyle2:
@nschloe Dropping the E is definitely valid

Per the Simcenter/NX 2019.2 Quick Reference Guide
image

keurfonluu:
Dropping the E is valid and common in Fortran when dealing with fixed-length floating point numbers. However, it only happens when the exponent field is longer than how it is defined in the exponent format by one char (otherwise, Fortran outputs `****`).

I agree with @nschloe and appending trailing zeros is probably cleaner (and faster) although the output file might be a bit heavier.

keurfonluu:
Another thing to remember if you are on a character budget: negative numbers take one extra character for the sign.

```python
x = 123.0
assert len(f"{x:.15e}") == len(f"{-x:.15e}")  # False, 20 != 21
```

vlukes:
The drawback of trailing zeros in `fixed-large` format, when filling in all 16 characters, is the following:
```
GRID*   1                               5.000000000000-1-5.00000000000-1
*       5.000000000000-1
GRID*   2                               5.000000000000-1-5.00000000000-1
*       -5.00000000000-1
```

It's not pretty at all and it's not "human" readable. I would prefer the current solution which gives:
```
GRID*   1                                           5.-1           -5.-1
*                   5.-1
GRID*   2                                           5.-1           -5.-1
*                  -5.-1
```

When we use `f"{x:.10e}"`, we unnecessarily reduce the number of usable decimal places. For `x = 123456789.123456789`, 
`f"{x:.10e}"` gives `1.2345678912e+08`, while `_float_to_nastran_string(x)` returns `1.234567891235+8`.

SteveDoyle2:
There’s also a print_float_16() In the file that I attached.  It’s
generally easier to read and can provide higher precision in most cases,
but again it’s slower.

nschloe:
I'm worried about code maintenance. I understand that you find that the expanded format doesn't look as nice, but that's not a priority. In fact, I think it's a misunderstanding that humans should be able to "read" mesh file formats, for much the same reason that you'd never try open a JPG with a text editor and try to see something.

SteveDoyle2:
I would worry too much about that.  Just put a note about checking
pyNastran if you find a bug.  The functions have been extensively tested.

nschloe:
Alright, fair enough. Let's get this merged. Thanks @vlukes for the work and everyone for the comments!

PR #961, opened 2020-11-05
gmsh41: don't explicitly store gmsh:dim_tags, physical and geometrical

nschloe:


dalcinl:
Please decline, I'm working on a more complete patch that also handle issues for Gmsh v2.2 format.

PR #962, opened 2020-11-05
Gmsh: Fix writers dumping redundant node and cell data ("gmsh:<name>")

dalcinl:
Update for #922. Also handle Gmsh v2.2 format.


dalcinl:
@keileg Please take a look to this PR. Are any of my changes breaking your use cases?

keileg:
@dalcinl  I think this will be fine for my case; if not, it will be on me to suggest further adjustments based on this.

And I like this approach much better than what I suggested.

Issue #970, opened 2020-11-14
test_xdmf3 failing on 32-bit arches: KeyError: 'uint32'


drew-parsons:
6 meshio tests are failing on 32-bit arches (i386 and armhf).
The failing tests are test_xdmf3  [XML-mesh] with meshes 4, 6, 8, 9, 10, 11.

The common error is 
```
>           fmt = dtype_to_format_string[data.dtype.name]
E           KeyError: 'uint32'
```

Full error is
```
____________________________ test_xdmf3[XML-mesh4] _____________________________

mesh = <meshio mesh object>
  Number of points: 9
  Number of cells:
    triangle6: 2
data_format = 'XML'

    @pytest.mark.parametrize("mesh", test_set_full)
    @pytest.mark.parametrize("data_format", ["XML", "Binary", "HDF"])
    def test_xdmf3(mesh, data_format):
        def write(*args, **kwargs):
            return meshio.xdmf.write(*args, data_format=data_format, **kwargs)
    
>       helpers.write_read(write, meshio.xdmf.read, mesh, 1.0e-14)

test/test_xdmf.py:42: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/helpers.py:347: in write_read
    writer(p, input_mesh)
test/test_xdmf.py:40: in write
    return meshio.xdmf.write(*args, data_format=data_format, **kwargs)
/usr/lib/python3/dist-packages/meshio/xdmf/main.py:530: in write
    XdmfWriter(*args, **kwargs)
/usr/lib/python3/dist-packages/meshio/xdmf/main.py:351: in __init__
    self.cells(mesh.cells, grid)
/usr/lib/python3/dist-packages/meshio/xdmf/main.py:431: in cells
    data_item.text = self.numpy_to_xml_string(cells[0].data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <meshio.xdmf.main.XdmfWriter object at 0xf4300a60>
data = array([[0, 1, 2, 3, 4, 5],
       [1, 6, 2, 8, 7, 4]], dtype=uint32)

    def numpy_to_xml_string(self, data):
        if self.data_format == "XML":
            s = BytesIO()
[...]

PR #971, opened 2020-11-23
fix uint32 name in xdmf dtype_to_format_string

drew-parsons:
Closes Issue #970.


nschloe:
Good catch, thank you!

PR #1026, opened 2021-02-19
Fix non integer error in FLAC3D binary writer

keurfonluu:
- Fixed: `struct.pack` requires pure Python integer.


nschloe:
What type is it if not casting?

keurfonluu:
> What type is it if not casting?

`numpy.uint32`.

PR #1044, opened 2021-03-08
Fix Tecplot reader

### LINKED ISSUE #1043: Tecplot reader: allow multiline VARIABLES list - incomplete
Hi again guys, 

Re: this issue: #1016 which you kindly fixed: #1018, it turns out the fix is incomplete (my bad).  

Evidently, field separators on a line (in a Tecplot record) can be white space or commas (or both), or in fact a new line.  Consequently, what is still failing is a multiline variables case where there is one variable on each line, with no trailing comma.  The current code assumed the last variable on each line would have a trailing comma.  

I believe the fix is within (_tecplot.py) read_buffer:
`if line.upper().startswith("VARIABLES"):
    ...`
`line = "".join(lines)`
which should be:
`line = " ".join(lines)`

I've attached a _bug2 .dat file and _workaround for  demo/test. 
Thanks!

[multilinevars_bug2.zip](https://github.com/nschloe/meshio/files/6098999/multilinevars_bug2.zip)

keurfonluu:
- Fixes #1043.

### DIFF
--- a/meshio/tecplot/_tecplot.py
+++ b/meshio/tecplot/_tecplot.py
@@ -131,7 +131,7 @@ def read_buffer(f):
                 else:
                     f.seek(i)
                     break
-            line = "".join(lines)
+            line = " ".join(lines)
 
             variables = _read_variables(line)
 


nschloe:
Thanks!

PR #1053, opened 2021-03-11
Update README.md

frankeye:
Add file_format strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,12 @@ meshio.Mesh(
 meshio.write_points_cells("foo.vtk", points, cells)
 ```
 For both input and output, you can optionally specify the exact `file_format`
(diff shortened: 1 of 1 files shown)


nschloe:
I don't think this is necessary. It will also add to the clutter in README and will quickly get out-of-sync with what file types are available. `meshio-convert -h` gives a complete list.

frankeye:
Hi Nico,
Good point about clutter, but how about this idea instead:  in the readme where it says:

"For both input and output, you can optionally specify the exact file_format (in case you would like to enforce ASCII over binary VTK, for example)."  

simply append: "meshio-convert -h gives a complete list of file_format strings".

nschloe:
Added that in #1111.

PR #1055, opened 2021-03-13
fix for #1049

### LINKED ISSUE #1049: med files written in v 4.3.11
I was trying to create a med file modifying an existing one and I notice that when importing the created med in ParaVis in Salome, I cannot see its content anymore (see figures below). The content is there as I can correctly read it with meshio and export it for example in a txt file but something seems to be broken in 4.3.11. Probably some of the modifications regarding the med names. In fact with v 4.3.10 it works fine. I am investigating it and will make a PR should I find what is wrong, I reckon it should be something trivial.

screenshot importing file created with v 4.3.10
image

screenshot importing file created with v 4.3.11
image

Jesusbill:
small mod to restore order.
X, Y, Z names used per MED specifications (can be verified with `mesh.attrs["NOM"]`)

Fixes #1049.

### DIFF
--- a/meshio/med/_med.py
+++ b/meshio/med/_med.py
@@ -239,7 +239,7 @@ def write(filename, mesh):
     med_mesh.attrs.create("UNI", numpy_void_str)  # spatial unit
     med_mesh.attrs.create("SRT", 1)  # sorting type MED_SORT_ITDT
     # component names:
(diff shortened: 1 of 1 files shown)


Jesusbill:
I need to make a fix for 1D/2D cases, pushing in a while

nschloe:
Alright, thanks!

Issue #1050, opened 2021-03-10
round-tripping a med file results in losing group names (tags)


Jesusbill:
I would like to make sure that I can round-trip (read, modify and write) a med file without losing information.
Doing some tests I see that the names of the point tags and cell tags are not written (checked with older versions as well).

I am using this simple snippet
```
import meshio

file = meshio.read("path/to/med/file", file_format="med")

meshio.Mesh(
    file.points,
    file.cells_dict,
    # Optionally provide extra data on points, cells, etc.
    point_data=file.point_data,
    cell_data=file.cell_data,
    # field_data=file.field_data
).write(
    "path/to/new/med/file",
    file_format="med"
)
```

With the file attached for example I can get from the original file the correct point and cell tags
```
"point_tags": {
        "2": [
            "base_columns"
        ],
        "1": [
            "base_plates"
        ],
        "3": [
            "point_masses"
        ]
    }
"cell_tags": {
        "-1": [
            "base_plates"
        ],
        "-2": [
            "columns"
        ],
        "-3": [
            "plates"
        ],
        "-4": [
            "point_masses"
        ]
    }
```

but with the same med file written with meshio these tag dictionaries are empty, although the tag assignments to the elements are correctly done.

I am investigating, although not of high priority right now, but I would like to have it fixed soon.

PS: Nico can we have a `med` label for the issues?  :)

[reinforcement.zip](https://github.com/nschloe/meshio/files/6116282/reinforcement.zip)

nschloe:
> PS: Nico can we have a med label for the issues? :)

Added.

Jesusbill:
Hi Nico I have provided a commit for this issue [here](https://github.com/Jesusbill/meshio/commit/5df8193f1cb191e43700d56eb399df0dc9561651); I was thinking if you wanted to review before the PR.

I have the impression that the dictionaries that hold the tag values and the corresponding names were initially named `point_tags` and `cell_tags`, as they are also called the tag values inside the `point_data` and `cell_data`, but then were renamed to `point_sets` and `cell_sets`. But then they were never updated in the `med` implementation and instead two [try/except](https://github.com/nschloe/meshio/blob/6896de02e850340e7496a40c3306a96076be545f/meshio/med/_med.py#L301-L314) statements were used to overcome the AttributeError. Does this train of thought makes sense?

I have tested it now and everything looks to work fine.

nschloe:
[...]

PR #1074, opened 2021-04-17
rename tags to sets, fix for #1050

Jesusbill:


Jesusbill:
Not sure what is failing here, I see L366 but it is not complete
> FAILED test/test_med.py::test_reference_file_with_mixed_cells - TypeError: uf...

By the way when I run `pytest` locally for `med` specifically I get these two errors
```
FAILED test/test_med.py::test_reference_file_with_mixed_cells - OSError: Unable to open file (file signature not found)
FAILED test/test_med.py::test_reference_file_with_point_cell_data - OSError: Unable to open file (file signature not found)
```
trying to open these med files `cylinedr.med` and `box.med` even in Salome I get an error:
`The file is incorrect, some data is missed`

nschloe:
Probably something wrong with your git-lfs.

Jesusbill:
I installed git-lfs but nothing changed really, and these files are quite small. Do they work in your system?

From the error I get and looking around it looks like these files could be `hdf4` and not `hdf5`. And in the [README](https://github.com/nschloe/meshio/blob/main/test/meshes/med/README.md) file it says that they were converted from med 4.0.0 to med 3.0.0. From what I know, if I am not mistaken, Salome supports backwards version up to 3.3.0 or something, otherwise one should probably install quite old versions to open this file.

In any case, looks like I have hit a wall here, do you have any suggestion how we can effectively proceed with this PR? thanks

nschloe:
For your reference:
```
md5sum test/meshes/med/cylinder.med
md5sum test/meshes/med/box.med
```
```
e36b365542c72ef470b83fc21f4dad58  test/meshes/med/cylinder.med
0867fb11bd14b83ad11ab20e2b1fd57d  test/meshes/med/box.med
```

> In any case, looks like I have hit a wall here, do you have any suggestion how we can effectively proceed with this PR? thanks

From what I can see, the error in the tests has nothing to do with files being corrupt or so. That's just a local config issue for you. Of course it's easier to proceed if you can run it all locally, but you could just throw thing at the gh-actions runner by pushing "guess" commits with prints etc. until it works out.

Jesusbill:
Ok thanks, I will work on it one way or another

Jesusbill:
I had to do a workaround locally, since I cannot run the tests :(, to get a working med file and I saw the problem was in the `helper.py` file that checks the read.write operation.

A check that was before [bypassed](https://github.com/nschloe/meshio/blob/main/test/helpers.py#L441) is now considered and the existing [implementation](https://github.com/nschloe/meshio/blob/main/test/helpers.py#L445) assumes the quantities were numbers but for `med` they are strings.
I though I could fix this line but then I got an error for `test_gmsh` so I ended up with a try/except statement to check if we can get a correct build and there it is !

But maybe there can be some improvements, if I can know that the function is called from test_med we could have an if statement there instead of the try/except.

nschloe:
There are still some `print()` in the diff.

I don't quite unterstand why the values of the cell sets would be _strings_. They should be integer arrays. What am I missing?

Jesusbill:
Mmmm I think that there is something wrong in my "train of thoughts" that I described initially, maybe cell_sets are different than cell_tags after all. I have always seen cell_tags for example when "exploring" the meshio schema after reading a med file like this:
"cell_tags": {
        -1: ["base_plates"],
        -2: [ "columns"],
        -3: ["plates"],
        -4: ["point_masses"]
    }
and accordingly I defined `cell_sets` with this format, so the keys are integers and the values are lists of strings.

Jesusbill:
I think we should hold this PR off; @Krande is also working on this and in fact handles tags and sets differently.

Krande:
Hi,

as @Jesusbill mentioned I think I have a working prototype of roundtripping MED files with intact set information using point_set and cell_set as a container for set information (opposed to point and cell tags). It also enables us to convert set information to/from other software with the MED format (which is always nice!). 

Currently I am working on my branch here https://github.com/Krande/meshio/tree/med-sets. I will try find time later today to start a new pull request. 

All tests were OK on my computer (windows).

nschloe:
> and accordingly I defined cell_sets with this format, so the keys are integers and the values are lists of strings.

No, cell_sets are named sets (aka lists) of cells, so `dict(str, List[int])`. That's similar to integer cell_data. The latter _must_ assign a value to each cell, so in some scenarios, `cell_sets` is actually what you want. Used to define boundary regions etc.

Jesusbill:
Thanks @nschloe for the explanation, sorry for any inconvenience and I am faithful @Krande will get it right!
Closing this PR.

PR #1075, opened 2021-04-18
Suggested prototype for MED support of reading/writing point and cell sets (opposed to tags)

Krande:
Hello, in relation to what was discussed in https://github.com/nschloe/meshio/pull/1074, here is my suggestion for how MED support for point and cell sets can be included. 

I still haven't had time to fix the actual test cases for the MED files yet though. The tests are asserting the point and cell tags, not sets so tests will fail. I hope it is OK that I created the PR even though it lacks a few final steps.

Visually the box.med and cylinder.med (and one more test MED file I created) seems to transfer all sets OK (although the cell element numbering seem to have changed during the conversion?).

I will try to find some time later this week to finalize the tests and crack the cell element numbering issue . 

Maybe if @Jesusbill have some time you could take a look and see if you can figure out a way to change the previous test format from tags to sets? And I could also need some help taking a look at the cell element numbering issue too! 

The visual comparison of `cylinder.med`

image

Another of visual comparison of `cantilever.med`.

image

Comparison of re-numbering of cell elements:

image

Oh, and I hope its OK that I added the workspace folder for pycharm (.idea) in the gitignore and added and :rtype: statement on the _helper.read() function. It just made it a bit simpler to navigate using my IDE. 

Anyways, let me know what you think and if you have any suggestions and/or comments.

Best Regards
Kristoffer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ dist/
 doc/_build/
 *.egg-info/
 .pytest_cache/
+.idea/
--- a/meshio/_helpers.py
+++ b/meshio/_helpers.py
@@ -42,6 +42,7 @@ def read(filename, file_format: Optional[str] = None):
     :type filenames: str
 
     :returns mesh{2,3}d: The mesh data.
+    :rtype: meshio._mesh.Mesh
     """
     if is_buffer(filename, "r"):
         if file_format is None:
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -43,6 +43,8 @@ abaqus_to_meshio_type = {
     "S8R": "quad8",
     "S8R5": "quad8",
     "S9R5": "quad9",
+    # 4 node rigid shell element
+    "R3D4": "quad",
     # "QUAD": "quad",
     # "QUAD4": "quad",
     # "QUAD5": "quad5",
--- a/meshio/abaqus/_abaqus.py
+++ b/meshio/abaqus/_abaqus.py
@@ -85,7 +85,8 @@ abaqus_to_meshio_type = {
     "C3D10MH": "tetra10",
     # "TETRA14": "tetra14",
     #
-    # "PYRAMID": "pyramid",
+    "C3D5": "pyramid",
+    "C3D5H": "pyramid",
     "C3D6": "wedge",
     #
     # 4-node bilinear displacement and pore pressure
--- a/meshio/med/_med.py
+++ b/meshio/med/_med.py
@@ -144,12 +144,12 @@ def _cell_tag_to_set(cell_data_array, cell_tags):
             for v in tag_names:
                 res = np.where(cell_data_array == tag_id)[0]
                 if len(res) > 0:
-                    shared_sets.append((v, np.add(res, 1)))
+                    shared_sets.append((v, res))
         else:
             tag_name = tag_names[0]
             res = np.where(cell_data_array == tag_id)[0]
             if len(res) > 0:
-                cell_sets[tag_name] = np.add(res, 1)
+                cell_sets[tag_name] = res
 
     for v, s in shared_sets:
         if v in cell_sets.keys():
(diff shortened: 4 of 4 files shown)


Krande:
A small update.

Successful conversion to abaqus from MED with sets intact:

image

The issue still remains the id numbering of cell and point sets is not 1:1.

Jesusbill:
> Maybe if @Jesusbill have some time you could take a look and see if you can figure out a way to change the previous test format from tags to sets? And I could also need some help taking a look at the cell element numbering issue too!

Sounds great @Krande I will work on these issues

nschloe:
I need to understand how cell sets are implemented in the MED format. Can you briefly explain?

Krande:
I can give it a shot.



As far as I know  The **cell_tags** and **point_tags** format mirrored the MED implementation of storing set information. 

More precisely: set information is split into a dict with keys=id integers and values=list of set names related to that integer. That dictionary is then related to a full length array (equal to length of point/cell arrays) with each indice equal to the id integer .

Here is an example of two sets "SetA" and "SetB" given unique integers of 2 and 3. An integer value=2 in `point_data` represents that the current point belongs to "SetA". Value=0 belongs to no set. For points belonging to both sets (see id=4, an additional set is created where both setnames are the value for the given integer.

```python
point_tags = {2: ["SetA"], 3: ["SetB"], 4: ["SetA", "SetB"]}
point_data = [0, 0, 0, 0, 2, 0, 2, 2, 3, 3, 0, 4]
```

Did that make sense?

nschloe:
> Did that make sense?

Sure. So those, in meshio lingo, are _not_ point sets. This is simply a function that associated one value (let's say an int) to every point. (cells likewise of course). I would suggest to store them as such. The mapping from from ints to strings can be stored in field_data.

In the `Mesh` class, there are methods for converting int data to sets and vice versa, so I'd simply use those if necessary.

Krande:
I see your point. 

But I would argue that the way the MED format stores its set information internally should not dictate how you should store the data in meshio (if that makes any sense)? Given that the information in the cell and point data we are converting right now is defined in Code Aster as `groups`. 

For my specific use-case I would say that converting to point and cell sets and having the ability to convert set information to/from other formats is worth the extra conversion step. But I understand that there are probably many stakeholders in the different formats, so it is not a problem if we keep it as is. I will solve my conversion to/from MED some other way :) That would also mean that the solution that exists per today is kept as is (the cell_tags and point_tags where located on cell_data and point_data respectively).

nschloe:
> But I would argue that the way the MED format stores its set information internally should not dictate how you should store the data in meshio

Nah, I think it should. meshio strives to be stupid, i.e., it sees data in a file and gives you exactly that. I mean, who knows what the user is going to do with it, right? So if there is exactly one int datum associated with every point, it goes into `point_data`.

Actual point/cell sets are more useful in some contexts and it's a pity that most mesh formats don't have those. I totally understand that a user would want to convert point/cell_data into point/cell_sets. That's why there is `mesh.int_data_to_sets()`. I just saw that it doesn't convert point data, but I'll fix that right away.

Would it work for you then already?

nschloe:
PS: I'd suggest to put the element additions for abaqus in a separate PR; those can be merged right away.

Krande:
I think that would work!

Regarding Abaqus support: Sure! Will create a separate pr

Jesusbill:
If `point_tags` and `cell_tags` are kept as is in med and given your vision Nico that meshio should be "stupid" and follow the format converting from/to then can we not also introduce these keywords as input for the "write" function?

Essentially that is what I wanted to solve with my PR and looking at the code I mistook "tags" for "sets". Instead I could just add the functionality missing for tags (which is the dicts associating strings to ints)

nschloe:
I just merged #1076 to fix a couple of things about int-data-to-set conversion.

I noticed that you don't need `field_data` either. Just call your point data key `"key1-key2-key3"` etc. (this should go in `_med.py`) and `int_data_to_sets()` will convert it automatically.

PR #1079, opened 2021-04-20
Fixed tecplot file key reading if 1st data is neg number

davidcortesortuno:
Converting a `.tec` file to any other format was failing for me if the first line of numerical data starts with a negative number, because the `isdigit` method was working only assuming the first character is `0-9` and not considering the `'-'` symbol. This caused numeric keys to be added as a zone entry. The fix is quite straightforward.


nschloe:
I can only consider this if the tests are passing. Let me know if you're still working on this.

davidcortesortuno:
Ok, I'll check the test errors and fix any problems

nschloe:
Something seems to have gone wrong with the commits here.

davidcortesortuno:
Yes, sorry, I just tried to merge with the changes of the main repo. Should I create a new pull request? The fix is very short

nschloe:
Yeah, make a new PR.

PR #1088, opened 2021-04-29
Fixed reading of tecplot files if the first data is a negative or decimal number

davidcortesortuno:
Fixed reading of tecplot files if the first data is a negative number. The fix is straightforward, when reading the `ZONE`, the first two chars of the header lines are read checking the cases `-0` , `0.`, `00` , `.0` (or any digit combination) instead of assuming only a positive number char


nschloe:
This PR now conflicts with main and should be redone.

PR #1099, opened 2021-05-08
Nastran long support

Spectre5:
I've added a new test with a valid Nastran grid in long format that currently fails.  The issue is that `chunks2[1:3]]` in the code below will only be length 1 since the second line in the test grid card is not long enough (not more than 16 characters long).  This leads leads to `ValueError: not enough values to unpack (expected 2, got 1)`.

```python
            chunks2 = _chunk_line(next_line)
            next_line = f.readline()
            points.append(
                [
                    _nastran_string_to_float(i + j)
                    for i, j in [chunks[5:7], chunks[7:9], chunks2[1:3]]
                ]
            )
```

The solution I've added here is to simply use 72 for the line length instead of `len(line)`, which will result in the fixed format line "chunks" always having 9 fields.  It fills the empty ones with empty strings, `''`.

### DIFF
--- a/src/meshio/nastran/_nastran.py
+++ b/src/meshio/nastran/_nastran.py
@@ -356,7 +356,7 @@ def _chunk_line(line):
         chunks = line.split(",")
     else:  # fixed format
         CHUNK_SIZE = 8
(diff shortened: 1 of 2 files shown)


nschloe:
Very nice, thank you!

PR #1103, opened 2021-05-17
Adding C3D15 / wedge15 type

reox:
It looks like that C3D15 was missing in Abaqus Format.
Is it that simple to fix or did I miss something? I checked with a few meshes and I can read them just fine.


nschloe:
Thanks!

PR #1109, opened 2021-05-26
Abaqus: Split elements by expected number not by lines

### LINKED ISSUE #1104: Read Abaqus files with extra commas in element definition
I came across some files which are generated by a mesh-generator which puts trailing commas after each line in the element definition.
It looks like this is accepted by Abaqus - at least I can read those files with Abaqus CAE  V6R2021 and also in Altair HyperMesh 2020 - although the documentation does not state this.

Reading such a mesh in meshio fails:
```
$ meshio-info example_broken.inp 
<meshio mesh object>
  Number of points: 63
  Number of cells:
    tetra10: 0
  Cell sets: Evolumes
  File "~/.local/bin/meshio-info", line 10, in <module>
    sys.exit(info())
  File "~/.local/lib/python3.7/site-packages/meshio/_cli/_info.py", line 30, in info
    point_is_used[cells.data] = True
IndexError: arrays used as indices must be of integer (or boolean) type
$ meshio-info --version
meshio 4.3.6 [Python 3.7.4]
Copyright (c) 2015-2020 Nico Schlömer et al.
```

[example_broken.inp.txt](https://github.com/nschloe/meshio/files/6495245/example_broken.inp.txt)

The issue seems to be this function:
https://github.com/nschloe/meshio/blob/d900512973bddd47021ad4f450936d5fb90833cd/src/meshio/abaqus/_abaqus.py#L245-L276

it reads ids as long as there are commas. The fix would be to get the number of expected nodes from https://github.com/nschloe/meshio/blob/main/src/meshio/_common.py#L7 and try to read as many IDs.

My naive approach would be to read all data items and split them up according to the expected length:
https://github.com/nschloe/meshio/compare/main...reox:abaqus

https://github.com/reox/meshio/blob/88c84b9b55c56be6be003d4f1d380cc53d38ca59/src/meshio/abaqus/_abaqus.py#L247-L282

However, this approach is not perfect, as the assumption that correct element definitions end with a line without a comma is not "checked" anymore.
I'm not sure what the best approach would be in that case. However, as there was no check for valid elements before, it does not change too much but adds the ability to read such files.

reox:
Abaqus files can be written with trailing commas.
While this seems to be not stated explicitly in the abaqus documentation,
abaqus will read such files just fine.
In order to read such files, check for the exact length of data items.

See more information at #1104 
please let me know what you think, I opened the PR for easier discussion on the code.

Fixes #1104.


src/meshio/abaqus/_abaqus.py
```diff
@@ -249,24 +250,28 @@ def _read_cells(f, params_map, point_ids):
...
-        if not line.endswith(","):
-            cell_ids[idx[0]] = counter
-            cells.append([point_ids[k] for k in idx[1:]])
-            idx = []
-            counter += 1
+
+    # Check for expected number of data
+    if len(idx) % num_data != 0:
+        raise ReadError("Expected number of data items does not match element type")
+
```
nschloe, line 270: Can't we just `idx = numpy.array(idx)` here and then pick the first column as id and the therest as point_ids?

reox: I changed that now, but I'm not sure if that is better.
The issue is that we have to translate the node ids and I do not know if there is a better way than using the list comprehension

nschloe: What about
```python
cells = point_ids[idx[:, 1:]]
```
?

nschloe: The `cells_ids` comprehensions looks weird, too. They could be handled with `dict(zip())`.

reox:
```
cells = point_ids[idx[:, 1:]]
```
does not work because point_ids is a `dict` and `idx` is `numpy.ndarray`.

nschloe:
Thanks!

PR #1123, opened 2021-06-04
tox environments for formatting and linting

Spectre5:
The main change is adding 2 new tox environments to format and lint the code.  Now you can use `tox -e fmt` to run isort and black on the code or `tox -e lint` to run black and isort in check mode as well as flake8 (same as is done in the ci checks).

A few other changes:

1. Ignore vscode settings directory
2. Change isort to use "black" compatibility mode (requires isort >5), which will ensure any future required setting changes get automatically applied too.  Let me know if you'd rather remove this though.
3. Change isort "skip" setting to "extend_skip" so that default skipped directories, such as .tox, still get skipped

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ doc/_build/
 .coverage
 .tox/
 foo.vtk
+.vscode/
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -4,4 +4,4 @@ include_trailing_comma=True
 force_grid_wrap=0
 use_parentheses=True
 line_length=88
-skip=test/performance.py
+extend_skip=test/performance.py
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,7 +1,3 @@
 [settings]
-multi_line_output=3
-include_trailing_comma=True
-force_grid_wrap=0
-use_parentheses=True
-line_length=88
+profile=black
 extend_skip=test/performance.py
--- a/Makefile
+++ b/Makefile
@@ -33,5 +33,6 @@ format:
 	blacken-docs README.md
 
 lint:
+	isort --check .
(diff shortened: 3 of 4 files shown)


nschloe:
Can you add a comment or two in the top file on how to use these? I mainly stick with make format, but I'll be happy to merge these, too.

Spectre5:
Ah, I missed the Makefile --- not used to using it for Python projects.  The way it is setup, it basically uses a system/global installation of black, isort, and flake8.  Using tox, they will be installations just for this project, which would also allow you to pin their version (or minimum version) in the tox.ini files deps section.  So if we add these two new tox sections, then I'd also suggest that you change the Makefile to instead call `tox -e fmt` and `tox -e lint` for those two sections.  I can also add blacken-docs to the tox lint section too.  Do you want me to do that?

nschloe:
Correctness is a value, but so is readability, and since the former hasn't been a problem I'd say let's leave the Makefile as is.

Spectre5:
I guess I'm thinking of any new third parties that may contribute in the future.  This way black, isort, blacken-docs, and flake8 would be installed automatically for them if they use `make format` or `make lint` (or the tox equivalents).  As it is, they would need to install them separately and have them globally accessible somewhere.  As for myself, I don't have all of those tools installed globally.

nschloe:
Like I said, it's never been a problem in the past 6 years and 52 contributors, so I guess we'll be fine.

Spectre5:
Ok - last question, do you want me to add `isort --check .` to the Makefile lint section?

nschloe:
> Ok - last question, do you want me to add isort --check . to the Makefile lint section?

Now that it works with `black`, let's add it.

PR #1124, opened 2021-06-04
check that there is a triangle cell

Spectre5:
Currently the STL writer fails if there is not a triangle block in the mesh.  The `pts` ends up being empty which results in `np.cross([], [])` and an exception.  This adds a check that there is a triangle block found in the mesh, or else raises a WriteError.  An empty mesh would also raise this error --- is there a desire to check for that condition too (for all applicable readers, not just STL)?  I see the empty_mesh test is commented out in the tests.


nschloe:
Thanks!

PR #1125, opened 2021-06-04
check cell_type in ascii format too

Spectre5:
Currently for the PLY writer, a check is performed when writing a binary file to make sure the cell type is supported.  This adds that same check to the ascii writer too.  As it is, I was able to write a tetrahedral mesh to PLY ascii output, which results in an invalid file.


nschloe:
Good catch.

PR #1138, opened 2021-06-24
Fix ParaView plugin crash

### LINKED ISSUE #1100: Unable to load plug-in for paraview
Hi I am a windows 10 user that is trying to load the plugin using ParaView. 
I am getting the following error, I can only attach a picture as when I try to the error log to a file, my paraview crashes. 
image

stefsmeets:
Hi all, as part of 574b88483d67044dd5afa4a7fade1c3576f4a8cf the path to the vtk module was updated from `meshio.vtk._vtk` to `meshio.vtk._vtk_42`. The ParaView plugin was not updated however, which causes it to crash directly when loaded (#1100). This PR fixes that.

Fixes #1100

### DIFF
--- a/tools/paraview-meshio-plugin.py
+++ b/tools/paraview-meshio-plugin.py
@@ -12,8 +12,8 @@ from vtkmodules.vtkCommonDataModel import vtkUnstructuredGrid
 import meshio
 
 paraview_plugin_version = meshio.__version__
-vtk_to_meshio_type = meshio.vtk._vtk.vtk_to_meshio_type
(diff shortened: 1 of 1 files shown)


nschloe:
Shouldn't this be `meshio._vtk_common.vtk_to_meshio_type`?

stefsmeets:
> Shouldn't this be `meshio._vtk_common.vtk_to_meshio_type`?

Yep, that also works, didn't realize these functions were shared 👍

nschloe:
Thanks for the contribution!

PR #1148, opened 2021-07-18
Skip some unknown medit keywords generated by mmg software

MoiseRousseau:
I ran into trouble when reading a medit mesh output by the remeshing software mmg due to unknown keyword and not supported information in the mesh: "RequiredVertices", "TangentAtVertices", "Tangents", and "Ridges".
I propose just to skip these information and inform the user that these keywords are not supported by meshio.
I attached an example mesh with the problematic keywords:
[plane.o.zip](https://github.com/nschloe/meshio/files/6835501/plane.o.zip)

Thanks for review,
Moise


nschloe:
You'll have to fix the linter errors. Also, use f-strings instead of `format()`.

nschloe:
You can run `just lint` or install pre-commit to see the lint results locally.

nschloe:
Thanks!

PR #1184, opened 2021-09-21
Handle STL empty mesh read and write

mmcardle:
* Handle Empty STL Mesh reading and writing


nschloe:
Closing in favor of #1189.

PR #1203, opened 2021-10-18
Handle CHEXA (20 nodes) and CPENTA (15 nodes) nodes ordering

Clemoreau:
Cf discussion https://github.com/nschloe/meshio/discussions/1202

Nodes ordering for hexahedron20 and wedges15 are different between vtk and nastran. The problem has been fixed in `_convert_to_vtk_ordering` function

### DIFF
--- a/src/meshio/nastran/_nastran.py
+++ b/src/meshio/nastran/_nastran.py
@@ -270,12 +270,7 @@ def write(filename, mesh, point_format="fixed-large", cell_format="fixed-small")
         cell_id = 0
         cell_refs = mesh.cell_data.get("nastran:ref", None)
         for ict, (cell_type, cells) in enumerate(mesh.cells):
-            # Second order elements
-            if cell_type in ["hexahedron20", "wedge15"]:
-                nastran_type = meshio_to_nastran_type[cell_type]
-            # Remove '_' for first order elements
-            else:
-                nastran_type = meshio_to_nastran_type[cell_type].replace("_", "")
+            nastran_type = meshio_to_nastran_type[cell_type]
             if cell_format.endswith("-large"):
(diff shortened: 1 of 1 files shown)


nschloe:
Tests are failing because the inverse orderings haven't been added yet.

nschloe:
> You mean the conversion to nastran ordering ?

Yep.

Clemoreau:
OK, I've removed it because of one of your comment. I just added it again. 
I assume that it doesn't change anything but it is needed for the tests ?

nschloe:
[0,1,3,2] means:

0 -> 0
1 -> 1
2 -> 3
3 -> 2

Think again! What's the reverse ordering here. (It's not [0,1,2,3].)

Clemoreau:
Actually I'm having a look at the test which fails. In `test_nastran`, the mesh from helpers.hex20_mesh (which contains 1 hexa20 element) is not ordered as the hexa20 element that I have in my .fem file.

So the input mesh (from helpers) is ordered like: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19]
And the output mesh (after write) is [0,1,2,3,4,5,6,7,8,9,10,11,16,17,18,19,12,13,14,15]
So the test fails because input and output are different. 

The question is: where do come from the hexa20 mesh in input ?

Clemoreau:
It's simpler like that and tests are ok on my side. Is it ok for you ? 
Is there still anything to change ? 

I saw that the code coverage was not OK but I guess that it is not linked with my modifications

nschloe:
Looks good to me now. Thanks!

Clemoreau:
Thank you for supporting me !

PR #1226, opened 2021-11-29
Add Windows and Mac CI

### LINKED ISSUE #1225: No Windows (or Mac) testing
It concerns me that there is no CI testing for Windows (or Mac, but Windows is the one that really worries me):

https://github.com/nschloe/meshio/blob/9dc6b0b05c9606cad73ef11b8b7785dd9b9ea325/.github/workflows/ci.yml#L23

Meshio is a hard dependency of PyVista, and I think we should set up multiplatform tests for this. I know meshio is pure Python and minimalist and all that jazz, but things can still be different on Windows, especially when we're dealing with text encodings and reading files.

cc @tkoyama010 @adeak @akaszynski

banesullivan:
Resolve #1225

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Run pre-commit
         uses: pre-commit/action@v2.0.3
 
-  build:
+  test-ubuntu:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


banesullivan:
Mac runners have less available time, so you may want to narrow the Python versions this runs on for those

banesullivan:
Looks like GitHub Actions doesn't want to execute these from my fork 😞

nschloe:
Fixed by #1228.

PR #1236, opened 2021-12-14
Depend on minimum numpy version

wiggin15:
meshio uses `numpy.typing` which was introduced in numpy 1.20.0 (https://numpy.org/doc/stable/release/1.20.0-notes.html#numpy-is-now-typed).

Specifying the minimum required version helps pip upgrade numpy if the environment has an older version installed.
It also lets pip catch conflicts when trying to install a big set of requirements with conflicting dependencies (numpy < 1.20.0 and meshio)


nschloe:
Good catch!

PR #1259, opened 2022-01-11
Fix crash when reading gmsh22 files with multiple cells

### LINKED ISSUE #1257: [BUG] Reading a file with multiple cell types and cell data causes a crash
In the latest version of `meshio` I run into an issue when I try to read a file that contains multiple cell types, *and* also contains cell data.

Here is a minimal failing example:

```python
from meshio import Mesh, read, write

mesh = Mesh(
    points=[
        [0., 0., 0.],
        [1., 0., 0.],
        [1., 1., 0.],
        [0., 1., 0.],
    ],
    cells=[
        ('line', [
            [0, 1],
            [0, 2],
            [0, 3],
            [1, 2],
            [2, 3],
        ]),
        ('triangle', [
            [0, 1, 2],
            [0, 2, 3],
        ]),
    ],
    cell_data={
        'a': [[63, 26, 4, 1, 81], [91, 60]],
    },
)

write('out.msh', mesh, file_format='gmsh22', binary=False)
read('out.msh')
# ValueError: Incompatible cell data. Cell block 0 ('line') has length 5, but corresponding cell data item has length 2.
```

The `read` function errors out with: 

```shell
  File "error.py", line 38, in <module>
    read('out.msh')
  File "c:\users\stef\python\meshio\src\meshio\_helpers.py", line 71, in read
    return reader_map[file_format](filename)
  File "c:\users\stef\python\meshio\src\meshio\gmsh\main.py", line 19, in read
    mesh = read_buffer(f)
  File "c:\users\stef\python\meshio\src\meshio\gmsh\main.py", line 48, in read_buffer
    return reader.read_buffer(f, is_ascii, data_size)
  File "c:\users\stef\python\meshio\src\meshio\gmsh\_gmsh22.py", line 89, in read_buffer
    return Mesh(
  File "c:\users\stef\python\meshio\src\meshio\_mesh.py", line 182, in __init__
    raise ValueError(
ValueError: Incompatible cell data. Cell block 0 ('line') has length 5, but corresponding cell data item has length 2.
```

I have traced the issue to https://github.com/nschloe/meshio/blob/a6175e0d9dfb2aa274392d1cd396e991f0487cbc/src/meshio/_common.py#L88-L90
It does not calculate the correct values for `cs` in this situation.

I also noticed this situation is not covered by the test suite.

I will shortly open a PR to propose a fix.

stefsmeets:
The gmsh22 reader reads cells as tuples , instead of `CellBlock`. This causes issues in determining where to split the data in `cell_data_from_raw()`. This applies when the number of cells > 1 and cell data is defined. This situation currently does not have a test in the test suite and therefore goes undetected.

This PR fixes the problem and adds tests for this case.

Closes #1257

### DIFF
--- a/src/meshio/_common.py
+++ b/src/meshio/_common.py
@@ -86,7 +86,7 @@ num_nodes_per_cell = {
 
 
 def cell_data_from_raw(cells, cell_data_raw):
-    cs = np.cumsum([len(c) for c in cells])[:-1]
+    cs = np.cumsum([len(c[1]) for c in cells])[:-1]
     return {name: np.split(d, cs) for name, d in cell_data_raw.items()}
 
 
--- a/src/meshio/_common.py
+++ b/src/meshio/_common.py
@@ -86,7 +86,7 @@ num_nodes_per_cell = {
 
 
 def cell_data_from_raw(cells, cell_data_raw):
-    cs = np.cumsum([len(c[1]) for c in cells])[:-1]
+    cs = np.cumsum([len(c) for c in cells])[:-1]
     return {name: np.split(d, cs) for name, d in cell_data_raw.items()}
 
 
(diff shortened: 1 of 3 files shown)


stefsmeets:
Closing this because it is out of date.

PR #1262, opened 2022-01-17
fix line 340 in su2

sn1979c:
In line 340 of __su.py, mesh.cells is a list containing 2 CellBlock objects and code is trying to unzip the CellBlock which is unzippable. When I'm trying to export su2 mesh files this problem raised a bug. After fixing it to this, the export works.

### DIFF
--- a/src/meshio/su2/_su2.py
+++ b/src/meshio/su2/_su2.py
@@ -337,7 +337,9 @@ def write(filename, mesh):
             f.write(f"MARKER_TAG= {tag}\n".encode())
             f.write(f"MARKER_ELEMS= {count}\n".encode())
 
-            for index, (cell_type, data) in enumerate(mesh.cells):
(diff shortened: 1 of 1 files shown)


chiefenne:
I checked this PR and for me it worked. Still I can't guarantee that it is not a false implementation in my code which made it fail before. In my code [PyAero](https://github.com/chiefenne/PyAero/tree/develop) I have tested this PR and the export works (including tags). I ran SU2 simulations (serial and parallel version) and the mesh which was exported with this PR worked in both simulations. At least I can guarantee that this PR together with my software is working.

peterdsharpe:
Paging @nschloe , can this please be merged? Confirming that this also fixes a bug in my workflow.

azzeddinetiba:
> Paging @nschloe , can this please be merged? Confirming that this also fixes a bug in my workflow.

same

PR #1295, opened 2022-02-28
bugfix: h5 and xdmf files are not under the same directory

YuanWenqing:


TianBM-codes:
time_series.py
line 265 change to:
self.h5_filename = self.filename.with_suffix(".h5")

PR #1312, opened 2022-04-06
Save all vertex values in obj files

wiggin15:
According to the Wikipedia [documentation](https://en.wikipedia.org/wiki/Wavefront_.obj_file#Geometric_vertex) of .obj files, vertex data can contain either 3 values (x, y, z), 4 values (x, y, z, w) or 6 values (x, y, z, r, g, b).
meshio can read obj files with these values, but when writing, meshio only writes the 3 first values for each point, potentially stripping away important information.
This fix will write all values of vertices, using similar code for `mesh.points` as what is used to save the other parts of the file.

### DIFF
--- a/src/meshio/obj/_obj.py
+++ b/src/meshio/obj/_obj.py
@@ -111,8 +111,11 @@ def write(filename, mesh):
                 __version__, datetime.datetime.now().isoformat()
             )
         )
-        for p in mesh.points:
-            f.write(f"v {p[0]} {p[1]} {p[2]}\n")
+
+        dat = mesh.points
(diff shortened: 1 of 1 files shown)


nschloe:
This will be handled more consistently in the upcoming version. (`"obj:w"`/`"obj:rgb"` in `point_data`.)

wiggin15:
Hi @nschloe . Is there any visibility to the code of the upcoming version? I can't see any recent changes in the `main` branch and no other branches that I can see. I didn't understand exactly what the more consistent fix will be.

PR #1224, opened 2021-11-29
Add back setup.py for editable installations

banesullivan:
Resolve #1223

Issue #1223, opened 2021-11-29
Add developer install instructions


banesullivan:
```
pip install -e .
``` 
does not work on Python 3.8:

```
>>> import scooby
>>> scooby.Report(['setuptools', 'distutils'])

--------------------------------------------------------------------------------
  Date: Mon Nov 29 11:40:24 2021 MST

                OS : Darwin
            CPU(s) : 16
           Machine : x86_64
      Architecture : 64bit
       Environment : Python

  Python 3.8.8 | packaged by conda-forge | (default, Feb 20 2021, 16:12:38)
  [Clang 11.0.1 ]

        setuptools : 49.6.0.post20210108
         distutils : 3.8.8
             numpy : 1.20.2
             scipy : 1.6.3
           IPython : 7.22.0
        matplotlib : 3.4.1
            scooby : 0.5.7
--------------------------------------------------------------------------------
```

It appears the setup.py was removed (without warning?) in https://github.com/nschloe/meshio/pull/795

Since the typical way of installing isn't working, there should be an explanation of the usage of `setup.cfg` in the README and how to install this locally

nschloe:
I just use tox for testing and install it with from pypi or `pip install .`.

banesullivan:
But then you cannot edit the source and it be reflected in the environment. For example check this out:

I updated the inint of `meshio` with this diff:

```diff
diff --git a/src/meshio/__init__.py b/src/meshio/__init__.py
index 035ae53..36c81ae 100644
--- a/src/meshio/__init__.py
+++ b/src/meshio/__init__.py
@@ -80,3 +80,5 @@ __all__ = [
     "topological_dimension",
     "__version__",
 ]
+
+print('this has been editted')
```

But then if I import meshio into Python, I cannot see that change:

```bash
$ meshio git:(main) ✗ python
Python 3.8.8 | packaged by conda-forge | (default, Feb 20 2021, 16:12:38)
[Clang 11.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import meshio
>>>
```
In order to see the change I have to install it again:

```bash
$ meshio git:(main) ✗ pip install .
...
$ meshio git:(main) ✗ python
Python 3.8.8 | packaged by conda-forge | (default, Feb 20 2021, 16:12:38)
[Clang 11.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import meshio
this has been editted
>>>
```

[...]

PR #1323, opened 2022-05-09
Synchronize `setup.cfg` and `pyproject.toml`, bump project versions in both, and add 2-line setup.py

goxberry:
# Summary

This pull request does three things:

- Attempts to synchronize `setup.cfg` and `pyproject.toml` settings so that installing from source via `build` doesn't yield an installation with a stale version like `5.0.0`.
- Bumps the project versions listed in `setup.cfg` and `pyproject.toml` from `5.3.4` and `5.0.0`, respectively, to `7.0.1`, reflecting a patch version increase above the most recent tag and release on GitHub.
- Adds a 2-line, *pro forma* `setup.py` script.

# Rationale

The current project metadata in both `setup.cfg` and `pyproject.toml` looks stale, and the data in each is in conflict. Synchronizing them would make dependency data and other metadata clearer to contributors and packagers.

The 2-line `setup.py` script is intended to ease repackaging of the project. I work at an HPC site and have to convert mesh data amongst multiple formats. This package makes my life much easier, and I am grateful for all the work that has gone into it. In order to use it on some on-site projects, I need to repackage the project from source, ideally both as a wheel and as a source distribution, for security auditing and approval. Part of the rationale for source repackaging is to enable portable installation on air-gapped systems. Given the sometimes ancient software toolchains used at HPC sites, it is much easier to repackage software with `setuptools` than it is to use `build`. To wit, I spent several hours today repackaging `flit_core`, `flit`, `pep517`, and `build` to submit for approval and auditing as a fallback plan in case this pull request adding `setup.py` is rejected. While I wholeheartedly agree that getting rid of `setup.py` is a good thing, adding the two-line version is a pragmatic alternative, particularly because it is easier to get approval where I work for code incorporated into an upstream project than it is to carry around a bespoke patch.

# Alternatives

- Regarding `setup.py`, if the goal is to avoid adding it to the project now and in the future, a simple expedient would be to document that policy in `CONTRIBUTING.md`. If that is indeed project policy, I am happy to add such text to that file. I saw the discussion in #1223 and #1224 also, but I suspect other potential contributors may not see this discussion.
- I saw a [`meshio-rewrite` project](https://pypi.org/project/meshio-rewrite/) on PyPI, but the homepage listed at https://github.com/meshpro/meshio is inaccessible, and it's not entirely clear to me what the relationship of that project is to this one. If the version of the rewrite is indeed 7.0.6, then perhaps these contributions need to be added to that repo instead, or maybe the version number used in this pull request should be different from version 7.0.1.
- It seems that `build` and `flit` both use `pyproject.toml` data by default. If the goal is to move away from `setuptools`, it may be worth getting rid of `setup.cfg` altogether in favor of `pyproject.toml`, though it's not clear to me how the ParaView meshio plugin would be distributed using `pyproject.toml`, given its lack of support for `data_files`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "meshio"
-version = "5.3.4"
+version = "7.0.1"
 description = "I/O for many mesh formats"
 readme = "README.md"
 requires-python = ">=3.7"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "meshio"
-version = "5.0.0"
+version = "5.3.4"
 description = "I/O for many mesh formats"
 readme = "README.md"
 requires-python = ">=3.7"
@@ -11,8 +11,6 @@ authors = [
   {name = "Nico Schlömer"}
 ]
 classifiers = [
-  "Development Status :: 4 - Beta",
-  "Programming Language :: Python",
   "Development Status :: 5 - Production/Stable",
   "Intended Audience :: Science/Research",
   "License :: OSI Approved :: MIT License",
@@ -22,12 +20,14 @@ classifiers = [
   "Programming Language :: Python :: 3.7",
   "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
   "Topic :: Scientific/Engineering",
   "Topic :: Utilities"
 ]
 dependencies = [
   "importlib_metadata; python_version<'3.8'",
-  "numpy>=1.20.0"
+  "numpy>=1.20.0",
+  "rich"
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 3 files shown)


MuellerSeb:
What about using [setuptools_scm](https://github.com/pypa/setuptools_scm) to automatically generate version from latest tag?

goxberry:
If maintainers want me to make those changes, I’m happy to do that. I’m also willing to drop `setup.py` entirely from the PR, if that’s the feedback I get. But as far as I could tell, `build` and `flit` read project metadata from `pyproject.toml`, so I’d like to see that file updated, particularly because the `justfile` in the project uses `build` to build the project. The version information is rather unclear to me, because as far as I could tell a few days ago, the most recent version of `meshio` listed on PyPI is 5.3.4, though there is also the `meshio-rewrite` package listed in my original post.

goxberry:
Closing due to inactivity.

PR #1330, opened 2022-06-08
bugfix: incorrect h5 file path

YuanWenqing:
Generated h5 file is under working dir, not the same dir of xdmf file


chu-k:
+1, addresses #1345

PR #1336, opened 2022-06-28
Ensure all Exodus element types have known topo-dims

smason:
I've just been exploring some ExodusII files and I'm getting an exception:

```none
  File "[...]meshio/src/meshio/_mesh.py", line 99, in __init__
    self.dim = topological_dimension[cell_type]
KeyError: 'tetra4'
```
It seems that there are a few element types, e.g. `tetra4`, that are known to the Exodus reader but don't have an entry in `topological_dimension`.

These changes adds a test that they all exist and fills in the missing entries.

AFAICT the order shouldn't matter so I've just shoved them in somewhere that looks nice – hope that's OK!


smason:
looks related to #1296

PR #1337, opened 2022-07-01
[MDPA] write meshio's cells set as mdpa's submodelparts

marandra:
With this PR, meshio can write cells sets (from gmsh "Physical ...", for example "Inlet", "Outlet", "Walls", "Fluid") as MDPA's submodelparts.

The main necessary change here is to discriminate between regular "Elements" and "Conditions" (i.e., elements where boundary conditions are applied, generally belonging to the skin of the model). 

I also commented out the functions to write NodalData and ElementalData, as they generate an invalid mdpa at the moment.

I added a test that compares the mdpa mesh obtained from a gmsh mesh, with a reference mdpa mesh.

### DIFF
--- a/src/meshio/mdpa/_mdpa.py
+++ b/src/meshio/mdpa/_mdpa.py
@@ -371,36 +371,37 @@ def _write_nodes(fh, points, float_fmt, binary=False):
     fh.write(b"End Nodes\n\n")
 
 
+def _compute_blocks_offset(cells, dimension):
+    """Computes the id offset for the different blocks of entities.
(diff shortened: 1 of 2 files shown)


loumalouomega:
Noice

marandra:
I have rewritten the functions for writing MDPA's submodelparts in a way that is more simple and robust than with the approach in this PR. I will close this PR for now and I will submit a new one.

philbucher:
@marandra any updates on this?

loumalouomega:
> @marandra any updates on this?

Maybe we should write from scratch now that we have geometries in addition to elements/conditions. We can do it more consistently.

PR #1350, opened 2022-08-15
:bug: (_nastran): fix converting float values into nastran string

ironheads:
if a float value is negative and preicise, then the constraint
(len(out)<=len) can not be satisfied. Example: -0.09466145283577683.
fix this using dynamic precision setting.
when convert float -0.09466145283577683, the length of variable `out` is 17 > 16.
after the bug-fix, the length will be 16.


src/meshio/nastran/_nastran.py
konstantin-777, line 1: Why not checking in?
This is an obvious problem where fixed length float field holds anything beyond x.[11 digits]E-1.
Exponent longer than 1 digit will fail. Negative will fail. The edge case is -x.[8 digits]E-100.
Breaks msh->bdf conversion.

Issue #1345, opened 2022-07-29
[BUG]


baaiy0610:
When i try to save files as xdmf in other path. It seems like *.h5 file can not be stored together with *.xdmf file

keurfonluu:
I've got that issue as well. In the meantime, here is an ugly workaround:

```python
import pathlib, shutil

filename = pathlib.Path(filename)
if str(filename.parent) != ".":
    shutil.move(f"{filename.stem}.h5", str(filename.parent))
```

where `filename` is the path to the XDMF file.

PR #1358, opened 2022-09-23
correct `h5` path when writing time series

bhaveshshrimali:
Hi Nico, 

I noticed that the path for the `.h5` file isn't correct especially when writing a time series to a file inside a sub-directory.
 
Specifically the line https://github.com/nschloe/meshio/blob/0138cc8692b806b44b32d344f7961e8370121ff7/src/meshio/xdmf/time_series.py#L265

should instead be `self.h5_filename = self.filename.with_suffix(".h5")`. 

This is consistent with what is in `main.py` https://github.com/nschloe/meshio/blob/0138cc8692b806b44b32d344f7961e8370121ff7/src/meshio/xdmf/main.py#L353

### DIFF
--- a/src/meshio/xdmf/time_series.py
+++ b/src/meshio/xdmf/time_series.py
@@ -262,7 +262,7 @@ class TimeSeriesWriter:
         if self.data_format == "HDF":
             import h5py
 
-            self.h5_filename = self.filename.stem + ".h5"
+            self.h5_filename = self.filename.with_suffix(".h5")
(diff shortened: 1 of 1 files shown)


matthiaskoenig:
Could this please be merged? I am having the same issue and multiple issues are related to this:
https://github.com/nschloe/meshio/issues/1345
https://github.com/nschloe/meshio/pull/1330

leolalson:
Can we merge this ?

PR #1363, opened 2022-10-03
Netgen format compatibility

mhochsteger:
Skip "facedescriptor" and "face_transparencies" data blocks, which are present in new Netgen formats.


mhochsteger:
Closed in favor of #1465

PR #1365, opened 2022-10-13
feature: add support for reading calculix result files FRD

Krande:
Hey,

for a long time I've wanted to add support for the Calculix results file format, `FRD`, but I never got into the FRD data structure. Then yesterday I came across https://github.com/rsmith-nl/calculix-frdconvert by @rsmith-nl (so credit should go to where it's due) which provided enough information about the FRD data structure, so it inspired me to give it a go :)

I tried using my code to create a meshio `Mesh` object and export it to `vtu` and visualize the results in Paraview and it looks OK:

STRESS
image

DISP
image

I tried following the code patterns, but let me know if there are any specific changes I should make to the code itself or my tests/implementations. 

Oh, and by the way. I had to bump the `black` version in the `.pre-commit-config.yaml` file as I was getting [this issue](https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click) apparently fixed by black v22.3.0++. Let me know if you prefer me to revert that change.

Best Regards
Kristoffer


akaszynski:
For those looking for a VTK-like reader, we've implemented this in PyVista and it will be out in the next minor release, v0.48.0 (expect release around April 2026). Thanks @3rav and @Krandle for your work on this.

Issue #1429, opened 2023-08-08
[BUG] Error when writing SU2 boundary data


vtpasquale:
**Describe the bug**
Trying to roundtrip the stock .su2 file fails

**To Reproduce**
This is the example mesh file: https://su2code.github.io/docs_files/square.su2

Running this code
```
import meshio
m = meshio.read("square.su2")
m.write("square_rt.su2")`
````

Results in this error
```
    File ~/.local/lib/python3.10/site-packages/meshio/su2/_su2.py:340 in write
    for index, (cell_type, data) in enumerate(mesh.cells):

    TypeError: cannot unpack non-iterable CellBlock object
```

**Diagnose**
I may ask you to cut and paste the output of the following command.
```
pip freeze | grep meshio
```
```
meshio==5.3.4
```

**To Fix**
I replaced line 340 with these three lines:
```
            for index, cell in enumerate(mesh.cells):
                cell_type = cell.type
                data = cell.data
```
I'll submit a pull request from my fork.

PR #1430, opened 2023-08-08
Update _su2.py

vtpasquale:
Fix for issue #1429


mlaero:
I would like to bump this PR. This fix is also required for me to correctly output SU2 meshes.

Issue #1440, opened 2023-10-10
[BUG] meshio CLI fails to read Abaqus .inp file


jeffrey-cochran:
**Describe the bug**
Runing the command `meshio info file.inp --input-format abaqus` fails with the following error:

```
Traceback (most recent call last):
  File "C:\Users\jeffr\anaconda3\envs\meshing\Scripts\meshio-script.py", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File "C:\Users\jeffr\anaconda3\envs\meshing\Lib\site-packages\meshio\_cli\_main.py", line 52, in main
    return args.func(args)
           ^^^^^^^^^^^^^^^
  File "C:\Users\jeffr\anaconda3\envs\meshing\Lib\site-packages\meshio\_cli\_info.py", line 21, in info
    mesh = read(args.infile, file_format=args.input_format)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jeffr\anaconda3\envs\meshing\Lib\site-packages\meshio\_helpers.py", line 71, in read
    return _read_file(Path(filename), file_format)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jeffr\anaconda3\envs\meshing\Lib\site-packages\meshio\_helpers.py", line 103, in _read_file
    return reader_map[file_format](str(path))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jeffr\anaconda3\envs\meshing\Lib\site-packages\meshio\abaqus\_abaqus.py", line 106, in read
    out = read_buffer(f)
          ^^^^^^^^^^^^^^
  File "C:\Users\jeffr\anaconda3\envs\meshing\Lib\site-packages\meshio\abaqus\_abaqus.py", line 154, in read_buffer
    [point_ids[point_id] for point_id in set_ids], dtype="int32"
     ~~~~~~~~~^^^^^^^^^^
KeyError: 18879
```
What I can tell you is that this node definitely exists, and is definitely in the .inp file.

**To Reproduce**
I cannot provide the Abaqus file because of NDAs. I can provide information about it on request.

**Diagnose**
Result of `pip freeze`:

```
cached-property @ file:///home/conda/feedstock_root/build_artifacts/cached_property_1615209429212/work
certifi==2023.7.22
[...]

PR #1441, opened 2023-10-10
Modified abaqus parser to enable inp files with multiple NODE blocks

jeffrey-cochran:
Tox fails for reasons unrelated to anything I've done, so I'll have to let you make sure the tests still pass.


jeffrey-cochran:
Also worth noting (if it's not stated anywhere else), that this abaqus parser should only really be used on so-called "flat" files for now. It's not set up to handle multiple parts and requires that the file use globale node indexing.

jeffrey-cochran:
Side note:

This pull request fixes issue #1440

PR #1445, opened 2023-11-27
Abaqus only allows 16 integer per line in the element definition part.

vincent-skywalker:
Abaqus only allows 16 integer per line in the element definition part, as shown in the document:

### First line

1. Element number.
2. First node number forming the element.
3. Second node number forming the element.
4. Etc., up to 15 node numbers on this line.

The order of nodes for each element type (the element's connectivity) is given in [About the element library](https://help.3ds.com/2020/english/dssimulia_established/SIMACAEELMRefMap/simaelm-c-general.htm?contextscope%20=%20all).

### Continuation lines (only needed if the previous line ends with a comma)

1. Node numbers forming the element.

Repeat this set of data lines as often as necessary, with up to 16 integer values per line.


vincent-skywalker:
Hi, will this PR be approved? @nschloe

PR #1465, opened 2024-03-27
netgen - skip 'facedescriptors' and 'face_transparencies' blocks

mhochsteger:
Netgen has new fields in the mesh data, just skip it when parsing.


mhochsteger:
Oh, I just saw that #1465 and #1363 also have the same fix.
Just choose one, I will close my other pr.

PR #1496, opened 2024-11-22
Add `py.typed` file to mark `meshio` as partially typed

user27182:
Add `py.typed` to let type-checkers (e.g. mypy) know that the package is partially typed. See https://peps.python.org/pep-0561/#packaging-type-information

Without this, type-checkers treat everything as `Any`

With `py.typed`:
``` python
from meshio import Mesh
reveal_type(Mesh)  # note: Revealed type is "meshio._mesh.Mesh"
```

Without `py.typed`:
``` python
from meshio import Mesh
reveal_type(Mesh)  # note: Revealed type is "Any"
```

### DIFF
--- a/src/meshio/py.typed
+++ b/src/meshio/py.typed
@@ -1,0 +1,1 @@
+partial


user27182:
This PR is not longer needed. It's possible to force analysis of packages without `py.typed` files with mypy 1.14
https://mypy.readthedocs.io/en/latest/changelog.html#optionally-analyzing-untyped-modules

Issue #1499, opened 2024-12-20
[BUG] Tests fail with NumPy 2.2.0


penguinpee:
Fedora has recently upgraded NumPy to 2.2.0 in the development branch. Building `meshio` with that version results in a number of tests failing:

```python
=========================== short test summary info ============================
FAILED tests/test_dolfin.py::test_dolfin[mesh3] - ValueError: could not conve...
FAILED tests/test_gmsh.py::test_gmsh22[False-mesh9] - ValueError: cannot resh...
FAILED tests/test_gmsh.py::test_gmsh22[False-mesh10] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh22[False-mesh11] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh22[False-mesh12] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh22[False-mesh13] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh22[False-mesh14] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh40[False-mesh8] - ValueError: cannot resh...
FAILED tests/test_gmsh.py::test_gmsh40[False-mesh9] - ValueError: cannot resh...
FAILED tests/test_gmsh.py::test_gmsh40[False-mesh10] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh40[False-mesh11] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh40[False-mesh12] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh40[False-mesh13] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh41[False-mesh8] - ValueError: cannot resh...
FAILED tests/test_gmsh.py::test_gmsh41[False-mesh9] - ValueError: cannot resh...
FAILED tests/test_gmsh.py::test_gmsh41[False-mesh10] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh41[False-mesh11] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh41[False-mesh12] - ValueError: cannot res...
FAILED tests/test_gmsh.py::test_gmsh41[False-mesh13] - ValueError: cannot res...
FAILED tests/test_ugrid.py::test_io[1e-07-.ugrid-mesh0] - meshio._exceptions....
[...]

PR #1506, opened 2025-01-31
Do not rely on `repr()` of `np.float64`

musicinmybrain:
Use `.item()` to extract a `float` before applying `repr()`.

Fixes a regression in numpy 2.2:

```
 FAILED tests/test_dolfin.py::test_dolfin[mesh3] - ValueError: could not convert string to float: 'np.float64(63.69616873214543)'
```

See https://github.com/nschloe/meshio/issues/1499.


mgorny:
@musicinmybrain, I've submitted a slightly different fix (using `str()`) in #1513. Not sure, but I think it's more reliable than using `.item()`.

musicinmybrain:
Thanks! Hopefully @nschloe will investigate and choose an approach.

PR #1537, opened 2025-08-22
Update time_series.py

guglielmopadula:
This commit modifies the xdmf time series writer to store the .h5 file in the same path as of the .xmdf


guglielmopadula:
Sorry, just noted already existed a pull request for this.

Issue #1544, opened 2025-11-06
[BUG] Read error when reading binary VTK file


HanNotSoSolo:
**Describe the bug**
When I try to read a binary .vtk file exported from Gmsh 4.11.1, I get  a ReadError. This doesn't happen with ASCII files, or with binary files that aren't exported by Gmsh.
I am (out of my control) using Meshio 4.4.6.

**To Reproduce**
With Gmsh, create a 3D mesh (any will do the job) and Export it as a binary VTK file. Then import meshio and run: `meshio.read('binary_mesh_example.vtk')`

A simple mesh example (~200kB) is provided here: [binary_mesh_example.vtk.zip](https://github.com/user-attachments/files/23396400/binary_mesh_example.vtk.zip)

**Diagnose**
meshio @ file:///home/conda/feedstock_root/build_artifacts/meshio_1622837851801/work

IMPORTANT NOTE: the raised `ReadError` is line 495 of ~/meshio/vtk/_vtk.py concerning the verification of the `line` variable. If I add this line:
`line='\n'`
before the `if` verification, not only the error is not raised (obviously), but I am able to work with the mesh and its caracteristics with Meshio.

**Did I help?**
Not **yet** :)))

If I was able to resolve your problem, consider [sponsoring](https://github.com/sponsors/nschloe) my work on meshio, or [buy me a coffee](https://ko-fi.com/nschloe) to say thanks.

PR #1547, opened 2025-11-12
Draft: support vtk files generated from Gmsh

GenevieveBuckley:
Closes issue https://github.com/nschloe/meshio/issues/1544
Possibly related to https://github.com/GenevieveBuckley/napari-meshio/issues/4 ?

I had a go at addressing issue https://github.com/nschloe/meshio/issues/1544. I am confident that commit [8c42f167ea2379570bce046548c1ddbcba2bcee0](https://github.com/nschloe/meshio/commit/8c42f167ea2379570bce046548c1ddbcba2bcee0) is correct, but I am not confident about the rest, and whether the overall result is correct.

## Details

I used the example file suggested in issue https://github.com/nschloe/meshio/issues/1544:
> A simple mesh example (~200kB) is provided here: [binary_mesh_example.vtk.zip](https://github.com/user-attachments/files/23396400/binary_mesh_example.vtk.zip)

This mesh can be displayed in Paraview:
Screenshot 2025-11-12 at 5 06 48 PM

When you checkout this PR branch:
```
git clone https://github.com/nschloe/meshio.git
cd meshio
gh pr checkout 1544
```
and read in the example Gmsh .vtk file:
```python
import meshio
mesh = meshio.read(filename)
```
You get this summary output:
```
<meshio mesh object>
  Number of points: 2238
  Number of cells:
    tetra: 6383
  Cell data: CellEntityIds
```
So everything seems to work ok so far, although I am not sure if this is the expected output.

To confirm whether it's working correctly, I would really like to load the mesh from meshio, into a visualization program to look at it. Then we could see if it looks like the original input or not.


GenevieveBuckley:
### Why I'm not confident this is correct
A few years ago I made a napari plugin that wraps the meshio library: https://github.com/GenevieveBuckley/napari-meshio/
I had hoped that with the changes from this PR, I could install meshio as an editable dependency locally, and run my napari plugin on top of the changes. Then I should be able to load and view the example Gmsh .vtk file (linked in the description above).

However, when I do that and try to open the Gmsh .vtk file with the napari plugin:
* It doesn't produce an error, and the surface layer that gets added does show reasonable looking data if you inspect it in the terminal
* But the napari canvas still appears black
* When you switch napari to 3D mode, there is a vispy.gloo error:
```python-traceback
RuntimeError: All attributes must have the same size, got:
<VertexBuffer size=25532 last_dim=1>: 25532
<VertexBuffer size=19149 last_dim=3>: 19149
<VertexBuffer size=0 last_dim=3>: 0
<VertexBuffer size=25532 last_dim=3>: 25532
```

I know this isn't a napari related project, but it does cast doubt on whether the proposed fix in this branch is actually correct.

To figure out if the problem is:
* with this PR branch, or
* with the napari-meshio plugin, or napari itself
I'd need to test the output mesh from this PR branch, and visualize it successfully with another program.

Alternatively, a round trip test inspecting all the vertexes and surface information of the mesh before and after the round trip could also work (but I don't have the .geo code used to generate that example Gmsh .vtk file originally, and might have to create a new example).

This might be the limit of the time I have available to work on this, but I wanted to put what I had so far in writing. Perhaps someone else can push it further along.

Issue #1328, opened 2022-06-07
[REQUEST] Support exodus time series


j-bowhay:
**How would you improve meshio?**

It would be nice to be able to read time series exodus files. I was interested in this comment
https://github.com/nschloe/meshio/blob/0138cc8692b806b44b32d344f7961e8370121ff7/src/meshio/exodus/_exodus.py#L136-L139
seems to imply that it might be possible? I would be happy to try attempt this given some guidance.

**What problem does it solved for you?**

With [MOOSE](https://github.com/idaholab/moose) becoming more popular for multiphysics simulation, meshio could be a nice way to get the data into python.

nschloe:
Duplicate of #381.

Issue #381, opened 2019-05-05
Output exodus file from reading in Vtk-ASCII file, point data now shown in exodus file


XiaoMaResearch:
Hi,

I am using the following script to output a exodus file from Reading in an VTK-ASCII file.

import meshio
import glob

input_folder = 'input_vtk'
output_folder = 'output_vtk'
filename = 'hex8_100m-zpos_t8980000.vtk'
input_Path_list = glob.glob(input_folder + '/' + '*.vtk')

for input_Path in input_Path_list:

    filename = input_Path.split('/')
    print(filename)
    output_Path = output_folder + '/' + 'test.exo'
    mesh = meshio.read(input_Path)
    print(output_Path)
    meshio.write_points_cells(
        output_Path,
        mesh.points,
        mesh.cells,
        point_data=mesh.point_data,
        )

However the output file exodus file I got have the two point data but there is no data , is there a similar example that I could follow ?  And also I would like to append all the time series vtk data to the same exodus file, is there a way to append to the same exodus file ? 

Thanks ,

nschloe:
For converting files, simply use `meshio-convert` on the command line.

XiaoMaResearch:
Thank you. Is it possible to convert multiple time series vtk files into one  exodus file ?

nschloe:
Exodus: Not yet. With XDMF you can: https://github.com/nschloe/meshio#time-series.

Issue #249, opened 2018-06-10
Node sets and side sets input/ouput


hillyuan:
Dear repository owner

I am sorry I cannot find where REAME.md wrong and thanks for your modification.

Besides above pull request, I have also completed the implementation of Node sets, Side sets input from Abaqus inp file and output to exodus just now. I pushed it into my dev branch but I am not going to apply pull request because it is far from complete. I will stop this modification at this point, beacuse
1. I don't know what is your plan and what's your design.
2. There are too many works left to implement this function on all those mesh formats

It is basically needed and I think it is better to consider its implementation.

nschloe:
If I understand you correctly, this issue is about how to handle node sets and side sets, and how to incorporate them in meshio. That is indeed something to consider.

Basically, this goes back to https://github.com/nschloe/meshio/issues/198 where the implementation of a container class is discussed. With this, one would read like
```
mesh = meshio.read(filename)
```
and `mesh` has `mesh.x`, `mesh.cells` etc. Additional data like sets could be stored in that object as well.

nschloe:
Alright, I've just now released meshio 2.0.0 with a `Mesh` container class. You can simply add node/side sets to that class and add i/o for Abaqus/Exodus.

PR #1550, opened 2026-02-02
Exodus: I/O of time series data, field data, cell data

Intergalactyc:
## Overview
Fairly significant overhaul to `_exodus.py` to provide the following features:
- Reading of time series data: optional `timestep` kwarg can either be an integer (default 0) to select a given timestep by index, or "all" to instead obtain a list with a Mesh for each timestep. Improved read performance by pre-fetching variable handles and disabling NetCDF auto-masking where not yet done.
- Writing of time series data: new optional kwargs for the write function to manually specify point_data and cell_data (list of dicts ordered by time) with multiple timesteps, as well as time array parameters (and time inference when these are not completely specified)
- Reading and writing of field data (global variables)
- Splitting of multidimensional arrays (vectors) into scalar components with uppercase suffixes (X, Y, Z) during write, and symmetric re-categorization via the (fixed to better accomodate 2d) categorize helper during read to reconstruct original meshio vector shapes; update of existing vector handling during read to work on cell data just as will point data
- Writing of cell data (reading was already supported, but writing was not) with multi-block consistency.
- Fix to element block indexing, as SEACAS based tools often require 1-indexed element blocks
- Writing of side sets (for now only when manually specified)
Minor changes were also made to `_helpers.py` to allow the additional `timestep` kwarg for read to be passed through (non-breaking).

## Backwards compatibility
Fully backwards compatible. All changes made are nonbreaking: call signature changes are only the addition of optional kwargs, and return values should be the same unless these kwargs are used.

## Tests
Added four new tests to `test_exodus.py` to comprehensively test new features. These and all existing exodus tests pass.

## Relevant issues
- #381 (and linked #1328): this provides the general time series support desired
- #249: not a complete solution for this (I didn't want to extend the already wide scope of this too much more), but the manual side-set specification added to `write` can easily be adapted to full support for side sets as described. Node sets are already supported.

Anyways, thanks @nschloe and everyone for this great tool; hopefully these changes can help someone.


KJSchwiebert:
I wanted to say thank you for making these changes. I used it to add side sets when writing a `.exo` and it fit my needs perfectly.

Although I'm only a user of meshio, I wanted to throw out my hope that this gets merged. Thanks again to the PR author and all those who've worked on meshio. It's very handy.

